PR #12, opened 2017-06-29
Remove [replace] of curl. Fix in windows-separate-dir merged into master branch of curl-rust.

reidwagner:
`cargo build` was failing with: 

```
error: failed to load source for a dependency on `curl`

Caused by:
  Unable to update https://github.com/alexcrichton/curl-rust?branch=windows-separate-dir

Caused by:
  failed to find branch `windows-separate-dir`

```
Checked out curl-rust and the fix was merged: https://github.com/alexcrichton/curl-rust/pull/175


danburkert:
Thanks!

PR #15, opened 2017-07-03
Use heck for ident conversions

danburkert:
This is a breaking change, because heck is aggressive about pruning
prefix and suffix underscores, whereas the previous algorithm was not.
This needs to wait for withoutboats/heck#3 to be resolved, because
identifiers like `FIELD_NAME11` are recommended according to the
[Protobuf style guide][1].

[1]: https://developers.google.com/protocol-buffers/docs/style


danburkert:
Merged manually.

PR #21, opened 2017-07-17
Benchmark

### LINKED ISSUE #4: varint encoding/decoding/len functions are not optimized
Protobuf performance tends to be dominated by varint operations.  The [varint functions](https://github.com/danburkert/prost/blob/v0.1.0/src/encoding.rs#L38-L98) in `prost` have not been optimized at all. There are some benchmarks, but I think they've bit-rotted at this point.

danburkert:
This series of patches improves/adds varint encoding/decoding/encoded-len microbenchmarks, and ports over optimizations from the cpp protobuf library. Performance improvement numbers are included in the individual commit messages.

Fixes #4 

CC @overvenus

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,3 +23,4 @@ rust:
 script:
   - cargo build --verbose --all
   - cargo test --verbose --all
+  - cargo bench --benches --verbose --all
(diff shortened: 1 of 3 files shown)


overvenus:
It is amazing, especially the decode performance, well done! :+1: 

The benchmark result is updated. https://github.com/tafia/quick-protobuf/pull/63/commits/bdeb7039f914e4b846bdf6736f9aa57ae467c719

PR #27, opened 2017-09-05
Dropped curl dependency, download protoc using reqwest instead

DamianX:
This makes it significantly easier to build prost-build on Windows, at least for me.

### DIFF
--- a/prost-build/Cargo.toml
+++ b/prost-build/Cargo.toml
@@ -22,7 +22,7 @@ prost-types = { path = "../prost-types" }
 tempdir = "0.3"
 
 [build-dependencies]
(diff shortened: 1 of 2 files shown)


danburkert:
Hi @DamianX, this seems like a fine change to me, but I'm curious what issues you ran in to?  I'm not a windows user, but the crate appears to work fine on appveyor without having to do anything special to get curl.  Anyway, I'm fine with moving to reqwest if it's easier, but please also switch the build.rs in the `protobuf` subproject over to it.

DamianX:
Oddly enough, I'm unable to reproduce the problem I encountered before. I can build curl just fine.

PR #45, opened 2017-11-03
Add optional serde derives in prost-build

LinearZoetrope:
This documents and enables a feature gate that allows
the prost-generated structs to be serialized by serde-compatible
encoders and decoders.

This does not and is not intended to make prost "serde-compatible",
simply to ease the use of serde on user structs that may wish to
contain decoded messages.

### DIFF
--- a/prost-build/Cargo.toml
+++ b/prost-build/Cargo.toml
@@ -21,6 +21,10 @@ prost = { version = "0.2", path = ".." }
 prost-types = { version = "0.2", path = "../prost-types" }
 tempdir = "0.3"
 
+[features]
+# We don't actually need serde because it's just text generation
+serde-1 = []
+
 [build-dependencies]
 curl = "0.4"
 tempdir = "0.3"
--- a/prost-build/src/code_generator.rs
+++ b/prost-build/src/code_generator.rs
@@ -161,7 +161,14 @@ impl <'a> CodeGenerator<'a> {
 
         self.append_doc();
         self.push_indent();
-        self.buf.push_str("#[derive(Clone, Debug, PartialEq, Message)]\n");
+
+        self.buf.push_str("#[derive(Clone, Debug, PartialEq, Message");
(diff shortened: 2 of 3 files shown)


danburkert:
Hi @LinearZoetrope, thanks for the PR!  Instead of using cargo features to enable this, what do you think about adding an option to `prost_build::Config`?  I think I'd prefer to have it not specific to serde, but instead allow deriving any trait, because the set of traits people may want is large.  For instance:

```rust
    /// Add derive annotations for the provided traits to generated message structs and oneof
    /// enums.
    pub fn derive_traits<I, S>(&mut self, traits: I) -> &mut Self
    where I: IntoIterator<Item = S>,
          S: AsRef<str> {
        ...
    }
```

vorner:
This seems to be dormant for a while. Would anybody mind if I gave it a shot (incidentally I may have a use for such a feature myself)?

I was wondering if it could be generalized even further. Instead of returning a list of traits, just provide a string with arbitrary attributes, and let it be matched on similar way as `btree_map`. Eg. something like:

```rust
config.attributes(".my_messages.MyMessageType", "#[derive(Serialize, Deserialize)] #[serde(rename-all = "kebab-case")]");
```

danburkert:
> I was wondering if it could be generalized even further. Instead of returning a list of traits, just provide a string with arbitrary attributes, and let it be matched on similar way as btree_map.

That sounds good to me.  I think the implementation would need to allow `attributes` to be called multiple times to add different sets of attributes to different messages.  May also make sense to only take a single attribute at a time with the brackets stripped:

```rust
config.attributes(".my_messages.MyMessageType", "derive(Serialize, Deserialize)")
config.attributes(".my_messages.MyMessageType", "serde(rename-all = \"kebab-case\")")
```

But that's more of a matter of taste, I could be swayed either way.

LinearZoetrope:
I'll close this because of the merge of #63

PR #47, opened 2017-11-05
Add support for recursive `oneof` fields

olix0r:
Currently, `oneof`s that reference a parent type result in compile errors.

By replicating the `boxed` check in field generation, oneofs may be recursive.

Fixes danburkert/prost#46

### DIFF
--- a/prost-build/src/code_generator.rs
+++ b/prost-build/src/code_generator.rs
@@ -378,14 +378,15 @@ impl <'a> CodeGenerator<'a> {
             self.push_indent();
             let ty = self.resolve_type(&field);
 
-            let repeated = field.label == Some(Label::LabelRepeated as i32);
-            let boxed = !repeated
-                    && type_ == Type::TypeMessage
-                    && self.message_graph.is_nested(field.type_name(), msg_name);
+            // NOTE: Unfortunately, `is_nested` doesn't know whether the reference is
+            // through a list or a map, so this may box variants when not strictly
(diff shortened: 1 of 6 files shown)


danburkert:
Good catch, thanks for the PR.  The CI failure is because the bootstrap test is failing - it's picking up nesting even through a map field.  You can reproduce with `cargo test --all`.  I think this is a bug even with the original message nesting; we're checking that it's not repeated, but we're not checking that it's not a map field.  Could you add your test case to `tests`, and extend it with repeated and map fields to make sure those don't box?

danburkert:
>  I think this is a bug even with the original message nesting; we're checking that it's not repeated, but we're not checking that it's not a map field.

olix0r:
@danburkert Thanks. I've pushed a change with a test case.

Unfortunately, when we check to see whether a oneof variant references an outer type, we don't know whether this reference is through a list or map. This causes the sometimes-unncessary boxing of oneof variants so that the size of a oneof is always known at compile-time.

Updating prost-types to have admit these unnecessary boxes allows the bootstrap test to pass.

I understand that this boxing is undesirable, but I don't understand this code base well enough to propose a solution that does not incur this cost. I'm open to other ideas.

danburkert:
I think I found a relatively simple workaround for the boxing issue.  We'll see what CI says :).

danburkert:
Squashed to master manually.  Thanks again for the PR.  Would a point release be helpful for y'all?

danburkert:
Fixed in 0.2.3.

PR #50, opened 2017-11-12
Upgrade itertools to 0.7

fabricedesre:
That matches the version in prost-derive to not depend on two version of the itertools crate.


danburkert:
Thanks!

PR #54, opened 2017-11-18
WIP: Custom Debug implementation

vorner:
This is my first attempt to https://github.com/danburkert/prost/issues/49. I plan to:

* Add similar support for the OneOf containers
* Do some cleanups and code deduplication (it feels some bits are quite similar, but I'm not sure if I'll able to make something usable of it).

However, I'd like some feedback to know early if I'm doing it completely wrong (like, in a wrong place). Also, an idea how to test this would be nice. I did test it manually in my own downstream project, but something automated (in addition to the fact the debug implementation compiles) would be nice ‒ but parsing how it looks formatted seems a bit overkill.

Also, it's a bit unfortunate, but it would be a breaking change in the sense the user can no longer do `#[derive(Debug)]` on manually written structs as that collides. Should I add some attribute to enable the debug implementation and default to the old behaviour, or is this OK?

### DIFF
--- a/prost-build/src/code_generator.rs
+++ b/prost-build/src/code_generator.rs
@@ -161,7 +161,7 @@ impl <'a> CodeGenerator<'a> {
 
         self.append_doc();
         self.push_indent();
-        self.buf.push_str("#[derive(Clone, Debug, PartialEq, Message)]\n");
+        self.buf.push_str("#[derive(Clone, PartialEq, Message)]\n");
         self.push_indent();
         self.buf.push_str("pub struct ");
         self.buf.push_str(&to_upper_camel(&message_name));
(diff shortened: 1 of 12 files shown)


danburkert:
> Also, it's a bit unfortunate, but it would be a breaking change in the sense the user can no longer do #[derive(Debug)] on manually written structs as that collides. Should I add some attribute to enable the debug implementation and default to the old behaviour, or is this OK?

I agree this should cause the next version to be 0.3, but that's no big deal.  I'm not really concerned about losing the ability to manually define Debug on non-generated message structs.  I'm not actually aware of anyone using prost with manually-declared structs, much less anyone doing that and needing control over the debug formatting.  If this ends up being an issue, we can always add it back later.

vorner:
OK, I'll have a look. I'll be probably pushing further fixup commits (I plan to squash them just before merge) for several days, I'll explicitly say when I think it is ready for further review.

vorner:
I think I've addressed all the points and it seems to do what it should. Could you have another look?

danburkert:
I rebased manually and pushed to master.  This is a really nice improvement, thanks again!

Issue #59, opened 2017-12-03
Documentation for the ServiceGenerator trait


vorner:
Hello

What is the status around services in prost? There seem to be some support in the `ServiceGenerator` trait in the `prost-build` crate, but it is unclear how it should work and what the result would be.

Would it be possible to have an example with a service?

danburkert:
There's an example service generator impl here: https://github.com/danburkert/prost/blob/master/prost-build/src/lib.rs#L419-L440.  There's no default service generator, because there's really no good default.  Service generation is an inherently higher-level concept than what `prost` provides. 

That being said, I don't think it will be up to individual applications to define their own service generator.  That will usually be the responsibility of an RPC framework.  For instance, if or when there is a grpc framework implementation using `prost`, it would provide the service generator.  `prost` can't provide it, because it doesn't know whether it should be a sync or async API, whether there needs to be framework-specific context passed in to the method handlers, whether it should be a trait or some other abstraction, etc.

vorner:
OK, so now that I have better understanding what this is about, I'll try to write some explanation into the documentation.

Out of curiosity, do you plan on writing something like prost-rpc yourself, or are you leaving it up to others?

danburkert:
> OK, so now that I have better understanding what this is about, I'll try to write some explanation into the documentation.

Thanks! This is super helpful.

Out of curiosity, do you plan on writing something like prost-rpc yourself, or are you leaving it up to others?

I have an alpha-quality RPC system I'm working on which leverages `prost`, but it's not really meant for public consumption.  It's an RPC protocol specific to Apache Kudu.  I'm slowly iterating on it over [here](https://github.com/danburkert/kudu-rs/tree/futures-wip/krpc).

Otherwise, @carllerche just published [`tower-grpc`](https://github.com/tower-rs/tower-grpc), which uses `prost`, which I believe is considered alpha at this point as well.

vorner:
OK, let's close this one then ☺

PR #60, opened 2017-12-04
Enhance docs for the service generator

vorner:
The expected usage wasn't clear to me, it is likely it wouldn't be clear
for others.

#59.


danburkert:
Thanks!

PR #63, opened 2017-12-08
Allow generation of attributes on messages and fields

vorner:
This allows messages to be for example Serialize or Eq.

This is my attempt at what is suggested in #45. I hope the usage is clear from the documentation.

If you know of any attribute that can be placed on a field without bringing another dependency, I'll be happy to replace that slightly hacky comment test.

### DIFF
--- a/prost-build/src/code_generator.rs
+++ b/prost-build/src/code_generator.rs
@@ -24,7 +24,7 @@ use ast::{
 };
 use ident::{
     to_snake,
-    match_field,
+    match_ident,
     to_upper_camel,
 };
 use message_graph::MessageGraph;
@@ -162,7 +162,7 @@ impl <'a> CodeGenerator<'a> {
         self.append_doc();
         self.push_indent();
         self.buf.push_str("#[derive(Clone, PartialEq, Message)]\n");
-        self.append_container_attributes(&fq_message_name);
+        self.append_type_attributes(&fq_message_name);
         self.push_indent();
         self.buf.push_str("pub struct ");
         self.buf.push_str(&to_upper_camel(&message_name));
@@ -223,12 +223,10 @@ impl <'a> CodeGenerator<'a> {
         }
     }
 
-    fn append_container_attributes(&mut self, msg_name: &str) {
+    fn append_type_attributes(&mut self, msg_name: &str) {
         assert_eq!(b'.', msg_name.as_bytes()[0]);
-        for &(ref matcher, ref attribute) in &self.config.container_attributes {
-            // We abuse the field matcher to match only whole message paths, since the ~~~ field is
-            // illegal. This saves code and hopefuly doesn't have any strange side effects.
-            if match_field(matcher, msg_name, "~~~") {
+        for &(ref matcher, ref attribute) in &self.config.type_attributes {
+            if match_ident(matcher, msg_name, None) {
                 self.push_indent();
                 self.buf.push_str(attribute);
                 self.buf.push('\n');
@@ -239,7 +237,7 @@ impl <'a> CodeGenerator<'a> {
     fn append_field_attributes(&mut self, msg_name: &str, field_name: &str) {
         assert_eq!(b'.', msg_name.as_bytes()[0]);
         for &(ref matcher, ref attribute) in &self.config.field_attributes {
-            if match_field(matcher, msg_name, field_name) {
+            if match_ident(matcher, msg_name, Some(field_name)) {
                 self.push_indent();
                 self.buf.push_str(attribute);
                 self.buf.push('\n');
@@ -336,7 +334,7 @@ impl <'a> CodeGenerator<'a> {
         let btree_map = self.config
                             .btree_map
                             .iter()
-                            .any(|matcher| match_field(matcher, msg_name, field.name()));
+                            .any(|matcher| match_ident(matcher, msg_name, Some(field.name())));
         let (annotation_ty, rust_ty) = if btree_map {
             ("btree_map", "BTreeMap")
         } else {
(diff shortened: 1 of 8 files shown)


vorner:
Hello.

Could I inquire over the status here? 😇

prost-build/src/code_generator.rs
```diff
@@ -219,6 +223,30 @@ impl <'a> CodeGenerator<'a> {
         }
     }
 
+    fn append_container_attributes(&mut self, msg_name: &str) {
+        assert_eq!(b'.', msg_name.as_bytes()[0]);
+        for &(ref matcher, ref attribute) in &self.config.container_attributes {
+            // We abuse the field matcher to match only whole message paths, since the ~~~ field is
```
danburkert: this is kind of hacky.  I think it would be simpler to change the `match_field` method to the following signature:

```
pub fn match_ident(matcher: &str, msg: &str, field: Option<&str>) -> bool;
```

And then reuse it for types (which would pass `None` for the field), as well as fields.  Internally all it has to do is not push on to the paths vec when the arg is `None`.

tests/src/lib.rs
```diff
@@ -211,6 +211,22 @@ mod tests {
         roundtrip::<foo::bar_baz::FooBarBaz>(&buf).unwrap();
     }
 
+    #[test]
+    fn test_custom_container_attributes() {
```
danburkert: Good test, but please extend it to cover enums and one-ofs.  In particular I'm not sure how the current code will work with one-ofs, since they don't really have a Protobuf type name.

tests/src/lib.rs
```diff
@@ -211,6 +211,22 @@ mod tests {
...
+        // We abuse the ident conversion protobuf for the custom attribute additions. We placed
+        // `Eq` on the FooBarBaz (which is not implemented by ordinary messages).
+        let msg1 = foo::bar_baz::FooBarBaz::default();
+        let msg2 = foo::bar_baz::FooBarBaz::default();
+        // This uses Eq, which wouldn't compile if the attribute didn't work
+        assert_eq!(msg1, msg2);
+    }
+
+    #[test]
+    fn test_custom_field_attributes() {
```
danburkert: Likewise, check that this covers oneof and enum variants.

prost-build/src/lib.rs
```diff
@@ -203,6 +205,9 @@ impl Config {
     /// matched on the fully qualified field name. If a Protobuf map field matches any of the
     /// paths, a Rust `BTreeMap` field will be generated instead of the default [`HashMap`][3].
     ///
+    /// The matching is done on the protobuf names, before converting to Rust-friendly casing
```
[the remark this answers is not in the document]
vorner: I might be unclear about the terminology. But I actually had to look into the code to know if it matched on the original (protobuf) names or on the generated. It only said it is on fully qualified ones.

prost-build/src/lib.rs
```diff
@@ -243,6 +248,75 @@ impl Config {
...
+    /// # let mut config = prost_build::Config::new();
+    /// // Nothing around uses floats, so we can derive real `Eq` in addition to `PartialEq`.
+    /// config.container_attribute(".", "#[derive(Eq)]");
+    /// // Some messages want to be serializable with serde as well.
+    /// config.container_attribute(".my_messages.MyMessageType",
+    ///                            "#[derive(Serialize)] #[serde(rename-all = \"snake_case\")]");
+    /// config.container_attribute(".my_messages.MyMessageType.MyNestedMessageType",
+    ///                            "#[derive(Serialize)] #[serde(rename-all = \"snake_case\")]");
+    /// ```
+    pub fn container_attribute<P, A>(&mut self, path: P, attribute: A) -> &mut Self
```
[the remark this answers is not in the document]
vorner: I'm OK with either. I'm not really good at naming things.

vorner:
I think I have the next iteration now. And you were right to point out the missing tests, another instance of what isn't tested is broken. I hope I haven't forgotten anything important.

vorner:
Hello. Ping again? 😇

tests/src/lib.rs
```diff
@@ -43,6 +43,20 @@ pub mod recursive_oneof {
     include!(concat!(env!("OUT_DIR"), "/recursive_oneof.rs"));
 }
 
+/// This tests the custom attributes support by abusing docs.
+///
+/// Docs really are full-blown attributes. So we use them to ensure we can place them on everything
+/// we need. If they aren't put onto something or allowed not to be there (by the generator),
+/// compilation fails.
+#[deny(missing_docs)]
```
danburkert: This is really clever 👍

prost-build/src/ident.rs
```diff
@@ -38,7 +38,7 @@ pub fn to_upper_camel(s: &str) -> String {
 }
 
 /// Matches a 'matcher' against a fully qualified field name.
```
danburkert: s/field name/identifier

tests/src/oneof_attributes.proto
```diff
@@ -0,0 +1,10 @@
+syntax = "proto3";
+
+package Foo.Custom.OneOfAttrs;
```
danburkert: Is it necessary for this to be in a separate file/package than custom_attributes.proto?  If not, I think it would be good to consolidate.  If there's some reason to have them separate than maybe a comment explaining would be good.

tests/src/lib.rs
```diff
@@ -211,6 +225,25 @@ mod tests {
...
+        let msg1 = foo::bar_baz::FooBarBaz::default();
+        set1.insert(msg1);
+        // Similar, but for oneof fields
+        let mut set2 = BTreeSet::new();
+        let msg2 = oneof_attributes::Msg::default();
+        set2.insert(msg2.field);
+    }
+
+    #[test]
+    fn test_custom_field_attributes() {
```
danburkert: Is this test case getting us more coverage than the #[deny(missing_docs)] trick?

vorner: Not really. I just never learned to kill tests that are no longer needed unless they fail. It is gone now, because it was so ugly anyway.

tests/src/oneof_attributes.proto
vorner: I added the explanation. Unfortunately, the companion module for an oneof field triggers the missing-docs error and I see no way how to place docs on that one (I don't think it makes sense to extend the support of the custom attributes to that kind of module).

danburkert: Makes sense, thanks for clarifying.

PR #65, opened 2018-01-04
Add ServiceGenerator::finalize

vorner:
It can be used to output things that need to exist only once per file
(eg. a module). Contains empty default implementation, for cases when no
such thing is needed.

This is a solution for a real-life problem, the `tower-grpc` wants to put its traits into submodules (`client` and `service`), but this fails if there's more than one service in a proto file. See https://github.com/tower-rs/tower-grpc/pull/10.

### DIFF
--- a/prost-build/src/code_generator.rs
+++ b/prost-build/src/code_generator.rs
@@ -106,16 +106,12 @@ impl <'a> CodeGenerator<'a> {
 
         if let Some(ref service_generator) = code_gen.config.service_generator {
             code_gen.path.push(6);
-            let mut generated = false; // We call finalize only if there's at least one service
             for (idx, service) in file.service.into_iter().enumerate() {
                 code_gen.path.push(idx as i32);
                 service_generator.generate(code_gen.unpack_service(service), &mut code_gen.buf);
                 code_gen.path.pop();
-                generated = true;
             }
-            if generated {
-                service_generator.finalize(&mut code_gen.buf);
-            }
+            service_generator.finalize(&mut code_gen.buf);
             code_gen.path.pop();
         }
     }
--- a/prost-build/src/code_generator.rs
+++ b/prost-build/src/code_generator.rs
@@ -106,10 +106,15 @@ impl <'a> CodeGenerator<'a> {
 
         if let Some(ref service_generator) = code_gen.config.service_generator {
             code_gen.path.push(6);
+            let mut generated = false; // We call finalize only if there's at least one service
             for (idx, service) in file.service.into_iter().enumerate() {
                 code_gen.path.push(idx as i32);
                 service_generator.generate(code_gen.unpack_service(service), &mut code_gen.buf);
                 code_gen.path.pop();
+                generated = true;
+            }
+            if generated {
+                service_generator.finalize(&mut code_gen.buf);
             }
             code_gen.path.pop();
         }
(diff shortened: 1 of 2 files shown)


vorner:
I suspect the failures on nightlies are https://github.com/rust-lang/rust/issues/47175.

prost-build/src/code_generator.rs
```diff
@@ -106,10 +106,15 @@ impl <'a> CodeGenerator<'a> {
 
         if let Some(ref service_generator) = code_gen.config.service_generator {
             code_gen.path.push(6);
+            let mut generated = false; // We call finalize only if there's at least one service
```
danburkert: It may be nice to actually call it regardless.  I admit I don't really have a usecase in mind, but it seems strictly more flexible to me.

prost-build/src/lib.rs
```diff
@@ -173,6 +173,20 @@ pub trait ServiceGenerator {
     /// Generates a Rust interface or implementation for a service, writing the
     /// result to `buf`.
     fn generate(&self, service: Service, buf: &mut String);
+    /// Finalizes the generation process.
```
danburkert: Add a blank line above

prost-build/src/lib.rs
```diff
@@ -173,6 +173,20 @@ pub trait ServiceGenerator {
...
+    ///
+    /// In case there's something that needs to be output at the end of the generation process, it
+    /// goes here. Similar to [`generate`](#method.generate), the output should be appended to
+    /// `buf.
```
danburkert: missing a closing tick.

prost-build/src/lib.rs
```diff
@@ -173,6 +173,20 @@ pub trait ServiceGenerator {
...
+    ///
+    /// An example can be a module or other thing that needs to appear just once, not for each
+    /// service generated.
+    ///
+    /// This still can be called multiple times in a lifetime of the service generator, because it
+    /// can be used through multiple `.proto` files.
```
danburkert: I think it would be clearer as ', because it is called once per `.proto` file.'

danburkert: Oh I suppose this is only true if it's changed to run whether or not there are any services defined.

prost-build/src/code_generator.rs
danburkert: If there's a compelling argument for keeping it the way it is, then I think I'd prefer skipping the boolean flag and just checking if the vector is empty, e.g.

+            if !file.service.is_empty() {
+                service_generator.finalize(&mut code_gen.buf);
             }

vorner: There wasn't a strong reason for that, only a feeling that the service generator shouldn't be bothered with files that don't have any services.

But on the other hand, the logic is needlessly more complex, so let's get rid of it. The service generator can decide on its own not to touch such files.

danburkert:
Looks great.  @vorner are you ready for another release?  I think there's enough new features that we should, but I'll hold off if you have more features you want to add in the short term.

danburkert:
Well I suppose I want to hold off at least until the nightly rustc thing is fixed, but hopefully that's just a day or two.  I don't want to release and then have people's first experience be a segfault :)

vorner:
I don't have any short-term plans myself, or actually any plans. I simply try to use it from time to time and if I find a problem, I think about fixing it or at least opening an issue.

So I'm for a release :-)

vorner:
Is it the right time for a release? 😇

PR #73, opened 2018-02-03
Simplify the field annotations for existing types

kestred:
This pull request makes it easier to use existing types (or handwritten types) to serialize to Protobuf.

Updates to `prost-derive` (see README for details):
  - Automatically infer scalar types from rust types
  - Add the `#[prost(tags = "sequential")]` struct annotation to infer tags, with support for skipping
     reserved/unused tags and specifying fields out of order.

Example of simpler derive (with explanatory comments):
```rust
#[derive(Clone, Debug, PartialEq, Message)]
#[prost(tags="sequential")]
struct Person {
  #[prost(fixed64)]
  pub id: u64, // tag=1, type=fixed64

  // NOTE: Old "name" field has been removed
  // pub name: String, // tag=2, type=string (Removed)

  // NOTE: Specify fields out of order
  #[prost(tag="6")]
  pub given_name: String, // tag=6, type=string
  pub family_name: String, // tag=7 type=string
  pub formatted_name: String, // tag=8, type=string

  #[prost(tag="3")]
  pub age: u32, // tag=3, type=uint32
  pub height: u32, // tag=4, type=uint32
  #[prost(enumeration="Gender")]
  pub gender: i32, // tag=5, type=enum

  // NOTE: Skip to less commonly occurring fields (encoded with 2+ bytes)
  #[prost(tag="16")]
  pub name_prefix: String, // tag=16  (eg. mr/mrs/ms)
  pub name_suffix: String, // tag=17  (eg. jr/esq)
  pub maiden_name: String, // tag=18
}

#[derive(Clone, Copy, Debug, PartialEq, Eq, Enumeration)]
pub enum Gender {
  Unknown = 0,
  Female = 1,
  Male = 2,
}
```


danburkert:
I've just done a very preliminary pass, but my first impression is a big 👍.  This will make prost definitions a lot easier to write by hand.  I'll be following up with a more thorough review.   Out of curiosity, are you using prost without codegen?

danburkert:
Does this change the existing annotations at all, or does it just add new ones + allow existing ones to be omitted?  I ask because there are backwards compat concerns, however if it's strictly more flexible then I think this can be merged into the 0.3.x line as long as we don't change codegen.  Once we want to move to 0.4.x we could consider having codegen output the simplified annotations.

kestred:
@danburkert: 

This does not change existing annotations at all.

Some type level annotations may be omitted (`repeated`, `optional`, `uint64`, `bool`, other scalars).
It currently does not support omitting `message` or `map`, but its reasonably easy to implement automatically detecting `map`, `btree_map`, etc automatically.

-----

It only adds one new annotation, which is the struct-level annotation (an optional annotation) which then further allows you to omit `tag` annotations.     There were no previous struct annotations for `#[derive(Message)]`, so this should also be backwards compatible.

I felt it important to make inference of tag-annotations explicit; because its very common for `.proto` definitions to skip to tag 16 and the behavior would be wrong (AND with no compiler error) for most beginner users of `prost`.

-----

For my use of `prost`, I intend to use it both with and without codegen.

Too encourage Service API stability, and make the rust code easier to explore and better integrated with more code editors out of the box, I'm checking the generated (+ hand modified) rust definitions along-side a usual codegen warning to not edit the file casually.

Also, while I know I can add annotations like Serde in a `build.rs` file (and for `serde` I do); there are other annotations/macros/changes that I do outside of the build so that I can expose my Protobuf definitions in a crate with more carefully crafted and documented Rust API.

I then separately have a little test module that double-roundtrips my hand-modified types with the automatically generated definitions using `prost-build` to ensure my modifications haven't led to incorrect serialization.

danburkert:
Sorry for the delay, I haven't forgotten about this, but unfortunately it's going to be a few days before I can get around to a proper review.  Thanks again for the PR!

prost-derive/src/field/map.rs
```diff
@@ -56,7 +56,7 @@ pub struct Field {
 
 impl Field {
 
-    pub fn new(attrs: &[Meta]) -> Result<Option<Field>, Error> {
+    pub fn new(attrs: &[Meta], default_tag: Option<u32>) -> Result<Option<Field>, Error> {
         let mut types = None;
         let mut tag = None;
```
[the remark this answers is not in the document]
danburkert: Actually I think you can just make the signature:

`new(attrs: &[Meta], mut tag: Option<u32>)` and get rid of line 61 and 106-109.

prost-derive/src/field/mod.rs
```diff
@@ -37,21 +38,25 @@ impl Field {
...
+        } else if let Some(field) = message::Field::new(&attrs, default_tag)? {
             Field::Message(field)
-        } else if let Some(field) = map::Field::new(&attrs)? {
+        } else if let Some(field) = map::Field::new(&attrs, default_tag)? {
             Field::Map(field)
         } else if let Some(field) = oneof::Field::new(&attrs)? {
             Field::Oneof(field)
         } else {
-            bail!("no type attribute");
+            if let Some(rt) = rust_type {
```
danburkert: collapse this into an `else if let` (remove one layer of nesting).

prost-derive/src/field/scalar.rs
```diff
@@ -35,7 +36,7 @@ pub struct Field {
 
 impl Field {
 
-    pub fn new(attrs: &[Meta]) -> Result<Option<Field>, Error> {
+    pub fn new(attrs: &[Meta], rust_type: Option<&Type>, default_tag: Option<u32>) -> Result<Option<Field>, Error> {
```
danburkert: likewise (`default_tag`)

prost-derive/src/field/scalar.rs
```diff
@@ -732,3 +747,78 @@ impl quote::ToTokens for DefaultValue {
...
+                    Some(syn::punctuated::Pair::End(&syn::GenericArgument::Type(ref inner))) => inner,
+                    _ => return None,
+                }
+            } else {
+                ty
+            };
+
+        let tok = quote!(#ty);
+        if label == Some(Label::Repeated) && tok == quote!(u8) {
+            Some(Inferred::new(Ty::Bytes, None))
```
danburkert: This is also covered on line 819, are they both necessary?

carllerche:
👍  on the idea.

I have not used prost_derive manually before, so I don't have much experience there. Is `#[prost(tags = "sequential")]` required? Could this be the default? It seems that adding the behavior could be backwards compatible as all prior fields would have needed a tag.

kestred:
Yeah the struct tag isn't really necessary; initially I'd thought a few more obvious ways to tag the struct would come up, but it wasn't so in the end.

People who forget that tags are a thing will get caught by forgetting them, but as long as the docs in the example specify a tag occasionally, it should be a sufficient reminder.

I will try to update the PR with the requested changes soon.

prost-derive/src/field/map.rs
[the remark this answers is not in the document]
kestred: This change cannot be made because `set_option` will return the `"duplicate tag attributes"` if `default_tag` is `Some`.

prost-derive/src/field/message.rs
```diff
@@ -17,7 +17,7 @@ pub struct Field {
 }
 
 impl Field {
-    pub fn new(attrs: &[Meta]) -> Result<Option<Field>, Error> {
+    pub fn new(attrs: &[Meta], default_tag: Option<u32>) -> Result<Option<Field>, Error> {
         let mut message = false;
         let mut label = None;
         let mut tag = None;
```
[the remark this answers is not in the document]
kestred: This change cannot be made because `set_option` will return the `"duplicate tag attributes"` if `default_tag` is `Some`.

prost-derive/src/field/scalar.rs
kestred: Yes.

The assignment at line ~785 (`let ty = ...`) strips either `Vec<T>` or `Option<T>` from the type causing `tok` to be just `quote!(T)`.

This check then notes, if the tok is `u8` and `Vec` was stripped from the type (e.g. if `label` is `Label::Repeated`), then we don't actually want a repeated type, we wanted protobuf `bytes` instead.

The later condition at line 819 then handles the case for `repeated bytes` and `optional bytes`.

I will try to add some comments to that effect.

prost-derive/src/field/scalar.rs
kestred: likewise `set_option` and errors

kestred:
@danburkert:

I have updated the PR as requested.

This includes removing the `#[prost(tags = "sequential")` attribute;  inferring tags is now the default behavior.

Additionally I have addressed your concern about `::std::vec::Vec<_>` et al;  now fully or partially qualified `Vec`, `String`, and `Option` will be inferred correctly.

carllerche:
Closing due to inactivity. If this needs to be revisited, a new PR will be best (reference this one).

PR #78, opened 2018-02-23
Reduce env_logger-related dependencies.

briansmith:
Use env_logger 0.5 everywhere, instead of a mix of 0.5 and 0.4, to
avoid needing to download and build two versions. This won't have an
effect on standalone builds of prost because prost has a dev-dependency
on quickcheck, which uses env_logger 0.4 still. However, it will have
a positive effect on things that depend on prost-build and that use
env_logger 0.5.

Also, don't automatically enable the "regex" default feature of
env_logger. Things that depend on Prost can still enable it (and will
enable it by default, since it is a default feature) if they need it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,7 @@ byteorder = "1"
 bytes = "0.4"
 
 [dev-dependencies]
-env_logger = "0.5"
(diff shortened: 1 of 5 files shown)


danburkert:
Thanks!

PR #79, opened 2018-02-24
Embed protobuf tools & includes instead of downloading during the build.

briansmith:
Implement the proper semantics for `cargo fetch` by avoiding any networking
during the build. Instead, embed the files we would download into prost
itself. This makes the build faster, more cacheable, and more reliable
overall.

Importantly, this eliminates very heavy and brittle non-Rust dependencies
including in particular curl and OpenSSL. This in turn reduces the
typical operating-system-level dependencies, allowing for more lightweight
build environments for prost-based applications.


protobuf/build.rs
```diff
@@ -1,124 +0,0 @@
-extern crate curl;
```
danburkert: I don't think you should remove this, this crate has to compile the protobuf project from source in order to build the conformance test runner.  Do the conformance tests pass with this removed?  I'd probably just avoid touching this crate at all in this PR, since it isn't published and is only used for tests.

third-party/protobuf/readme.txt
```diff
@@ -0,0 +1,18 @@
+`LICENSE` was taken from https://github.com/google/protobuf/archive/v3.5.1.zip.
```
danburkert: Please name this `README.md` to remain consistent with the rest of the repo.

third-party/protobuf/readme.txt
```diff
@@ -0,0 +1,18 @@
...
+* https://github.com/google/protobuf/releases/download/v3.5.1/protoc-3.5.1-osx-x86_64.zip
+* https://github.com/google/protobuf/releases/download/v3.5.1/protoc-3.5.1-win32.zip
+
+The `include` directory's contents were taken from the linux-x86_64 package.
+
+The naming convention was changed to match naming conventions:
+
+* "x86_32" became "i686",
+* "aarch_64" became "aarch64"
+* "osx" became "darwin
```
danburkert: missing `"`

prost-build/build.rs
```diff
@@ -1,137 +1,35 @@
...
-        Err(env::VarError::NotPresent) => return false,
-        Err(env::VarError::NotUnicode(..)) => panic!("PROTOC environment variable is not valid UTF-8"),
-    };
-
-    let protoc_include = match env::var("PROTOC_INCLUDE") {
-        Ok(val) => PathBuf::from(val),
-        Err(env::VarError::NotPresent) => panic!("PROTOC_INCLUDE environment variable not set (PROTOC is set)"),
-        Err(env::VarError::NotUnicode(..)) => panic!("PROTOC_INCLUDE environment variable is not valid UTF-8"),
+    // Recursively copy all the protobuf files to `OUT_DIR` so that they are
+    // available for use at runtime. The binaries for all platforms are copied
```
danburkert: > The binaries for all platforms...

I don't really follow this reasoning, maybe clarify it a bit?

I think an alternate strategy could be to create a crate per platform that each bundle a single `protoc` and have Cargo include the right one through platform `cfg`s, but honestly that's going to complicate the publishing and updating process so much that I'd rather keep it as is.

prost-build/build.rs
```diff
@@ -1,137 +1,35 @@
...
+    let options = {
+        let mut options = dir::CopyOptions::new();
+        options.overwrite = true;
+        options
     };
-    format!("https://github.com/google/protobuf/releases/download/v{version}/protoc-{version}-{platform}.zip",
-            version = VERSION,
-            platform = platform)
+
+    println!("../third-party/protobuf");
```
danburkert: leftover debug logs?

prost-build/build.rs
briansmith: Sure. In my case I use Windows 10 and so I have Win32 and Microsoft Linux host architectures that use the same (target) output directories on the same system. In another case I have Linux using SAMBA to share the directories with a remote Windows system. I'm trying to find some reasonable way to support this kind of situation.

briansmith:
> Only other thing I'd like to see is a simplification to prost-types. Right now prost-types has pre-compiled .rs files with a bootstrapping test that ensures they are kept up to date. Since we're vendoring the source .protos now, it should be possible to simply use prost-build to generate the corresponding .rs files during prost-types build.

I suggest we attempt to solve that bootstrapping problem in a separate issue/PR. It is more complicated than one might think. In particular, prost-build depends on prost-types so we prost-types's build script can't depend on the current prost-build. It could depend on the previous version (e.g. if we're compiling 0.3.1, we could use prost-build 0.3.0). However, that's counterproductive to the goal of this PR, which is to eliminate dependencies and improve build time and build reliability, since that would require users of prost 0.3.1 to build prost 0.3.0, which would depend on curl again!

Also, if somebody is using prost-types they might not necessarily want to use prost-build for a variety of reasons, e.g. because prost-build doesn't build on their particular platform (say, SPARC Solaris, just for argument's sake) whereas prost-types would work fine otherwise.

briansmith:
OK, this seems to be working and is ready to be reviewed.

* I changed the directory and naming structure of the imported executables because that simplified things.
* I didn't include the 32-bit macOS executable since 32-bit executables on macOS are being deprecated (by Apple).
* I only copy the executable for the current host platform to OUT_DIR, matching the previous behavior of prost-build.
* I didn't try to do anything with prost-types like you suggested. I liked the idea you suggested, but I couldn't figure out how to make it actually work, and regardless, refactoring prost-types is really outside the scope of the issue that this PR is attempting to solve.

prost-build/build.rs
```diff
@@ -1,44 +1,54 @@
...
-    protoc.push("protoc");
-    protoc.set_extension(env::consts::EXE_EXTENSION);
+    let src_dir = PathBuf::from("../third-party/protobuf");
+    static INCLUDE: &str = "include";
+
+    let protoc_bin_name = match (env::consts::OS, env::consts::ARCH) {
+        ("linux", "x86") => "protoc-linux-x86_32",
+        ("linux", "x86_64") => "protoc-linux-x86_64",
+        ("linux", "aarch64") => "protoc-linux-aarch_64",
+        ("macos", _) => "protoc-osx-x86_64",
```
danburkert: Will an x86_64 binary run on a `macos` 32bit system?  I'm behind the decision to drop 32bit support, but I think it'd be better to have this panic fire in that case, since it should be easier to debug than whatever error is presented when trying to execute 64bit binary on a 32bit system.

danburkert: to be clear, I'm wondering if the match on line 24 should be `("macos", "x86_64")`.

prost-build/build.rs
```diff
@@ -1,44 +1,54 @@
...
+        ("windows", _) => "protoc-win32.exe",
+        _ => panic!("no precompiled protoc binary for the current platform: {}-{}",
+                    env::consts::OS, env::consts::ARCH),
+    };
+
+    let dest_dir = {
+        let out_dir = PathBuf::from(env::var("OUT_DIR")
```
danburkert: Should this be `var_os`, as you pointed out in the other PR?

third-party/protobuf/readme.txt
```diff
@@ -0,0 +1,12 @@
+`LICENSE` was taken from https://github.com/google/protobuf/archive/v3.5.1.zip.
+
+The `protoc-*` and `protoc-*.exe` binaries here were taken from:
+
+* https://github.com/google/protobuf/releases/download/v3.5.1/protoc-3.5.1-linux-aarch_64.zip
+* https://github.com/google/protobuf/releases/download/v3.5.1/protoc-3.5.1-linux-x86_32.zip
+* https://github.com/google/protobuf/releases/download/v3.5.1/protoc-3.5.1-linux-x86_64.zip
+* https://github.com/google/protobuf/releases/download/v3.5.1/protoc-3.5.1-osx-x86_32.zip
```
danburkert: remove this line

Issue #80, opened 2018-02-26
Consider generating idiomatic enum value names


warrenfalk:
Idiomatic protocol buffer definitions for enums is to prefix the enum value names with the name of the enum.  This is because in protocol buffers, the enum value names are in the same namespace as the enum itself (which means you aren't allowed to have two different values named ```GOOD``` even if they are in different enums.  And therefore this is common and idiomatic in protocol buffers:

```protobuf
enum ResultCode {
    RESULT_CODE_GOOD = 0;
    RESULT_CODE_BAD = 1;
    RESULT_CODE_UGLY = 2;
}
```

Rust doesn't suffer from this unfortunate design and so if you were to code the same thing in idiomatic Rust, it would look like this:

```rust
enum ResultCode {
    Good = 0,
    Bad = 1,
    Ugly = 2,
}
```

Not like this:

```rust
enum ResultCode {
    ResultCodeGood = 0,
    ResultCodeBad = 1,
    ResultCodeUgly = 2,
}
```

And you'll find that all of the newer "official" protocol buffer generators do this (take C# for example). 
They will detect if the enum value names are prefixed with the enum's name and strip it off if so.

I've already written this behavior for Prost, controlled by config and will send a PR in case you're interested.  Currently it defaults to existing behavior, but I think you might want to consider making idiomatic Rust the default behavior.  For those who don't prefix their protobuf enum value names, well they won't notice either way, and for those that do, they'll probably appreciate the improvement to their code.

briansmith:
Rather than having an option to control this, maybe we should just do this by default in a new major version of prost.

danburkert:
This was by @warrenfalk in a87d216f1190cebce507ebd1bdb48fe91e652c0b.

PR #81, opened 2018-02-26
prost-build: add Config option, use_idiomatic_enum_values

warrenfalk:
Fix for #80 

When enabled, this will attempt to create idiomatic Rust enum variants
from idiomatic protocol buffer enum value names by detecting value
names which are prefixed with the enum name and removing the prefix.

For now this defaults to false (previous behavior) and must be
explicitly enabled. This is for backward compatibility, but enabling by
default should be considered as this won't affect any who aren't
prefixing their protocol buffer enum value names, and will probably
be welcomed by most who are.

### DIFF
--- a/prost-build/src/lib.rs
+++ b/prost-build/src/lib.rs
@@ -199,6 +199,7 @@ pub struct Config {
     type_attributes: Vec<(String, String)>,
     field_attributes: Vec<(String, String)>,
     prost_types: bool,
+    idiomatic_enum_values: bool,
 }
 
 impl Config {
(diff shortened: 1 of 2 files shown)


prost-build/src/lib.rs
```diff
@@ -352,6 +353,11 @@ impl Config {
         self
     }
 
+    pub fn use_idiomatic_enum_values(&mut self, idiomatic: bool) -> &mut Self {
```
danburkert: Add a method doc about what this does; the prose from the commit message does a good job explaining it.

PR #82, opened 2018-03-01
Disable unneeded features of the petgraph library.

briansmith:
This drops the `ordermap` dependency, for example.


danburkert:
Thanks!

PR #85, opened 2018-03-04
Move third-party/ to prost-build

danburkert:
This is a followup / fix for e0317f83958892d716e99423f07525db5c7469e6.

While attempting to publish prost-build 0.3.1 I discovered something
interesting about Cargo: during the 'package' and 'publish' step it will
execute the build from a completely different OUT_DIR directory than
during a normal build. As a result, relative paths from a sub-crate up
to the top level workspace are broken. However, this also means that
Cargo will _automatically_ copy all files in the crate to the build
directory.  We can take advantage of this to remove all of the copying
logic that had been added to `prost-build/build.rs` by simply moving
the `third-party` directory into `prost-build`. Another side-effect of
this Cargo behavior is that the published version of `prost` 0.3.1 is
7MiB because it embeds the `protoc` binaries, even though it doesn't
need them. So with this change `prost` should go back to the previous
small size, and instead `prost-build` will get the embedded binaries.


briansmith:
LGTM! Thanks for fixing this. I had tested 0.3.1 by using a cargo `[patch.crates.io]` to override my project's crates.io-based dependencies on prost with ones using the git master branch and it worked fine. I didn't realize it would break when we actually switched to the crate packaging.

danburkert:
Yeah this was definitely a surprise for me as well.

PR #87, opened 2018-03-07
Remove unused failure_derive dependency.

briansmith:
Remove the indirect and unneeded dependency on failure_derive. In our
prost-based application, this change removes the following dependencies
from our Cargo.lock (partial output from `cargo update -p prost-derive`):

    Removing failure_derive v0.1.1
    Removing quote v0.3.15
    Removing syn v0.11.11
    Removing synom v0.11.3
    Removing synstructure v0.6.1
    Removing unicode-xid v0.0.4

Note in particular that prost itself is using a newer version of `syn` and
`quote` than `failure` is using.

### DIFF
--- a/prost-derive/Cargo.toml
+++ b/prost-derive/Cargo.toml
@@ -13,7 +13,7 @@ description = "A Protocol Buffers implementation for the Rust Language."
 proc_macro = true
 
 [dependencies]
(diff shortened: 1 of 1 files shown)


danburkert:
Thanks!

PR #90, opened 2018-04-01
Update dependencies

hcpl:
`def_site` was replaced by `call_site` because new `proc-macro2` doesn't expose it anymore. Besides, `prost-derive` only used spans for literals which shouldn't depend on span information at the same degree as identifiers.


danburkert:
This was fixed by #89, but thanks for the PR!

PR #92, opened 2018-04-07
prost-derive: infer tag values of hand-written types

danburkert:
This contains the tag inference logic extracted from #73.


danburkert:
Corrected.

PR #104, opened 2018-05-30
Update dependencies for `prost-derive`

hcpl:
Use patch number-sensitive version requirements because of the deprecated API present in `proc-macro2` 0.4.0..0.4.3 and both `quote` 0.6.0..0.6.2 and `syn` 0.14.0 use that API.


danburkert:
Thanks!

PR #110, opened 2018-06-14
Update minimum version of bytes.

adeschamps:
The get_*_be and get_*_le variants are not present in 0.4.6 and below.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ exclude = [
 
 [dependencies]
 byteorder = "1"
-bytes = "0.4"
+bytes = "0.4.7"
 
 [dev-dependencies]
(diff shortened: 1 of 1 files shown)


gsquire:
I think you can just run `cargo update -p bytes` to get the latest patch version. Specifying "0.4" in the manifest will allow 0.4.7 to be pulled in based on the [caret requirements](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#caret-requirements) of the documentation.

danburkert:
Looks good, thanks!

adeschamps:
@gsquire that's true, it's not too hard to deal with. I had an existing dependency on bytes 0.4.6, so updating prost didn't cause bytes to be updated.

PR #111, opened 2018-06-14
Extend known types in prost-build

ebkalderon:
### Added

* Add `known_type()` to `CodeGenerator`.
* Add `map_type()`, `map_types()`, and `mapped_types` field to `Config`.

### Changed

* Return a `String` instead of `Cow` in `resolve_type()` due to lifetime constraints.

---

### Rationale

Let's say that you have two Rust crates which depend on Prost. The first crate contains the following message:

```proto
// first_crate/uuid.proto 

syntax = "proto3";
package first_crate;

message Uuid {
    string uuid_str = 1;
}
```

The second crate contains the following message:

```proto
// second_crate.proto

syntax = "proto3";
package second_crate;

import "first_crate/uuid.proto";

message MyMessage {
    first_crate.Uuid message_id = 1;
    string some_payload = 2;
}
```

As you can see, the Protobuf message in the second crate refers to the `Uuid` message defined in the first crate. In order for the Protobuf files in the second crate to compile successfully, we add the Protobuf files from the first crate as an include in `compile_proto()`.

When the build script gets run for each crate, Prost fails to see that the `Uuid` message type in the first crate and the `Uuid` message type in the second crate are the same type, and generates two incompatible Rust versions for both. This causes problems down the line: we can't compare them, implement traits on them, and otherwise use them interchangeably even though they're fundamentally the same type. For example, the following code does not compile:

```rust
// `lib.rs` of `first_crate`

extern crate prost;
#[macro_use]
extern crate prost_derive;
extern crate prost_types;

include!(concat!(env!("OUT_DIR"), "/uuid.rs"));

pub trait DoSomething {
    fn do_it();
}

impl DoSomething for Uuid {
    fn do_it() {
        println!("Done");
    }
}
```

```rust
// `lib.rs` of `second_crate`

extern crate first_crate;
extern crate prost;
#[macro_use]
extern crate prost_derive;
extern crate prost_types;

use first_crate::{DoSomething, Uuid};

include!(concat!(env!("OUT_DIR"), "/second_crate.rs"));

pub fn process_message(msg: MyMessage) {
    if let Some(uuid) = msg.message_id {
        uuid.do_it();  // Error! `first_crate::DoSomething` not implemented for `second_crate::Uuid`.
    }
}
```

To solve this, we added two new methods to `prost-build::Config`: `map_type()` and `map_types()`. The `map_type()` method replaces the `Uuid` message type in the second crate with the `Uuid` message type from the first crate so that the two can be used interchangeably. The `map_types()` method works similarly, taking in a set of types to be replaced. In the build script of `second_crate`, we would call `map_type()` like so:

```rust
extern crate prost_build;

fn main() {
    prost_build::Config::new()
        .map_type(".first_crate.Uuid".into(), "::first_crate::Uuid".into())
        .compile_protos(
            &["./proto/second_crate.proto"],
            &["../../first_crate/proto"],
        )
        .unwrap()
}
```

With these changes, the example code above will compile successfully and behave as expected.


danburkert:
This looks really useful, thanks for the PR!  It's going to take me a bit to get around to a full review, but just wanted to let you know that I'm very positive on the premise.  It's a little bit like 'extern' declarations.

danburkert:
Sorry for taking so long on this, but I wanted to take this feature a bit further, and it took me longer than I expected.  I didn't want to push to y'alls branch so I made a new PR on #122 which builds on this.  Would y'all take a look at that and see if it meets your needs?

ebkalderon:
Sure, we'll gladly take a look at it. Thanks for building on this feature!

ebkalderon:
@danburkert Looks great! We left our comments on the other PR.

PR #117, opened 2018-07-13
Keep unknown fields in decode/encode roundtrips

per-gron:
This is a work-in-progress PR for closing #2. I would appreciate a preliminary review of the approach before I spend more time trying to fix more corner casey things.

Remaining things to fix:

* [ ] Make the generation of the `unknown_fields` field of generated structs optional. Should it be on or off by default? I think it should be on by default, at least eventually because otherwise the library is non-conformant by default.
* [ ] Decide what to do with well-known types like BoolValue, UInt32Value etc. They are currently automatically converted to native types which are not able to store unknown fields. To me this auto-conversion seems unnecessary and it might make sense to turn this off by default.
* [ ] Decide what to do with the entry messages of maps; when they are stored as Rust maps they don't retain unknown fields. I think the options here are to have custom map types or to accept that unknown fields there will be discarded. I think for this case the convenience of native Rust maps might be better than keeping unknown fields in this particular case.
* [ ] Decide how to handle known fields of the wrong type. [Relevant ticket](https://github.com/google/proto-lens/issues/125).
* [ ] Decide how to handle oneof fields with more than one set field.
* [ ] Preserve groups (can probably be done in a later PR since it is not supported at all atm).
* [ ] Handle the case of proto messages with a field called `unknown_fields` (the generated field should get a different name then).
* [ ] Address TODOs in the code.

### DIFF
--- a/conformance/failing_tests.txt
+++ b/conformance/failing_tests.txt
@@ -1,3 +1,0 @@
-# TODO(danburkert/prost#2): prost doesn't preserve unknown fields.
-Required.Proto2.ProtobufInput.UnknownVarint.ProtobufOutput
-Required.Proto3.ProtobufInput.UnknownVarint.ProtobufOutput
--- a/conformance/tests/conformance.rs
+++ b/conformance/tests/conformance.rs
@@ -17,7 +17,6 @@ fn test_conformance() {
 
     let status = Command::new(env!("CONFORMANCE_TEST_RUNNER"))
                          .arg("--enforce_recommended")
-                         .arg("--failure_list").arg("failing_tests.txt")
                          .arg(proto_conformance)
                          .status()
                          .expect("failed to execute conformance-test-runner");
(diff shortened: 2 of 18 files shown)


timthelion:
This could be much more simply implemented, without breaking changes, by adding two new methods to the `Message` trait: `encode_with_unknowns` and `decode_with_unknowns`.

- `encode_with_unknowns` would take an extra parameter which would be vector of bytes which would consist of the non-decoded parts of the message.
- `decode_with_unknows` would return a Result in which the first value was a tuple, `<Self, Vec<u8>>`. 

Users of the library would then have the choice of preserving unknowns, without any performance penalty if they did not wish to, and without any breaking changes :)

danburkert:
Keeping the unknowns as an external value is an interesting idea.  I think
it’s a bit more complex than just a vector of bytes since nested messages
can have unknown fields, but a nested tree structure could be used.

timthelion:
Aha, you are right, and a problematic case just occurred to me. How does one deal with repeated fields of messages. How would one know which message in the list to re-apply the old defaults too :/

carllerche:
Thanks for the work :+1:. I'm going to close this due to inactivity. If this needs to be revisited, a new PR will be best (reference this one).

PR #122, opened 2018-08-02
Extern path

danburkert:
This builds on #111, solving the same usecase but in a slightly more general way (entire packages can be mapped to Rust modules).  I also took the opportunity to clean up the special cased well-known-types handling using the new feature.

@ebkalderon @trha would you mind taking a look at this API?  Does it meet y'alls needs?


ebkalderon:
@danburkert Is this planned to be released sometime soon? We would really like to have this feature available on Crates.io, if possible.

PR #133, opened 2018-11-08
Add Default supertrait for Message

KiChjang:
Since prost-derive already implements Default for the messages, I think we should also require Default to be implemented for implementors of Message.


KiChjang:
OK, I found another way of solving my problem, thanks anyway!

Issue #31, opened 2017-09-11
optimize for protos containing large blobs 


rajatgoel:
One small optimization prost can do for proto message containing large blobs is - specialize decoding protobuf from reference counted buffers and give a reference to a slice of the underlying buffer as bytes instead of copying them. Similarly, when encoding the message simply chain these big blobs instead of copying them to output buffer. I think the C++ does this optimization for all `bytes` fields that specify `ctype=cord`.

danburkert:
There's no support for this yet, but it should be possible to add.  The 'plan' is to:

* Add an option to `prost_build::Config` to output `string` and `bytes` fields as `bytes::Bytes` instead of `String` and `Vec<u8>`.
* Add specialization into the decoding stack so that when decoding into a `bytes::Bytes` field from a `bytes::Bytes` source, it does a shallow copy (`bytes::Bytes` is ref-counted).

This solution unfortunately isn't too general - it only works with types known to `prost` through explicit specializations, but I think `bytes::Bytes` will cover the majority of cases.  Perhaps if there's a rope data structure that gains prominence down the road that could be added too.

nerdrew:
Has anyone started on this? What does "specialization in the decoding stack" look like?

vorner:
Looking at it, using `Bytes` instead of `String` feels a little wrong, doesn't it? I mean, something that checks the unicode and derefs into `&str` would be nice.

Or is the plan to provide some wrapper?

quininer:
@vorner I think [`String<Bytes>`](https://github.com/carllerche/string) is what you want.

nrc:
Using `Bytes` for `bytes` would be useful for small bytes fields too - at the moment we spend a huge amount of time allocating small Vecs for such fields. Being able to reference the underlying buffer, or even allocate a large slab of memory once would be a big performance win.

mzabaluev:
As a container for string buffers backed by `Bytes`, I have published the [strchunk crate](https://crates.io/crates/strchunk).

vorner:
What is the advantage of using that over that [`String<Bytes>`](https://docs.rs/string/0.2.1/string/struct.String.html) thing above?

mzabaluev:
@vorner I did not look at it in detail, but it seems, as a more generic container, it can't make use of some optimizations that are available for the specific `Bytes`-backed container. On the other hand, the `TakeRange` impls could be implemented for `String<Bytes>` as well, since I have already implemented them for `Bytes`.

[...]

PR #135, opened 2018-11-13
WIP: Specialization of decoding for Bytes

vorner:
I don't know if I understood #31 suggestion correctly, so I'd like to ask about it. I think it could work somehow like this:

* There would be a special method for decoding from bytes, that would default to just decoding from `Buf`.
* Generated `Message` instances would implement it similar as they implement the basic decoding.
* While most values can use the default, the newly supported `Bytes` would get a specialization and just use the part of the original `buf`.

While this probably *would* work, the in here looks like a big kludge to me. So did you have something else in mind?

I guess specialization (or how the feature is called) would greatly help, but it's not coming soon into Rust, is it?

### DIFF
--- a/src/message.rs
+++ b/src/message.rs
@@ -1,7 +1,7 @@
 use std::fmt::Debug;
 use std::usize;
 
-use bytes::{Buf, BufMut, IntoBuf};
+use bytes::{Buf, BufMut, Bytes, IntoBuf};
 
 use DecodeError;
 use EncodeError;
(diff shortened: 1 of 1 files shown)


danburkert:
@vorner there shouldn't be a special method on `Message` for decoding from `Bytes`; specialization makes that unnecessary.  To expand on the second bullet point in #31 specifically, there should be a new trait for `string` and `bytes` fields which allows for specialized decoding when the input buffer is a `Bytes`.  That way you shouldn't need to change the call sites doing decoding at all, it should 'just work' as long as the message type has zero-copy configured fields, and the buffer is a `Bytes` instance.

danburkert:
Closing this out per the comment I left in #31, please feel free to file a new issue if the solution in master doesn't work for you.  Thanks!

PR #143, opened 2019-01-07
Replace tempdir dependency with tempfile

hawkw:
`prost-build` currently depends on the the `tempdir` crate, which is
deprecated. Users of `tempdir` are encouraged to migrate to `tempfile`. 

This branch replaces all the uses of `tempdir` in PROST with `tempfile`.

Signed-off-by: Eliza Weisman <eliza@buoyant.io>


hawkw:
The motivation behind this change is that Linkerd 2's dependencies are currently pulling in multiple incompatible versions of the `rand` crate (see https://github.com/linkerd/linkerd2-proxy/pull/169#discussion_r245458161), and we'd like to only have to download and build one version of `rand`. Since `tempdir` isn't actively maintained, it still depends on `rand` 0.4. Updating `prost` to depend on `tempfile` instead allows us to remove the `rand` 0.4 dependency.

PR #149, opened 2019-02-09
Add comment explaining behavior of add_message in prost_build

### LINKED ISSUE #148: Question: Why are repeated field relations not added to the message graph?
I see that when building the message graph in the compiler code, repeated fields are ignored. Why?

https://github.com/danburkert/prost/blob/master/prost-build/src/message_graph.rs#L52

timthelion:
Fixes #148


danburkert:
Thanks!

PR #153, opened 2019-02-17
Fully-qualify type names from `std` in codegen

jeffparsons:
Specifically `Vec` and `String`.

Imagine you have a protobuf message type define like this:

```proto
message String {
    option (a_private_extension_namespace.boxed_message).is_presence_box = true;
    optional string value = 1;
}
```

Previously the Rust code generated by prost-build looks like this:

```rust
pub struct String {
    #[prost(string, optional, tag="1")]
    pub value: ::std::option::Option<String>,
}
```

The inner reference to `String` above should be referring to `std::string::String`,
but that is shadowed by the local definition of `String`. Fully-qualifying all `std`
types in codegen solves this.

Fixes https://github.com/danburkert/prost/issues/152


danburkert:
Thanks!

PR #155, opened 2019-02-25
Permit using `FileDescriptorSet` as an intermediate compilation artifact 

nrc:
Currently based on #147.

Motivation for doing this is so we can pre-compile the FileDescriptorSet and also so that the set can be inspected to find the package name and thus the name of the generated file.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "prost"
 # NB: When modifying, also modify html_root_url in lib.rs
-version = "0.4.0"
+version = "0.5.0"
 authors = ["Dan Burkert <dan@danburkert.com>"]
 license = "Apache-2.0"
 repository = "https://github.com/danburkert/prost"
@@ -40,7 +40,7 @@ criterion = "0.2"
 env_logger = { version = "0.5", default-features = false }
 failure = "0.1"
 log = "0.4"
-prost-derive = { version = "0.4.0", path = "prost-derive" }
+prost-derive = { version = "0.5.0", path = "prost-derive" }
 protobuf = { path = "protobuf" }
 quickcheck = "0.6"
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,7 @@ readme = "README.md"
 description = "A Protocol Buffers implementation for the Rust Language."
 keywords = ["protobuf", "serialization"]
 categories = ["encoding"]
+edition = "2018"
 
 [badges]
 travis-ci = { repository = "danburkert/prost" }
--- a/README.md
+++ b/README.md
@@ -326,6 +326,16 @@ pub enum Gender {
   But it is possible to place `serde` derive tags onto the generated types, so
   the same structure can support both `prost` and `Serde`.
 
+2. **I get errors when trying to run `cargo test` on MacOS**
+
+  If the errors are about missing `autoreconf` or similar, you can probably fix
+  them by running
+
+  ```
+  brew install automake
+  brew install libtool
+  ```
+
 ## License
 
 `prost` is distributed under the terms of the Apache License (Version 2.0).
--- a/README.md
+++ b/README.md
@@ -338,11 +338,9 @@ pub enum Gender {
 
 3. **I'm getting compile errors when compiling a 2018 edition project**
 
-  If you're using prost-build, you need to use the `build-2018` feature:
+  If you're using prost-build, you will need to set `edition` to `Rust2018` in
+  the `Config`.
 
-  ```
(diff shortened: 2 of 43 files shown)


danburkert:
@nrc can you give an example of where this is additional API is useful?

nrc:
> @nrc can you give an example of where this is additional API is useful?

Sure, one motivation is pre-compiling protobufs to file descriptors, so the user can generate custom Rust code, without needing protoc installed.

Another is knowing which packages have been compiled and therefore which files have been created. We use this because for every Rust file Prost generates, we want to read that in and generate another file with more generated code.

nrc:
@danburkert ping - does the above sounds reasonable? If so I can rebase this PR.

danburkert:
@nrc I'm not (yet) convinced that this would be a generally useful feature - obviously I don't know all the details, but it seems like this is a workaround for a missing feature, but it's not clear whether that's something like a generic message API or protoc plugin equivalent.

It should be possible to do this relatively easily out-of-tree, since all of `prost_build::protoc()`, `prost_build::protoc_include()` and `prost-types::FileDescriptorSet` are public (in other words, I'm pretty sure just copy/pasting `generate_descriptor_set()` into a downstream application should work).

nrc:
OK, I'll try and do it out of tree.

I get what you mean about this not feeling quite right, but I can't think of anything better right now. It does seem to me that the `FileDescriptorSet` is a good intermediate artifact since it is part of the protobuf spec, not just something Prost-specific. FWIW, it is also used as an intermediate artifact by rust-protobuf.

Could you make a 0.5.0 release with the 2018 changes (and other stuff), please?

danburkert:
> Could you make a 0.5.0 release with the 2018 changes (and other stuff), please?

Done!

nrc:
Thanks!

lucab:
> Sure, one motivation is pre-compiling protobufs to file descriptors, so the user can generate custom Rust code, without needing protoc installed.

Just chipping in that we have a similar usecase, where we try not to require `protoc` everywhere. If this ends up making the prost codegen artifacts easier to cache/commit, that would be helpful.

nrc:
> It should be possible to do this relatively easily out-of-tree

It turned out not to be possible since `Config::generate` is private. IMO, it makes more sense to add a method that can do the complete compilation from `FileDescriptorSet` than it does to make `generate` pubic (since `generate` does a non-intuitive subset of compilation).

nrc:
And you can't copy/paste `generate` because it uses a bunch of private stuff.

danburkert:
@nrc I may be misunderstanding - are you modifying the `FileDescriptorSet` and then trying to convert that into Rust code?   I was originally assuming you weren't, so the workaround could be copy/pasting `generate_descriptor_set` into your application, and separately doing a full compilation from .proto to .rs through the existing API.

nrc:
I'm not modifying `FileDescriptorSet`. So for one use case (just getting info from `FileDescriptorSet`), that is a reasonable workaround (albeit, it does a full run of protoc twice). For our second use case we want to generate the `FileDescriptorSet` offline and commit it to a repo, then build the Rust code from the binary `FileDescriptorSet` so that the user does not have to have protoc present. For that use case we need some way to generate Rust code from `FileDescriptorSet`s, which it seems can't be done at the moment.

So, (with the workaround) I am not stuck right now, but it would be great to have *some* API for building from a `FileDescriptorSet` in the near-ish future so that we can avoid the double run of protoc and can pre-compile the `FileDescriptorSet`.

danburkert:
I see, thanks for the detailed response.  In case you aren't aware, `prost-build` ships with `protoc` embedded, so unless your users are on something other than x86 windows/linux/macos or aarch64 linux you shouldn't need to worry about `protoc` being available.

In terms of moving forward on this, I'm not against the idea of exposing APIs to work with `FileDescriptorSets` as an intermediate build artifact, but it still seems like it may not really be the best fit for the use-cases at hand.  Namely, if caching of the `FileDescriptorSets` is the goal then you'd probably also want to put `prost-build`'s bundling of `protoc` behind a feature flag.  This is unfortunately not trivial, I think to do it properly the individual executable would need to be split into their own crates.

And for the `FileDescriptorSets` processing, I've typically been pushing people towards creating custom derives and injecting them into the generated code, but perhaps you need something more flexible?  On the simpler end of the spectrum It's also possible to add methods to generated types via `impl MyMessageType { fn frobulate() { .. } }` in the crate which is including the prost generated code ([example](https://github.com/danburkert/prost/blob/master/prost-types/src/lib.rs)).

nrc:
Closing for now, since we can work around

PR #156, opened 2019-02-27
Add `repr(i32)` to enums

nrc:
To guarantee that their representation matches the assumed i32, allowing clients to safely transmute.

r? @danburkert


danburkert:
@nrc is it not possible to do `my_enum_value as i32`?  That has the advantage of not requiring unsafe.  My only concern with adding `repr(i32)` is that iiuc it unnecessarily pessimizes the common case where an enum has just a few values, and otherwise could be only 1 or 2 bytes layout?

danburkert:
wait that's not an issue, enums are always an `i32` field in a message anyway.

PR #159, opened 2019-03-03
Add Rust 2018 edition support

danburkert:
Notable changes:

* `prost` crates are now edition 2018
* Code generation is now compatible with edition-2018 crates, as well as
  edition-2015 crates. No explicit configuration is necessary to enable
  edition-2018 support.
* Reserved field names now use raw identifiers instead of adding an
  underscore.
* The tests crate now executes for both 2018 and 2015 edition via
  separate crates with a symlink between the src/ directory.

This is essentially @nrc's work in #147, but with some simplifications and additional tests.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,7 @@ readme = "README.md"
 description = "A Protocol Buffers implementation for the Rust Language."
 keywords = ["protobuf", "serialization"]
 categories = ["encoding"]
+edition = "2018"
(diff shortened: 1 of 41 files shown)


danburkert:
The tests-2015 crate is failing to compile on windows, presumably because the symlink doesn't work.  Will need to research a fix.

nrc:
\o/

PR #167, opened 2019-03-13
[WIP] wasm-bindgen support

Tarnadas:
Hey!

I am trying to add wasm-bindgen support and stumbled across several problems. This PR is still very wip.

- I don't know how to define a feature for a sub-crate for conditional compilation. See commented out cfg! code
- wasm-bindgen does not support pub vectors within structs, so I just removed the pub for them (see https://github.com/rustwasm/wasm-bindgen/issues/439)
- probably the biggest problem so far: oneof gets converted to enums holding data, which isn't supported by wasm-bindgen and it will probably take a lot of effort to support them within wasm-bindgen (see https://github.com/rustwasm/wasm-bindgen/issues/31). It might be easier to change code generation of proto files for oneofs instead

### DIFF
--- a/prost-build/Cargo.toml
+++ b/prost-build/Cargo.toml
@@ -19,9 +19,14 @@ petgraph = { version = "0.5", default-features = false }
 prost = { version = "0.6.1", path = "..", default-features = false }
 prost-types = { version = "0.6.1", path = "../prost-types", default-features = false }
 tempfile = "3"
+wasm-bindgen = { version = "0.2", optional = true }
 
 [build-dependencies]
 which = { version = "4", default-features = false }
 
 [dev-dependencies]
 env_logger = { version = "0.8", default-features = false }
+
+[features]
+default = []
+use-wasm-bindgen = []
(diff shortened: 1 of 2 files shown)


danburkert:
Hi sorry for slow feedback - is this still a going concern?  I would ideally like `prost` to be compatible with wasm, but the concerns about enums with data seems like a pretty hard blocker.  For context I have no wasm experience, so I don't know if the state of the world has changed in the last 18 months.

Tarnadas:
Hey,
it doesn't look like anything has changed regarding enums, so a workaround for a WebAssembly version would be required.

I was rather new to Rust when I started this PR and I think I might get this working somehow.
As a reference, I will see how other Protobuf libraries in JavaScript do their code generation and try to make something similar for the WebAssembly version.

ImJeremyHe:
It is better to do this work after `wasm-bindgen` supporting enums: https://github.com/rustwasm/wasm-bindgen/pull/2631

caspermeijn:
@Tarnadas Are you still working on this, or can it be closed?

Tarnadas:
Hey, it can be closed

Issue #16, opened 2017-07-04
Support for `groups`


iduartgomez:
While groups where deprecated and never made it into the official version of Protobuf there still are used in some codebases (by Google nonetheless, why, I don't know, probably due to legacy issues) because the original C++ and Python implementations can deal with them fine. AFAIK none of the Rust implementations support Groups properly (which can be understood for the reason stated before) but I wonder if there is interest in adding support for them.

My personal use case: I'm trying to implement/partially port a natural language processing model from Google which uses Tensorflow and some custom C++ extensions. Protobuf is used extensively in this settings, but some of the defined messages use "Group". Refactoring the code to not use groups would be a major pain as it permeates a big chunk of the codebase in this case.

`prost` does not seem to work outright with groups, it generated the code just fine, but when it comes to compile the library it will fail to derive type information correctly.

An example below:
protobuf file:
```
// LINT: ALLOW_GROUPS

syntax = "proto2";

package syntaxnet;

// Task input descriptor.
message TaskInput {
  // Name of input resource.
  required string name = 1;

  // Name of stage responsible of creating this resource.
  optional string creator = 2;

  // File format for resource.
  repeated string file_format = 3;

  // Record format for resource.
  repeated string record_format = 4;

  // Is this resource multi-file?
  optional bool multi_file = 5 [default = false];

  // An input can consist of multiple file sets.
  repeated group Part = 6 {
    // File pattern for file set.
    optional string file_pattern = 7;

    // File format for file set.
    optional string file_format = 8;

    // Record format for file set.
    optional string record_format = 9;
  }
}
```

generated code:
```
/// Task input descriptor.
#[derive(Clone, Debug, PartialEq, Message)]
pub struct TaskInput {
    /// Name of input resource.
    #[prost(string, required, tag="1")]
    pub name: String,
    /// Name of stage responsible of creating this resource.
    #[prost(string, optional, tag="2")]
    pub creator: Option<String>,
    /// File format for resource.
    #[prost(string, repeated, tag="3")]
    pub file_format: Vec<String>,
    /// Record format for resource.
    #[prost(string, repeated, tag="4")]
    pub record_format: Vec<String>,
    /// Is this resource multi-file?
    #[prost(bool, optional, tag="5")]
    pub multi_file: Option<bool>,
    #[prost(group, repeated, tag="6")]
    pub part: Vec<task_input::Part>,
}
pub mod task_input {
    /// An input can consist of multiple file sets.
[...]

PR #168, opened 2019-03-13
Add group support

hanya:


hanya:
Related to #16.

danburkert:
Hi @hanya, thanks for the PR!  After taking a look at this implementation I was curious whether group support could be added without adding a new `Group` trait, and instead simply treat groups as messages.  I wasn't sure whether this was possible, so I took your PR and played with it till I got something working!  So I'm going to open a new PR shortly based on this one.  Thanks again.

PR #185, opened 2019-05-19
Fix well-known-type enum field compilation error

danburkert:


sdbondi:
Any chance this fix will be released on crates?

PR #186, opened 2019-05-23
Implement a recursion limit

### LINKED ISSUE #176: Add a recursion limit for decoding
Rust-protobuf and protobuf implementations in other languages support a recursion limit for decoding. Prost decoding is also recursive, and I believe supporting a recursion limit is necessary if messages can be defined recursively (as they can be for TiKV). Could we add this limit to Prost? I can implement this if it would be accepted, but I'm not sure where to put the option - currently Prost does not support any options for decoding/encoding (as far as I can tell). Rust-protobuf adds the recursion limit to the buffer, but because Prost uses a trait for the buffer that seems not the best solution (I think it would work, we would have a recursion-limited buffer which wrapped any implementation of buffer like a smart pointer, but it would be pretty hacky and I think might be inefficient).

The Rust-protobuf solution was added in https://github.com/stepancheg/rust-protobuf/pull/248

nrc:
This PR implements a recursion limit, which can be disabled by not setting a Cargo feature (which is on by default). This change is technically not backwards compatible, but in practice very few programs should hit the recursion limit.

Although adding a context argument complicates the API somewhat, most clients would not implement or use the methods with the added argument and would instead only use `decode`, which is not changed.

Fixes #176 

r? @danburkert

### DIFF
--- a/src/encoding.rs
+++ b/src/encoding.rs
@@ -77,15 +77,17 @@ where
     }
 }
 
+/// Additional information passed to every decode/merge function.
 #[derive(Clone, Debug)]
 pub struct DecodeContext {
     #[cfg(feature = "recursion-limit")]
+    /// How many times we can recurse in the current decode stack before we hit
+    /// the recursion limit.
+    ///
+    /// The recursion limit is defined by `RECURSION_LIMIT` and cannot be
+    /// customized. The recursion limit can be ignored by building the Prost
+    /// crate without the `recursion-limit` feature (which is set by default).
     recurse_count: u32,
-}
-
-pub(crate) struct RecursionGuard {
-    #[cfg(feature = "recursion-limit")]
-    ctx: *mut DecodeContext,
 }
 
 impl Default for DecodeContext {
@@ -95,23 +97,21 @@ impl Default for DecodeContext {
             recurse_count: crate::RECURSION_LIMIT,
         }
     }
+
     #[cfg(not(feature = "recursion-limit"))]
     fn default() -> DecodeContext {
         DecodeContext {}
     }
 }
 
-#[cfg(feature = "recursion-limit")]
-impl Drop for RecursionGuard {
-    fn drop(&mut self) {
-        unsafe {
-            (*self.ctx).recurse_count += 1;
-        }
-    }
-}
-
 impl DecodeContext {
     #[cfg(feature = "recursion-limit")]
+    /// Call this function before recursively decoding.
+    ///
+    /// This function returns a guard object which will automatically restore
+    /// the recursion counter when it is destroyed by going out of scope.
+    ///
+    /// See the safety note on `RecursionGuard` for important information.
     pub(crate) fn enter_recursion(&mut self) -> RecursionGuard {
         self.recurse_count -= 1;
         RecursionGuard { ctx: self }
@@ -124,8 +124,49 @@ impl DecodeContext {
     }
 }
 
+/// RAII guard created by `DecodeContext::enter_recursion` to ensure recursion is
+/// tracked correctly.
+///
+/// ## Safety note
+///
+/// This object uses a raw pointer and unsafe code to avoid dynamic ownership
+/// checking of it's reference to a `DecodeContext`. This could be implemented
+/// using `Rc` and `RefCell` or closures, but that could be expensive since we
+/// might expect decoding to be in a program's hot path.
+///
+/// Usage is safe under normal usage patterns:
+///
+/// ```ignore
+/// fn foo(..., ctx: &mut DecodeContext) {
+///     let _guard = ctx.enter_recursion();  // `_guard` keeps a mutable reference to ctx.
+///     some_recusive_fn(..., ctx);          // `ctx` is passed to `some_recusive_fn`.
+/// }
+/// ```
+///
+/// In the above scenario, `ctx.recurse_count` must be the same before and after
+/// the call to `some_recusive_fn`. In particular, it must still be valid memory.
+/// `_guard` should not be passed out of `foo` nor should it be stored.
+pub(crate) struct RecursionGuard {
+    #[cfg(feature = "recursion-limit")]
+    ctx: *mut DecodeContext,
+}
+
+#[cfg(feature = "recursion-limit")]
+impl Drop for RecursionGuard {
+    fn drop(&mut self) {
+        unsafe {
+            (*self.ctx).recurse_count += 1;
+        }
+    }
+}
+
 impl RecursionGuard {
     #[cfg(feature = "recursion-limit")]
+    /// Checks whether the recursion limit has been reached in the stack of
+    /// decodes described by the `DecodeContext` at `self.ctx`.
+    ///
+    /// Returns `Ok<()>` if it is ok to continue recursing.
+    /// Returns `Err<DecodeError>` if the recursion limit has been reached.
     pub(crate) fn limit_reached(&self) -> Result<(), DecodeError> {
         unsafe {
             if (*self.ctx).recurse_count == 0 {
@@ -142,6 +183,7 @@ impl RecursionGuard {
         Ok(())
     }
 }
+
 /// Decodes a LEB128-encoded variable length integer from the slice, returning the value and the
 /// number of bytes read.
 ///
--- a/src/encoding.rs
+++ b/src/encoding.rs
@@ -112,55 +112,18 @@ impl DecodeContext {
     /// the recursion counter when it is destroyed by going out of scope.
     ///
     /// See the safety note on `RecursionGuard` for important information.
-    pub(crate) fn enter_recursion(&mut self) -> RecursionGuard {
-        self.recurse_count -= 1;
-        RecursionGuard { ctx: self }
+    pub(crate) fn enter_recursion(&mut self) -> DecodeContext {
+        DecodeContext {
+            recurse_count: self.recurse_count - 1,
+        }
     }
 
     #[cfg(not(feature = "recursion-limit"))]
     #[inline(always)]
-    pub(crate) fn enter_recursion(&mut self) -> RecursionGuard {
-        RecursionGuard {}
+    pub(crate) fn enter_recursion(&mut self) -> DecodeContext {
+        DecodeContext {}
     }
-}
 
-/// RAII guard created by `DecodeContext::enter_recursion` to ensure recursion is
-/// tracked correctly.
-///
-/// ## Safety note
-///
-/// This object uses a raw pointer and unsafe code to avoid dynamic ownership
-/// checking of it's reference to a `DecodeContext`. This could be implemented
-/// using `Rc` and `RefCell` or closures, but that could be expensive since we
-/// might expect decoding to be in a program's hot path.
-///
-/// Usage is safe under normal usage patterns:
-///
-/// ```ignore
-/// fn foo(..., ctx: &mut DecodeContext) {
-///     let _guard = ctx.enter_recursion();  // `_guard` keeps a mutable reference to ctx.
-///     some_recusive_fn(..., ctx);          // `ctx` is passed to `some_recusive_fn`.
-/// }
-/// ```
-///
-/// In the above scenario, `ctx.recurse_count` must be the same before and after
-/// the call to `some_recusive_fn`. In particular, it must still be valid memory.
-/// `_guard` should not be passed out of `foo` nor should it be stored.
-pub(crate) struct RecursionGuard {
-    #[cfg(feature = "recursion-limit")]
-    ctx: *mut DecodeContext,
-}
-
-#[cfg(feature = "recursion-limit")]
-impl Drop for RecursionGuard {
-    fn drop(&mut self) {
-        unsafe {
-            (*self.ctx).recurse_count += 1;
-        }
-    }
-}
-
-impl RecursionGuard {
     #[cfg(feature = "recursion-limit")]
     /// Checks whether the recursion limit has been reached in the stack of
     /// decodes described by the `DecodeContext` at `self.ctx`.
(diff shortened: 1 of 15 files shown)


nrc:
Design question: it occurs to me that this could be made simpler by passing the context by value and have `enter_recursion` create a new context with a decremented counter, rather than modifying the current context. Then we don't need to worry about incrementing the counter when we return (so don't need the RAII object, etc). However, if we add anything to the context in the future, we will have to use `Rc<RefCell>` and pay a cost of cloning, whereas the current implementation is more future-proof. So, if you think we might add more to the context in the future I would land this PR as is, but if you think that is unlikely, then I could change the implementation to the 'by value' version.

nrc:
Test failures are because 1.32 does not have great support for NLL, I believe 1.33 should be ok.

danburkert:
> Design question: it occurs to me that this could be made simpler by passing the context by value and have `enter_recursion` create a new context with a decremented counter, rather than modifying the current context. Then we don't need to worry about incrementing the counter when we return (so don't need the RAII object, etc). However, if we add anything to the context in the future, we will have to use `Rc<RefCell>` and pay a cost of cloning, whereas the current implementation is more future-proof. So, if you think we might add more to the context in the future I would land this PR as is, but if you think that is unlikely, then I could change the implementation to the 'by value' version.

That sounds like a good idea.  If we need to add additional shared state to the context in the future we can do it through an inner referenced context objected.  As I understand it this would also mean removing the need for `unsafe`, right?  There's no unsafe in the decoding path currently, and if possible I'd like to maintain that.

src/message.rs
```diff
@@ -93,34 +99,38 @@ pub trait Message: Debug + Send + Sync {
...
-        message.merge_length_delimited(buf)?;
+        message.merge_length_delimited(buf, &mut DecodeContext::default())?;
         Ok(message)
     }
 
     /// Decodes an instance of the message from a buffer, and merges it into `self`.
     ///
     /// The entire buffer will be consumed.
-    fn merge<B>(&mut self, buf: B) -> Result<(), DecodeError>
+    fn merge<B>(&mut self, buf: B, ctx: &mut DecodeContext) -> Result<(), DecodeError>
```
danburkert: This is part of the public API, so it shouldn't have a context param, it should be handled the same as you did `decode_length_delimited` above.

src/message.rs
```diff
@@ -93,34 +99,38 @@ pub trait Message: Debug + Send + Sync {
...
         Ok(())
     }
 
     /// Decodes a length-delimited instance of the message from buffer, and
     /// merges it into `self`.
-    fn merge_length_delimited<B>(&mut self, buf: B) -> Result<(), DecodeError>
+    fn merge_length_delimited<B>(
+        &mut self,
+        buf: B,
+        ctx: &mut DecodeContext,
```
danburkert: likewise

src/message.rs
danburkert: In other words, `merge_field` should be the only method with a context param.

tests/src/lib.rs
```diff
@@ -341,6 +341,27 @@ mod tests {
         };
     }
 
+    #[test]
+    fn test_deep_nesting() {
+        fn build_and_roundtrip(depth: usize) -> Result<(), prost::DecodeError> {
```
danburkert: great test, but could you also add a case which covers the other nesting types, map, oneof, message and group?  This could be separate cases for each, or one case which nests those types within each other.

src/encoding.rs
```diff
@@ -77,6 +77,113 @@ where
...
+/// Additional information passed to every decode/merge function.
+#[derive(Clone, Debug)]
+pub struct DecodeContext {
+    #[cfg(feature = "recursion-limit")]
+    /// How many times we can recurse in the current decode stack before we hit
+    /// the recursion limit.
+    ///
+    /// The recursion limit is defined by `RECURSION_LIMIT` and cannot be
+    /// customized. The recursion limit can be ignored by building the Prost
+    /// crate without the `recursion-limit` feature (which is set by default).
```
danburkert: As I understand it, one side effect of modeling this as a on-by-default feature is that if there are two crates in a build graph that use `prost` and one disables it and another does not, it will be not be disabled.  Is that correct?

Alternatively, if it were an opt-in feature to enable turning off the recursion limit, than anyone in the build graph disabling it would disable it for everyone.

I'm not sure which is preferred, but it's an interesting distinction.  Was there a reason to prefer this way as opposed to the other?

src/encoding.rs
```diff
@@ -77,6 +77,113 @@ where
...
+
+impl DecodeContext {
+    #[cfg(feature = "recursion-limit")]
+    /// Call this function before recursively decoding.
+    ///
+    /// This function returns a guard object which will automatically restore
+    /// the recursion counter when it is destroyed by going out of scope.
+    ///
+    /// See the safety note on `RecursionGuard` for important information.
+    pub(crate) fn enter_recursion(&mut self) -> RecursionGuard {
```
danburkert: Yeah I think this can be significantly simplified through passing an owned struct which includes the counter, e.g. `fn recurse(&self) -> Result<Self, DecodeError>` to pass the context, and automatically increment or decrement the counter and check against 0 or max (I don't love the name `recurse`, though, so maybe there's a better option for that).


To enforce that `recurse` is used, the type should opt-out of `Clone` and `Copy`.

danburkert: > To enforce that recurse is used, the type should opt-out of Clone and Copy.

hmm actually that would be problematic for decoding `repeated` values.  Maybe a suitably named method which is effectively a clone but wouldn't be used accidentally would solve that.

tests/src/lib.rs
danburkert: and `repeated`.

danburkert:
The test failure is happening when using prost in edition-2015 mode.  I want to keep that an option for the time being, so you may need to adjust the problematic code to not rely on NLL.  You can test locally with `cargo test -p tests-2015`.  I have a feeling this may fall out naturally from removing the stack guards anyway.

src/message.rs
nrc: That doesn't work unfortunately because merge is called recursively so if we don't include the context, then we lose track of the recursion count

src/message.rs
nrc: We can change this one, although I thought it might be cleared for all `merge` methods to take a context

src/encoding.rs
nrc: Interesting, I hadn't thought about that, and instead was thinking only to avoid 'negative variables', but in this case I think it is better to use one.

src/encoding.rs
nrc: I think using Clone but opting out of Copy is fine. It doesn't force you to opt-in to recursing, but I don't think you can in general because the context can always be moved. Where there is a loop, it is nice to `clone` rather than implicitly copy I think - there is at least a hint that in some cases there is something else to do.

nrc:
> You can test locally with cargo test -p tests-2015.

This doesn't work - I think you do need an old compiler to test it (because NLL is backwards compatible, I think 2015 mode does not turn it off)

> I have a feeling this may fall out naturally from removing the stack guards anyway.

Sadly not because I use it in the test code to create the deeply nested proto.

nrc:
@danburkert  A few questions in replies, but otherwise all points addressed. This should work on 1.33, but I'll keep an eye on the CI.

nrc:
CI errors seem to be due to a missing lib or something breaking the conformance tests. Any idea what is going on here?

nrc:
@danburkert  ping: PR is updated, but I don't know how to fix CI. Comments above ^

src/encoding.rs
```diff
@@ -1295,7 +1407,12 @@ mod test {
                 ));
             }
 
-            if let Err(error) = merge(wire_type, &mut roundtrip_value, &mut buf) {
+            if let Err(error) = merge(
```
danburkert: Likewise, this is a consumer of the public API

danburkert:
Also rustc 1.32 compat is a hard requirement at this point.

src/encoding.rs
```diff
@@ -178,6 +178,80 @@ where
...
+    ///
+    /// The recursion limit is defined by `RECURSION_LIMIT` and cannot be
+    /// customized. The recursion limit can be ignored by building the Prost
+    /// crate with the `no-recursion-limit` feature (which is set by default).
+    recurse_count: u32,
+}
+
+impl Default for DecodeContext {
+    #[cfg(not(feature = "no-recursion-limit"))]
+    #[inline(always)]
```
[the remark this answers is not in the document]
nrc: `inline` is a fairly weak hint to the compiler, `inline(always)` is stronger and I wanted to ensure that adding the context does not add any overhead if we disable the recursion limit.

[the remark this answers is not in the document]
nrc: I don't think it actually matters - the compiler is likely to inline such a trivial function

nrc:
>Also rustc 1.32 compat is a hard requirement at this point.

I think this PR is 1.32 compatible now that I changed implementation.

src/encoding.rs
```diff
@@ -275,9 +349,14 @@ pub fn check_wire_type(expected: WireType, actual: WireType) -> Result<(), Decod
 
 /// Helper function which abstracts reading a length delimiter prefix followed
 /// by decoding values until the length of bytes is exhausted.
-pub fn merge_loop<T, M, B>(value: &mut T, buf: &mut B, mut merge: M) -> Result<(), DecodeError>
+pub fn merge_loop<T, M, B>(
```
nrc: @danburkert  Do you consider this and other similar functions to be part of the API? Technically they are visible to users, but they are doc(hidden) and are not very user friendly.

src/message.rs
```diff
@@ -29,6 +29,7 @@ pub trait Message: Debug + Send + Sync {
         tag: u32,
         wire_type: WireType,
         buf: &mut B,
+        ctx: DecodeContext,
```
nrc: And `merge_field`? It is part of the `Message` trait, but it is also `doc(hidden)` and the comment suggests it should not be used by users.

nrc:
@danburkert rebased and addressed your comments above. I have not removed some uses of the `DecodeContext` arg pending answers to my questions above.

src/encoding.rs
danburkert: No, anything `doc(hidden)` is not part of the public API, including this entire module, and the doc(hidden) methods on `Message`.

src/message.rs
danburkert: Right, this is not part of the public API, so this addition is fine.

danburkert: In reality this may still end up needing to be released as a minor version bump so as to avoid cargo resolution issues, but I wouldn't feel the need to release note the change.

nrc:
Given the answers above, I think this PR is ready to go. @danburkert could you re-review please?

src/encoding.rs
```diff
@@ -178,6 +178,80 @@ where
...
+/// The context should be passed by value and can be freely cloned. When passing
+/// to a function which is decoding a nested object, then use `enter_recursion`.
+#[derive(Clone, Debug)]
+pub struct DecodeContext {
+    /// How many times we can recurse in the current decode stack before we hit
+    /// the recursion limit.
+    ///
+    /// The recursion limit is defined by `RECURSION_LIMIT` and cannot be
+    /// customized. The recursion limit can be ignored by building the Prost
+    /// crate with the `no-recursion-limit` feature (which is set by default).
```
danburkert:
```suggestion
    /// crate with the `no-recursion-limit` feature.
```

src/encoding.rs
```diff
@@ -178,6 +178,80 @@ where
...
+    /// Checks whether the recursion limit has been reached in the stack of
+    /// decodes described by the `DecodeContext` at `self.ctx`.
+    ///
+    /// Returns `Ok<()>` if it is ok to continue recursing.
+    /// Returns `Err<DecodeError>` if the recursion limit has been reached.
+    #[cfg(not(feature = "no-recursion-limit"))]
+    #[inline(always)]
+    pub(crate) fn limit_reached(&self) -> Result<(), DecodeError> {
+        if self.recurse_count == 0 {
+            Err(DecodeError::new("Recursion limit reached"))
```
danburkert:
```suggestion
            Err(DecodeError::new("recursion limit reached"))
```

(this is just a matter of consistency with other errors prost generates)

src/encoding.rs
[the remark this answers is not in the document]
danburkert: I ask because everything piece of advice I've seen on the subject recommends to use `#[inline]`, and not `#[inline(always)]`.  Anecdotally I've found that `#[inline(always)]` doesn't usually help.

src/message.rs
```diff
@@ -101,14 +102,27 @@ pub trait Message: Debug + Send + Sync {
...
+    {
+        self.merge_with_context(buf, DecodeContext::default())
+    }
+
+    /// Decodes an instance of the message from a buffer, and merges it into `self`.
+    /// This version takes a `DecodeContext` which contains some state for the
+    /// current decoding. Most users will want to use `merge` without the context.
+    ///
+    /// The entire buffer will be consumed.
+    fn merge_with_context<B>(&mut self, buf: B, ctx: DecodeContext) -> Result<(), DecodeError>
```
danburkert: AFAICT the only caller of this is the merge function right above -- could this be removed?

danburkert: If not, it should be marked doc hidden, since DecodeContext isn't part of the public API.

nrc: Hmm, it turns out it is, I thought I called it from the code generated by prost-derive, but I don't

nrc:
@danburkert rebased and all comments addressed

BusyJay:
Any plan to make it possible to adjust the limit? TiKV uses 1000 as the default value for configuration, and we found it was occasionally changed in practice.

PR #187, opened 2019-05-25
Micro-optimise decoding

nrc:
A bunch of micro-optimisations for the decoding path. Despite the introduction of unsafe code, these are all safe.

r? @danburkert

### DIFF
--- a/src/encoding.rs
+++ b/src/encoding.rs
@@ -422,7 +422,7 @@ macro_rules! encode_repeated {
     };
 }
 
-/// Helper macro which emits a `merge_repeated_numeric` function for the numeric type.
+/// Helper macro which emits a `merge_repeated` function for the numeric type.
 macro_rules! merge_repeated_numeric {
     ($ty:ty,
      $wire_type:expr,
@@ -846,6 +846,16 @@ pub mod bytes {
     where
         B: Buf,
     {
+        // Append `s` to `v` using a `memcpy`.
+        // Safety: `v` must have enough capacity to accommodate `s`.
+        unsafe fn copy_bytes_to_vec(v: &mut Vec<u8>, s: &[u8]) {
+            let v_len = v.len();
+            let s_len = s.len();
+            let dst = v.as_mut_slice().get_unchecked_mut(v_len) as *mut u8;
+            ::std::ptr::copy_nonoverlapping::<u8>(&s[0] as *const u8, dst, s_len);
+            v.set_len(v_len + s_len);
+        }
+
         check_wire_type(WireType::LengthDelimited, wire_type)?;
         let len = decode_varint(buf)?;
         if len > buf.remaining() as u64 {
@@ -853,13 +863,21 @@ pub mod bytes {
         }
 
         let mut remaining = len as usize;
+        value.reserve(remaining);
+
         while remaining > 0 {
             let len = {
                 let bytes = buf.bytes();
                 debug_assert!(!bytes.is_empty(), "Buf::bytes returned empty slice");
                 let len = min(remaining, bytes.len());
                 let bytes = &bytes[..len];
-                value.extend_from_slice(bytes);
+                unsafe {
+                    debug_assert!(
+                        value.capacity() >= bytes.len(),
+                        "Should be ensured by the `reserve` statement above",
+                    );
+                    copy_bytes_to_vec(value, bytes);
+                }
                 len
             };
             remaining -= len;
--- a/src/encoding.rs
+++ b/src/encoding.rs
@@ -64,12 +64,12 @@ where
         return Err(DecodeError::new("invalid varint"));
     }
 
-    let byte = bytes[0];
+    let byte = unsafe { *bytes.get_unchecked(0) };
     if byte < 0x80 {
         buf.advance(1);
         Ok(u64::from(byte))
     } else if len > 10 || bytes[len - 1] < 0x80 {
-        let (value, advance) = decode_varint_slice(bytes)?;
+        let (value, advance) = unsafe { decode_varint_slice(bytes) }?;
         buf.advance(advance);
         Ok(value)
     } else {
@@ -82,32 +82,37 @@ where
 ///
 /// Based loosely on [`ReadVarint64FromArray`][1].
 ///
+/// ## Safety
+///
+/// The caller must ensure that `bytes` is non-empty and either `bytes.len() >= 10` or the last
+/// element in bytes is < `0x80`.
+///
 /// [1]: https://github.com/google/protobuf/blob/3.3.x/src/google/protobuf/io/coded_stream.cc#L365-L406
 #[inline]
-fn decode_varint_slice(bytes: &[u8]) -> Result<(u64, usize), DecodeError> {
+unsafe fn decode_varint_slice(bytes: &[u8]) -> Result<(u64, usize), DecodeError> {
     // Fully unrolled varint decoding loop. Splitting into 32-bit pieces gives better performance.
 
     let mut b: u8;
     let mut part0: u32;
-    b = bytes[0];
+    b = *bytes.get_unchecked(0);
     part0 = u32::from(b);
     if b < 0x80 {
         return Ok((u64::from(part0), 1));
     };
     part0 -= 0x80;
-    b = bytes[1];
+    b = *bytes.get_unchecked(1);
     part0 += u32::from(b) << 7;
     if b < 0x80 {
         return Ok((u64::from(part0), 2));
     };
     part0 -= 0x80 << 7;
-    b = bytes[2];
+    b = *bytes.get_unchecked(2);
     part0 += u32::from(b) << 14;
     if b < 0x80 {
         return Ok((u64::from(part0), 3));
     };
     part0 -= 0x80 << 14;
-    b = bytes[3];
+    b = *bytes.get_unchecked(3);
     part0 += u32::from(b) << 21;
     if b < 0x80 {
         return Ok((u64::from(part0), 4));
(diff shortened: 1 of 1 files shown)


nrc:
My primary benchmark is a micro-benchmark based on a very common workload for TiKV (https://github.com/nrc/prost-bench/blob/07cd31b7022cd38cd3a129b5c425a5fc088856aa/src/bin/proto.rs). On that benchmark, I saw an improvement of 7.6% (all benchmarking done on a CPU-optimised Digital Ocean droplet with the highest spec available, running Ubuntu 19.04). I saw similar improvements on my Mac laptop (although since it is a working laptop there was a lot more variation between runs).

Running `cargo bench`, the encoding Criterion benchmarks were `Change within noise threshold.`. The decode and merge Criterion benchmarks were improved by 0-3% (there was a lot more noise than with my benchmark, I'm not sure why).

For the varint benchmarks, the encoding benchmarks were unchanged. The decoding benchmarks improved by 6%, 33%, 31%, and 21% (small, mixed, medium, large).

nrc:
Tests failing on 1.32 because `resize_with` was stabilised in 1.33.

src/encoding.rs
```diff
@@ -762,20 +815,38 @@ pub mod bytes {
...
                 debug_assert!(!bytes.is_empty(), "Buf::bytes returned empty slice");
                 let len = min(remaining, bytes.len());
                 let bytes = &bytes[..len];
-                value.extend_from_slice(bytes);
+                unsafe {
+                    debug_assert!(
+                        value.capacity() >= bytes.len(),
+                        "Should be ensured by the `reserve` statement above",
+                    );
+                    copy_bytes_to_vec(value, bytes);
```
danburkert: Were you able to isolate this change and get improved numbers?  As I understand it the savings here is one removed bounds check per loop iteration on the `value` buffer.  Otherwise, `extend_from_slice` should boil down to the same call to `copy_nonoverlapping` internally.

Importantly, when the buf type is `Vec` or `Bytes` (e.g., almost always), this loop will only be iterated once, since `buf.bytes()` will return the full buffer.  As such I'd be surprised if this meaningfully moved the needle, and if it did I'd suspect it came down to inlining changes more than bounds checks.

src/encoding.rs
```diff
@@ -832,9 +903,9 @@ pub mod message {
         B: Buf,
     {
         check_wire_type(WireType::LengthDelimited, wire_type)?;
-        let mut msg = M::default();
-        merge(WireType::LengthDelimited, &mut msg, buf)?;
-        messages.push(msg);
+        let len = messages.len();
+        messages.resize_with(len + 1, M::default);
```
danburkert: nice, there are probably more opportunities in the codebase for improvements like this.  With prost's model of inlining message fields, resulting in large structs, these can be a big win.

src/encoding.rs
```diff
@@ -159,6 +165,53 @@ fn decode_varint_slice(bytes: &[u8]) -> Result<(u64, usize), DecodeError> {
...
+///
+/// ## Safety
+///
+/// The caller must ensure that `bytes` is non-empty and either `bytes.len() >= 10` or the last
+/// element in bytes is < `0x80`.
+///
+/// [1]: https://github.com/google/protobuf/blob/3.3.x/src/google/protobuf/io/coded_stream.cc#L365-L406
+#[inline]
+#[cfg(target_pointer_width = "64")]
+unsafe fn decode_varint_slice(bytes: &[u8]) -> Result<(u64, usize), DecodeError> {
```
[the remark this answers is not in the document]
danburkert: (The `step` macro is great, thanks for doing that)

src/encoding.rs
```diff
@@ -82,32 +82,38 @@ where
...
 #[inline]
-fn decode_varint_slice(bytes: &[u8]) -> Result<(u64, usize), DecodeError> {
+#[cfg(not(target_pointer_width = "64"))]
+unsafe fn decode_varint_slice(bytes: &[u8]) -> Result<(u64, usize), DecodeError> {
     // Fully unrolled varint decoding loop. Splitting into 32-bit pieces gives better performance.
 
     let mut b: u8;
     let mut part0: u32;
-    b = bytes[0];
+    b = *bytes.get_unchecked(0);
```
danburkert: I'm fine with making these unchecked, but I do want to keep the bar high for introducing unsafety on the decoding side.  I'm pretty sure I benchmarked making these unchecked in the past and wasn't able to move the needle on the in-tree varint decoding benchmarks.  Did you measure improvements with this change in isolation?

src/encoding.rs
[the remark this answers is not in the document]
nrc: > Other than the unchecked accesses, is the only difference here that the results are aggregated into a u64 instead of a pair of u32s?

Yes. I'll remove the 32 bit version.

src/encoding.rs
nrc: > Were you able to isolate this change and get improved numbers?

I didn't get accurate numbers for each change, but I did see an improvement for each one (and tried lots of things which gave no improvement and dumped them). I'll try and get something more precise.

> Otherwise, extend_from_slice should boil down to the same call to copy_nonoverlapping internally.

Because `extend_from_slice` is generic and there is a lot of indirection in the implementation I'm not sure whether we hit the fast path or not.

> Importantly, when the buf type is Vec or Bytes (e.g., almost always)

This is not the case for us - we get a buffer from grpc and it can come a piece at a time.

src/encoding.rs
nrc: I did see a small but significant change, I'll try and get some good numbers here too.

nrc:
Fun with benchmarking! So, I got wildly different results between Mac OS and Linux. Unfortunately on Mac OS the results were not very consistent or reliable, I am 90% that every commit has a small but positive effect, with a total win of around 5-10% (yep, it's that noisy).

On Linux I was able to get pretty precise numbers for each commit. Most surprising was that 6b1bf72989a884f0f6a0e1da9f04e76a775f2fe7 caused 9% performance regression! Given that the improvement on Mac OS was only a few percent, I propose not landing that commit.

You asked about 3dceed2e45180fb4fe2507aa952ea56a52903a8f (using `get_unchecked` in `decode_varint`), that gave an almost 5% improvement, so I think that is worth keeping. 

You also asked about cb871d8a04799641819314c758bab7dfdcd85a0b (implementing our own memcpy), that gave a 0.3% improvement, so I guess we could take or leave that.

After removing 6b1bf72989a884f0f6a0e1da9f04e76a775f2fe7, the total performance improvement was almost 17%

nrc:
Oh all these benchmarks were using a repeated encode/decode. I wasn't able to run `cargo bench` because of some errors about the conformance tests (even on master).

nrc:
@danburkert do you have any more questions about this PR? Are the numbers above good enough to land this?

danburkert:
I'm running these commits on the in-tree benchmark suite from the `benchmarks` branch and recording results/notes here (output is either from criterion, or from critcmp):

1. f90eb7cca0 Strengthen some inlines and remove a cast

Improves the performance of merge/decode benchmarks by between 0 and 11%.  This appears to be due to the addition of `inline(always)`, and not due to removing the cast, but the code is nicer without the cast regardless.  Example:

    dataset/google_message2/decode
                            time:   [857.33 us 858.48 us 859.68 us]
                            thrpt:  [93.817 MiB/s 93.948 MiB/s 94.074 MiB/s]
                    change:
                            time:   [-10.638% -10.551% -10.458%] (p = 0.00 < 0.05)
                            thrpt:  [+11.680% +11.795% +11.904%]
                            Performance has improved.

2. cb871d8  Optimise bytes::merge by pre-allocating and using an efficient copy

Improves the performance of merge/decode benchmarks by between 1 and 4%. This appears to be due to the optimized byte copy function, and not due to the addition of the reserve() call.  I'm wasn't super happy with this addition of unsafe code, so I went spelunking in `bytes` and found the `BytesMut::put()` method, which does essentially the same thing except with a panic instead of assuming the capacity is present.  This results in a net improvement in code cleanliness in `bytes::merge` since it no longer has to deal with the loop:

```rust
pub fn merge<B>(
    wire_type: WireType,
    value: &mut Vec<u8>,
    buf: &mut B,
    _ctx: DecodeContext,
) -> Result<(), DecodeError>
where
    B: Buf,
{
    check_wire_type(WireType::LengthDelimited, wire_type)?;
    let len = decode_varint(buf)?;
    if len > buf.remaining() as u64 {
        return Err(DecodeError::new("buffer underflow"));
    }
    let len = len as usize;
    value.reserve(len);

    value.put(buf.take(len));
    Ok(())
}
```

And appears to be faster than cb871d8: 

    group                                    buf-put                                cb871d8                                 baseline (f90eb7cca0)
    -----                                    -------                                ----------                              ---------------------
    dataset/google_message3_2/merge          1.00    217.7±0.06ms   462.4 MB/sec    1.03    223.3±0.17ms   450.8 MB/sec     1.09    236.9±2.30ms   425.1 MB/sec

@nrc can you try that impl on your benchmarks?

3. 45d51d140 (Use get_unchecked rather than indexing where it is safe to do so)

TODO

4. 019aae (Specialise decode_varint_slice for 64bit platforms)

TODO

nrc:
> @nrc can you try that impl on your benchmarks?

I made that modification, leaving all the other changes in place. My changes brought an improvement of 11.5% over master, with your change I saw 13.7% (results were fairly consistent: CV of 0.8%). Tested on Mac OS using jemalloc (fast MacBook Pro).

danburkert:
> My changes brought an improvement of 11.5% over master, with your change I saw 13.7% (results were fairly consistent: CV of 0.8%).

Sounds great, sounds like a win win.

@nrc I've finished reviewing the patch set, sorry again for the delay.  Everything looks good except for the 'Specialise decode_varint_slice for 64bit platforms' commit (and the final commit, which builds off of it).  I've tested on a [Skylake Xeon](https://ark.intel.com/content/www/us/en/ark/products/120498/intel-xeon-platinum-8180m-processor-38-5m-cache-2-50-ghz.html), a [lowerpower Broadwell Xeon](https://ark.intel.com/content/www/us/en/ark/products/87039/intel-xeon-processor-d-1540-12m-cache-2-00-ghz.html) and a [mobile Coffee Lake i7](https://ark.intel.com/content/www/us/en/ark/products/137979/intel-core-i7-8559u-processor-8m-cache-up-to-4-50-ghz.html), and the results are consistently no-change on small (1 byte) varint decoding, and around ~15% decrease in large (10 byte) varint decoding.  Are you seeing something different in your tests?  May be worth digging into if so.

nrc:
> Sounds great, sounds like a win win.

Indeed!

> 'Specialise decode_varint_slice for 64bit platforms

Hmm, I did find this commit improved things for my benchmark. I ran on a Xeon on Digital Ocean and on my MacBook Pro (i9), and got improvements both times. I reran the benchmark on my mac and got some consistent results - with the whole PR I got an improvement of 13.5% over master and without the last two commits ("Specialise decode_varint_slice for 64bit platforms" and "Remove the 32 bit version of `decode_varint_slice`") I got a 9.9% improvement. (I was testing a whole encode/decode cycle, not just decoding varints). I think the varints would mostly be small.

> May be worth digging into if so.

Yes, I would like to land the other commits sooner rather than later, if that is OK, so I will remove those from this PR and send a new PR with just those commits and I can investigate further.

nrc:
I think this is what we both agree we should land. Once landed, it would be awesome if you could publish a new version with these changes and the recursion check.

danburkert:
> Yes, I would like to land the other commits sooner rather than later, if that is OK, so I will remove those from this PR and send a new PR with just those commits and I can investigate further.

Sounds great!

PR #196, opened 2019-06-17
Reintroduce unsafe in encoding::string::merge

danburkert:
#194  introduced a ~3% regression on some decoding
benchmarks. My best guess is that swapping the value temporarily with a
new string is not being optimized well. This commit reverts to using
String::as_mut_vec, but carefully clears the string on decode failure.
It also fixes up the associated regression test, which wasn't properly
covering the malformed UTF-8 case.

    // Before #194
    dataset/google_message3_2/decode
                            time:   [219.45 ms 219.48 ms 219.51 ms]
                            thrpt:  [458.64 MiB/s 458.72 MiB/s 458.78 MiB/s]

    // #194
    dataset/google_message3_2/decode
                            time:   [226.08 ms 226.11 ms 226.18 ms]
                            thrpt:  [445.12 MiB/s 445.25 MiB/s 445.32 MiB/s]

    // This commit
    dataset/google_message3_2/decode
                            time:   [219.05 ms 219.76 ms 221.78 ms]
                            thrpt:  [453.94 MiB/s 458.12 MiB/s 459.61 MiB/s]

cc @mzabaluev @nrc

### DIFF
--- a/src/encoding.rs
+++ b/src/encoding.rs
@@ -5,6 +5,7 @@
 use std::cmp::min;
 use std::convert::TryFrom;
 use std::mem;
+use std::str;
 use std::u32;
 use std::usize;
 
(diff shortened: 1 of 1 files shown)


danburkert:
The benchmark numbers above were run on the https://github.com/danburkert/prost/tree/benchmarks branch.  I'm putting some TLC into the benchmark suite so I can properly review #187, which lead to finding this regression.

src/encoding.rs
```diff
@@ -815,11 +815,23 @@ pub mod string {
...
-        bytes::merge(wire_type, &mut value_bytes, buf, ctx)?;
-        *value = String::from_utf8(value_bytes)
-            .map_err(|_| DecodeError::new("invalid string value: data is not UTF-8 encoded"))?;
-        Ok(())
+        unsafe {
+            // String::as_mut_vec is unsafe because it doesn't check that the bytes
+            // inserted into it the resulting vec are valid UTF-8. We check
+            // explicitly in order to ensure this is safe.
+            let value = value.as_mut_vec();
+            super::bytes::merge(wire_type, value, buf, ctx)?;
```
mzabaluev: This still leaves the `String` value in invalid state if any method on `buf` panics.
A scope guard, like originally proposed in 55f644e7401c3cf0dfef2b078a73c5362641f6a2 for #194 but working on the `&mut Vec<u8>` to avoid technical unsoundness, would be more bulletproof.

danburkert:
@mzabaluev 

> This still leaves the String value in a potentially invalid state if any method on buf panics.
> A scope guard, like originally proposed in 55f644e for #194 but working on the &mut Vec<u8> to > avoid technical unsoundness, would be more bulletproof.

Great point.  I've pushed a new implementation with a drop guard.  It's simpler than the original, since it turned out the implemention should have been clearing the string/bytes values on merge all along.  This was revealed to be an issue in the conformance tests during the move from protobuf 3.7 to 3.11.  I'm happy to report this drop guard impl has the same performance as without the drop guard.

PR #212, opened 2019-08-10
Remove byteorder-dependency

lukaslueg:
This removes the byteorder-dependency, which does not seem to be referenced anywhere


danburkert:
Thanks!

Issue #211, opened 2019-08-10
panicked at 'expected a package' in prost_build::Config::compile_protos


dckc:
why require a package? Am I missing something?

The [hello world example in the docs](https://developers.google.com/protocol-buffers/docs/proto3#simple) doesn't have one:

```proto
syntax = "proto3";

message SearchRequest {
  string query = 1;
  int32 page_number = 2;
  int32 result_per_page = 3;
}
```

https://github.com/danburkert/prost/blob/1944c27c3029d01ff216e7b126d846b6cf8c7d77/prost-build/src/message_graph.rs#L25

danburkert:
Yes, `prost` requires a package declaration.

dckc:
Oh. I see that in the README now. In case it would help, I submitted a PR to document it in the API.

fenollp:
Turns out compiling a package-less https://github.com/hyperium/tonic app fails in a very unhelpful error (from an `unwrap`ped `tonic_build::compile_protos` call):
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: NotPresent', src/libcore/result.rs:1165:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

How can we make this output a more helpful error message?

cc @hyperium @LucioFranco

danburkert:
I'm going to close this out, there have been two commits landed to master which improve the situation:

1. 8df3a14e8c58d2b5dbf7ec7eb2ee8b22b412afe1 by @joelgallant improves the error message associated with a missing package specifier.
2. 8c0c18031fadbb51be887514ff8b0d208517d613 by @danburkert changes the panic to an error, and adds the error condition to the `compile_protos` function doc.

Please feel free to re-open or file a new issue if there's something more that can be improved.

PR #213, opened 2019-08-10
document panic on lack of package declaration

dckc:
This is documented in the README, so I suppose it should go here too.

See also #211


danburkert:
Hi @dckc , thanks for the PR!  I've implemented something slightly different in #8c0c18031fadbb51be887514ff8b0d208517d613 - instead of documenting the panic, I've changed it so that it returns an error instead, and documented the API as such.

PR #214, opened 2019-08-12
In proto3, everything is optional

matklad:
Just a drive-by change. I haven't used protobuf lately, but I *think* this should be an Option

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,7 +122,7 @@ the Rust field:
 | --- | --- | --- |
 | `proto2` | `optional` | `Option<T>` |
 | `proto2` | `required` | `T` |
-| `proto3` | default | `T` |
+| `proto3` | default | `Option<T>` |
 | `proto2`/`proto3` | repeated | `Vec<T>` |
 
 #### Map Fields


snnsnn:
Exactly, with current implementation it seems all fields are required:

```
missing fields `abused`, `bot`, `room` and 8 other fields in initializer of `log_service::CreateLogRequest`
missing `abused`, `bot`, `room` and 8 other fieldsrustc(E0063)
```

> Required fields are no longer available.
https://cloud.google.com/apis/design/proto3

Reasoning is: 

> Required Is Forever You should be very careful about marking fields as required. If at some point you wish to stop writing or sending a required field, it will be problematic to change the field to an optional field – old readers will consider messages without this field to be incomplete and may reject or drop them unintentionally. You should consider writing application-specific custom validation routines for your buffers instead. Some engineers at Google have come to the conclusion that using required does more harm than good; they prefer to use only optional and repeated. However, this view is not universal.
https://developers.google.com/protocol-buffers/docs/proto#required_warning

danburkert:
The documentation changed in this PR is about the Rust code generated by `prost`, and the change is incorrect, so I'm going to close the issue.  I'm not exactly sure what or where that error is coming from @snnsnn, open an issue if you have problems.

matklad:
That means that the statement that follows is incorrect:

>except that proto3 message fields without a modifier (the default) will be wrapped in an Option

danburkert:
The full sentence is

>  The table of field modifiers above applies to message fields, except that `proto3` message fields without a modifier (the default) will be wrapped in an `Option`.

and is correct.  That sentence describes an exception to the table.

matklad:
Ah, I see now, sorry for the noise!

danburkert:
No problem, thanks for taking the initiative to send a PR!

PR #219, opened 2019-08-25
Update proc-macro2, syn and quote to 1.0

### LINKED ISSUE #218: Update proc-macro2, syn & quote to 1.0
The three proc macro libraries were recently released as 1.0, so it would be great to have a updated version of prost with those for less duplicate dependencies.

hobofan:
CLOSES #218

The biggest changes in syn 1.0 relevant to prost are the switch from `ident` to `path` in `Meta`, and the removal of `.value()` on literals.

The biggest logical change I had to make is how negative default values are parsed. Apart from that the printing of an attribute via `attr.path().into_token_stream()` (since nothing straightforward printable like `.name()` exists that I could find), is the only other logical change.

Everything else should be the 1:1 equivalent to the old logic after renaming.

### DIFF
--- a/prost-derive/Cargo.toml
+++ b/prost-derive/Cargo.toml
@@ -16,6 +16,6 @@ proc_macro = true
 [dependencies]
 failure = { version = "0.1", default-features = false, features = ["std"] }
 itertools = "0.8"
-proc-macro2 = "0.4.4"
-quote = "0.6.3"
-syn = { version = "0.15", features = [ "extra-traits" ] }
+proc-macro2 = "1.0"
+quote = "1.0"
+syn = { version = "1.0", features = [ "extra-traits" ] }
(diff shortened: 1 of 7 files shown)


LucioFranco:
@hobofan looks like CI failure is due to a dependency using a newer rust feature than what CI is using right now.

@danburkert any way we can get this merged? I'd like to avoid multiple syn versions for `tonic`. Happy to help anyway I can.

hobofan:
Looks like the problem of the increased MSRV is `flate2`, which is a direct dependency of `prost`, so I don't think there is anything I can do about that 🤷‍♂

danburkert:
These changes (or something very close to them) have been merged as part of #5e2bc154fab6e.  Thanks for the PR, @hobofan !

ErichDonGubler:
My only objection to how this PR and mine was handled is that @hobofan is not listed in the 0.6 release as having contributed. He most certainly did, and I'd hate to see him continue unmentioned! :slightly_smiling_face:

danburkert:
@ErichDonGubler great point, that's a definite oversight.  I've updated the release notes to include @hobofan.  Thanks for calling this out.

PR #220, opened 2019-08-31
Fix compilation error for types named "Option" with a repeated field

Slowki:
This fixes a compilation error I ran into with [this protobuf](https://github.com/bazelbuild/bazel/blob/20b056ae8506e489933cf236b9401cddad90afac/src/main/protobuf/command_line.proto#L74-L102).


danburkert:
Thanks for the PR!  This has been merged to master.

PR #224, opened 2019-09-27
fix: ident Self field keyword as Self_

hyyking:
Issue generating: https://github.com/Blizzard/s2client-proto

`Self` is used in a field in `raw.pb` and is overridden by `r#Self` which is incorrect rust

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ os:
   - osx
 
 rust:
-  - 1.32.0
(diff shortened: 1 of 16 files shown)


danburkert:
Hi @hyyking, thanks for the PR.  This issue was fixed in #180, which will be included in the next release.  This PR has no effect since a 'Self' value is not possible after calling `to_snake_case`

PR #225, opened 2019-09-29
Add "std" feature everywhere to support no_std envs

dflemstr:
I made these changes to be able to run prost in a no_std environment where alloc is still available.

This is similar to #215 except it uses slightly different naming conventions in different places (i.e. the feature is called `std` instead of `use_std` consistently and so on).  Truth be told, I created this version without being aware of the other one. Maybe the two approaches could be combined!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ os:
   - osx
 
 rust:
-  - 1.32.0
+  - 1.36.0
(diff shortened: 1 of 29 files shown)


dflemstr:
Synced with the authors offline, and I will close this PR in favor of #215

PR #232, opened 2019-10-14
Don't write prost-build output if its content is unchanged

### LINKED ISSUE #139: prost-build triggers rebuild on arbitrary file change
Let's say I have some rust project and have done `cargo build`. Now I touch an arbitrary file around there, eg `touch x`. If I run `cargo build` now, it is very fast and mostly figures out it doesn't need to do anything.

However, this changes if I introduce `prost-build` into the project. Now the `cargo build` will rebuild the whole crate (and all the tests and other things if it is not `cargo build` but eg. `cargo test --all`).

My guess is, this has something to do with the fact that the file change provokes `build.rs` to be re-run. That one, in turn, overwrites the `rs` files generated from protobufs, which triggers the rebuild (even though the produced files are the same).

If my guess is correct, would it make sense to generate them to some temporary file first (eg. `generated_file.rs.tmp`), compare with the previous version (if present) and then move it into place only if it differs?

### LINKED ISSUE #231: Only write files when the contents have changed
As a workaround to code completion not working (#229), and because I've decided it is a good idea anyway, I changed my `build.rs` to output the generated code to the source directory.

Unfortunately it completely breaks rust-analyzer which watches the code for changes, then runs `build.rs` and gets stuck in a loop.

Probably should be fixed there, but I think a simple and worthwhile fix would be for prost-build to only write files if their contents have changed. I've used this technique in a CMake based project and it works very well.

Might fix #139 too, who knows.

Timmmm:
This means things like cargo-watch work better. Fixes #231

### DIFF
--- a/prost-build/src/lib.rs
+++ b/prost-build/src/lib.rs
@@ -114,7 +114,7 @@ use std::collections::HashMap;
 use std::default;
 use std::env;
 use std::fs;
-use std::io::{Error, ErrorKind, Read, Result, Write};
+use std::io::{Error, ErrorKind, Result};
 use std::path::{Path, PathBuf};
 use std::process::Command;
 
@@ -536,18 +536,24 @@ impl Config {
             ));
         }
 
-        let mut buf = Vec::new();
-        fs::File::open(descriptor_set)?.read_to_end(&mut buf)?;
+        let buf = fs::read(descriptor_set)?;
         let descriptor_set = FileDescriptorSet::decode(&buf)?;
 
         let modules = self.generate(descriptor_set.file)?;
         for (module, content) in modules {
             let mut filename = module.join(".");
             filename.push_str(".rs");
-            trace!("writing: {:?}", filename);
-            let mut file = fs::File::create(target.join(filename))?;
-            file.write_all(content.as_bytes())?;
-            file.flush()?;
+
+            let output_path = target.join(&filename);
+
+            let previous_content = fs::read(&output_path);
+
+            if previous_content.is_ok() && previous_content.unwrap() == content.as_bytes() {
+                trace!("unchanged: {:?}", filename);
+            } else {
+                trace!("writing: {:?}", filename);
+                fs::write(output_path, content)?;
+            }
         }
 
         Ok(())


Timmmm:
This is also necessary if you are using [`cross`](https://github.com/rust-embedded/cross) because it makes the source directory read-only.

I think the CI failure is unrelated.

dbrgn:
Couldn't this solved with this in `build.rs`?

```rust
println!("cargo:rerun-if-changed=path/to/schema.proto");
```

Timmmm:
No. That just means that cargo will re-trigger a build if `schema.proto` changes, which it already does by default as long as `schema.proto` is inside your crate directory.

If anything you would want something like:

    println!("cargo:DONT-rerun-if-changed=path/to/output.rs");

But that is way more ugly and less robust than this change.

dbrgn:
> No. That just means that cargo will re-trigger a build if schema.proto changes, which it already does by default as long as schema.proto is inside your crate directory.

Ah, I assume that didn't work for me because my proto files are in a submodule adjacent to the `src` directory, not inside.

Timmmm:
> Ah, I assume that didn't work for me because my proto files are in a submodule adjacent to the `src` directory, not inside.

Yeah I'm not sure actually what the default behaviour is meant to be - the documentation is not clear:

> Normally build scripts are re-run if any file inside the crate root changes, but this can be used to scope changes to just a small set of files.

Is "the crate root" the directory containing `Cargo.toml`, or is it `src`, and does it check recursively?

In any case, this patch is a much better approach.

Timmmm:
However there may be a separate bug in that it should *also* be printing `println!("cargo:rerun-if-changed=path/to/schema.proto");` for all of the input files. That could be left up to the user I guess but if so it should probably be added to the documentation for this crate because it's a pretty obscure thing to know about!

Timmmm:
Is there anything left to do before landing this?

danburkert:
> However there may be a separate bug in that it should _also_ be printing `println!("cargo:rerun-if-changed=path/to/schema.proto");` for all of the input files. That could be left up to the user I guess but if so it should probably be added to the documentation for this crate because it's a pretty obscure thing to know about!

This is 'discussed' [in the source](https://github.com/danburkert/prost/blob/master/prost-build/src/lib.rs#L504-L508), but it's not enabled since it isn't the right thing to do for a build.rs which is also handling other responsibilities, e.g. compiling a native C dylib and doing `prost` codegen.  I think I'd be in support of an additional codegen flag which caused these lines to be emitted, and I'd almost go so far as to say it should be on by default.

danburkert:
Merged to master with minor changes in #8c647.  Thanks for the PR, @Timmmm!  If anyone wants to continue the discussion of `rerun-if-changed`, I invite you to open an issue to discuss.

shahn:
`rerun-if-changed`  is already discussed in #139

PR #234, opened 2019-10-18
improved generated docstrings

### LINKED ISSUE #233: improve deny(missing_docs) compatibility
Trying the example from https://docs.rs/prost-build/0.5.0/prost_build/ with #![deny(missing_docs)] on has some issues:

1)  The comment for the `Shirt` message doesn't get applied to the generated `shirt` module.

2) The `::prost::Message` proc macro seems to generate at least one method, and that method doesn't have a doc string.

Ideally I'd rather not make an `allow(missing_docs)` exception because I like the idea of forcing my users to document their messages.  Would you be open to a PR that fixes these things?

dunmatt:
This (hopefully) resolves #233 

This is my first time looking through the prost source, so it's entirely possible I didn't understand something, review with the assumption that I'm missing the obvious.

### DIFF
--- a/prost-build/src/code_generator.rs
+++ b/prost-build/src/code_generator.rs
@@ -683,7 +683,6 @@ impl<'a> CodeGenerator<'a> {
     }
 
     fn push_mod(&mut self, module: &str) {
-        self.push_indent();
         self.append_doc();
         self.push_indent();
         self.buf.push_str("pub mod ");
--- a/prost-build/src/code_generator.rs
+++ b/prost-build/src/code_generator.rs
@@ -683,7 +683,7 @@ impl<'a> CodeGenerator<'a> {
     }
 
     fn push_mod(&mut self, module: &str) {
-        self.append_doc();
+        // self.append_doc();
         self.push_indent();
         self.buf.push_str("pub mod ");
         self.buf.push_str(&to_snake(module));
--- a/prost-build/src/code_generator.rs
+++ b/prost-build/src/code_generator.rs
@@ -684,6 +684,8 @@ impl<'a> CodeGenerator<'a> {
 
     fn push_mod(&mut self, module: &str) {
         self.push_indent();
+        self.append_doc();
+        self.push_indent();
         self.buf.push_str("pub mod ");
         self.buf.push_str(&to_snake(module));
         self.buf.push_str(" {\n");
--- a/prost-derive/src/field/map.rs
+++ b/prost-derive/src/field/map.rs
@@ -222,10 +222,14 @@ impl Field {
                 quote!()
             };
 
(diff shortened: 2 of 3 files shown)


danburkert:
@dunmatt the test issue is that you need to check in the changes to prost-types; whenever code-gen changes this happens.  They are somewhat akin to snapshot tests, however they are actually load bearing code.

danburkert:
Which should be as simple as running the test suite, which modifies the files

dunmatt:
I get:

       Compiling criterion v0.2.11
    error: failed to run custom build command for `protobuf v0.0.0 (/home/mrd/code/prost/protobuf)`

    Caused by:
      process didn't exit successfully: `/home/mrd/code/prost/target/debug/build/protobuf-c5fd84c2727b695e/build-script-build` (exit code: 101)
    --- stderr
    + mkdir -p third_party/googletest/m4
    + autoreconf -f -i -Wall,no-obsolete
    configure.ac:104: error: possibly undefined macro: AC_PROG_LIBTOOL
          If this token and others are legitimate, please use m4_pattern_allow.
          See the Autoconf documentation.
    autoreconf: /usr/bin/autoconf failed with exit status: 1
    thread 'main' panicked at 'protobuf autogen.sh failed', protobuf/build.rs:150:9
    note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.


when I try to run the tests.  Any idea what I should do?

danburkert:
That's a failure to build `protobuf`, which is required by the tests suites.  For a bit more context this is the test that matters: https://github.com/danburkert/prost/blob/master/tests/src/bootstrap.rs.  I'm not really going to be able to help debug why protobuf isn't building correctly via GH issues, but it's probably something along the lines of a missing build tools or the wrong versions.

dunmatt:
I see.  So here's the thing, chasing down C++ build issues is obnoxious enough that I've almost completely abandoned the language.  If I send you a .patch file that does what this PR is supposed to do can you apply it and use the tests to generate the remainder of the PR?

danburkert:
Yeah I can take over just based on this PR, no need to send a patch.  Thanks for the PR!

dunmatt:
Thanks for your help with it!

danburkert:
@dunmatt I've opened up #250 to land just the getter/setter method doc changes.  The docs for generated modules as implemented in this PR are just copying the outer struct documentation, which I don't think is appropriate.  I think it'd be better to have a docstring for generated modules be along the lines of

> Nested message and enum types in `Foo`.

where `Foo` is the outer message type.

danburkert:
So the bulk of this landed a year ago, I'm going to close this out since I suspect there's nothing more to do here.  If I'm wrong please feel free to re-open or send a new PR.  Thanks!

dunmatt:
No, thank you

PR #240, opened 2019-11-19
Update to `syn`/`quote`/`proc-macro2` 1.0

### LINKED ISSUE #218: Update proc-macro2, syn & quote to 1.0

ErichDonGubler:
Where possible, I've tried to preserve the semantics of the existing code. It's possible that I misunderstood some intent -- feel free to tell me what I've borked! :)

Supersedes #219, resolves #218. I created this PR by myself, and then reviewed it against #219 to minimize differences and hopefully improve the implementation. Vivint uses this in its stack, so we're interested in keeping this PR up-to-date!

The only non-mechanical change I think made here I didn't expect was needing to modify the parsing of integer literals in the case of signed types and negative values. Since `syn` no longer provides the parsed values of integer and float literals, we need to provide the minus again ourselves when going to use the existing `FromStr` implementation for the related primitives.

### DIFF
--- a/prost-derive/Cargo.toml
+++ b/prost-derive/Cargo.toml
@@ -16,6 +16,6 @@ proc_macro = true
 [dependencies]
 failure = { version = "0.1", default-features = false, features = ["std"] }
 itertools = "0.8"
-proc-macro2 = "0.4.4"
-quote = "0.6.3"
-syn = { version = "0.15", features = [ "extra-traits" ] }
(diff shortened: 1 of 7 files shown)


danburkert:
Merged to master with the following tiny changes:

PR #241, opened 2019-11-21
Add ServiceGenerator::finalize_package

parasyte:
Similar in motivation to `ServiceGenerator::finalize`, this adds a new trait method finalizer for an entire package, not just individual proto files.

This resolves a challenge in `tonic-build` that causes multiple `client` and `server` modules to appear in the generated code when service definitions for a _single package_ are spread across _multiple proto files_. See https://github.com/hyperium/tonic/issues/38 for existing discussion.

With this feature, `tonic-build` can very easily support the use case with complex proto packages broken into smaller proto files. The following comment contains a simplified example of such a proto file layout: https://github.com/hyperium/tonic/issues/38#issuecomment-557213434


parasyte:
This is not currently working the way `tonic` expects. The unordered nature of `HashMap` iteration is causing some breaking changes. I'm looking into a way to fix this now.

parasyte:
This is fixed. I will add some tests in a bit. @danburkert if this looks good to you, I'd like to propose a new patch release with this feature included.

The unordered iteration wasn't _really_ the problem. `tonic-build` just needed a way to associate the `finalize_package` calls with an actual package name.

parasyte:
Added tests for the new feature.

fujiapple852:
Nice work @parasyte.  I'm also blocked on this whilst looking to migrate from `protoc-grpcio` to `tonic`.  

I've been using a specific commit from `master` for a while now as it has a fix for another (unrelated) blocking issue and so rather than a patch release specifically for _this_ issue it would be good to get a full release of `prost` with all a rollup of all the latest changes and fixes when feasible.

danburkert:
Merged.  Thanks @parasyte!

PR #246, opened 2019-11-30
Update to `bytes` 0.5

LucioFranco:
This PR updates bytes to `0.5`. This also bumps the MSRV to `1.39` since that is the MSRV of `bytes` 0.5. Mainly, this update is required for https://github.com/hyperium/tonic to update to `tokio` 0.2 and `hyper` 0.13.

Some notes:

-  `bytes` is renamed to `bytes1` in the main crate to allow the use of the the `pub mod bytes` within `src/encoding.rs`.
- Some unsafe code has been removed with the usage of `MaybeUninit`.

All tests pass but I am having some trouble compiling some of the benchmarks and conformance tests but I want to make sure we see no regressions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ os:
   - osx
 
 rust:
-  - 1.32.0
+  - 1.39.0
   - nightly
 
 script:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,7 +37,7 @@ default = ["prost-derive"]
 no-recursion-limit = []
 
 [dependencies]
-bytes = "0.4.7"
+bytes1 = { version = "0.5.2", package = "bytes" }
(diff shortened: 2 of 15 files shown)


dingelish:
Hi @LucioFranco , I think I fixed all the benchmarks and conformance tests. Please feel free to look at my commit!

https://github.com/dingelish/prost/commit/c92b47d55bc9b8310a1a03adb2d2f8a33eb73a0b

danburkert:
Overall looks great!  I see that in most Cargo.toml's you've set the bytes version to `0.5.2`; is the patch version necessary, or would `0.5` work?  I typically prefer not to set the patch version unless there's a specific bug being worked around (that was the case previously for `0.4.7`).

danburkert:
Landed via #251, thanks for the PR, @LucioFranco!

PR #250, opened 2019-12-13
Document generated methods

danburkert:
Extracted just the method getter/setter changes from #234, plus a few improvements.

### DIFF
--- a/prost-derive/src/field/map.rs
+++ b/prost-derive/src/field/map.rs
@@ -256,10 +256,14 @@ impl Field {
                 quote!()
             };
 
+            let get_doc = format!("Look up the given key in {:?}.", ident);
+            let insert_doc = format!("Inserts a key value pair into {:?}.", ident);
             Some(quote! {
+                #[doc=#get_doc]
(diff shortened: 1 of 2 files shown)


prost-derive/src/field/map.rs
```diff
@@ -256,10 +256,18 @@ impl Field {
                 quote!()
             };
 
+            let get_doc = format!(
+                "Returns the enum value for the corresponding key in `{}`, \
+                 or `None` if the entry does not exist or it is not a valid enum value.",
```
danburkert:
```suggestion
                 or `None` if the entry does not exist or is not a valid enum value.",
```

danburkert:
@dunmatt apologies, I merged this via 'squash', which I did not realize would overwrite the git author metadata.  I've corrected that mistake on master by forcibly rewriting the commits.

dunmatt:
Lol, I'm not worried about credit (but on behalf of those who do, thanks
for offering to fix it!), I care about things working well!

danburkert:
No worries, I personally feel strongly that the PR credit should be accurately reflected in the author metadata.  There are very few things which I'll rewrite history for, but this is one of them!

PR #253, opened 2019-12-18
Re-export `bytes` and update derive

### LINKED ISSUE #127: Prost should not require user to include a `bytes` dependency.
By doing [`extern crate bytes`](https://github.com/danburkert/prost/blob/e6aaddc259409db2193b4eae6ac064fdea319fb6/prost-derive/src/lib.rs#L178) in the generated derive, Prost requires the user to add a bytes dependency.

Instead, the prost crate should re-export bytes. I have done this in the past by having a hidden `codegen` module w/ the re-exports. [Example](https://github.com/carllerche/tower-web/blob/master/src/codegen/mod.rs).

LucioFranco:
This change removes the need to include `bytes` in your `Cargo.toml` when using the `prost-derive` macros.

Here is a preview of the docs.

image

Fixes #127 

cc @danburkert

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,6 @@ First, add `prost` and its public dependencies to your `Cargo.toml` (see
 ```
 [dependencies]
 prost = <prost-version>
-bytes = <bytes-version>
 # Only necessary if using Protobuf well-known types:
 prost-types = <prost-version>
 ```
--- a/prost-derive/src/lib.rs
+++ b/prost-derive/src/lib.rs
@@ -173,7 +173,7 @@ fn try_message(input: TokenStream) -> Result<TokenStream, Error> {
     let expanded = quote! {
         impl ::prost::Message for #ident {
             #[allow(unused_variables)]
-            fn encode_raw<B>(&self, buf: &mut B) where B: ::bytes::BufMut {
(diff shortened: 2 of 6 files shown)


danburkert:
Thanks for the PR.  This looks good to me in spirit, but I want to initially be a bit more conservative, and treat the re-export as an internal implementation detail.  As such, could you back out all changes except to `prost-derive/src/lib.rs`, and `src/lib.rs`, and then add a `#[doc(hidden)]` tag on the addition in `src/lib.rs`.

The primary reason is that this massively increases the surface area of the `prost` API if it's something that's publicly exposed and advertised to consumers.  I'd prefer to take it slow, and still maintain that if a consumer of `prost` needs to use `Buf`/`BufMut`/`Bytes` etc, they go through an explicit `bytes` crate dependency.

I'm positive on the cut down version of the PR because it does simplify at least two common scenarios: users of `tonic` and `tonic`-like frameworks, and crates which exist only to hold `prost` generated types, which will no longer need a direct `bytes` dependency.

LucioFranco:
@danburkert sounds good to me! I'll do that shortly.

LucioFranco:
Ok opened #254

PR #255, opened 2019-12-25
Well-Known-Types Any support with JSON serialization and deserialization

fdeantoni:
I have a project where we have receive following proto message:

```proto
message Request {
    string requestId = 1;
    google.protobuf.Any payload = 2
}
```

The `payload` field can contain any proto message, for example:

```proto
package test;

message Foo {
    string data = 1;
    google.protobuf.Timestamp timestamp = 2;
}
```

The request message, with something like Foo packed in it as payload, can be received as binary, or as JSON. The JSON format would look as follows:

```
{
  "requestId": "message1",
  "payload": {
    "@type": "type.googleapis.com/test.Foo",
    "data": "Hello World!",
    "timestamp": "2019-11-29T02:11:40.066Z"
  }
}
```

To support this, the following needed to be added:
* Support packing and unpacking of messages into an Any
* Support binary serializing and deserializing of an Any
* Support JSON serialization and deserialization of an Any

The changes in this commit add support for the above by using the `typetag` 
and `mopa` crates.

To support JSON serialization and deserialization, a new `include_serde`
property needs to be set in the prost_build `build.rs` configuration:

```rust
fn main() {
    let mut prost_build = prost_build::Config::new();
    prost_build.include_serde();
    prost_build.compile_protos(&["src/frontend.proto", "src/backend.proto"],
                               &["src"]).unwrap();
}
```

This will then add the necessary macro properties for `prost-derive` to use
to enable `Message` serialization and deserialization with the `serde`
framework.

All `Message` structs can then be packed into an Any type, and correctly
serialized to JSON. Likewise they can also be deserialized again to an
Any struct, and the original message can be `unpacked` to either a
generic trait object, or the concrete struct.

This pull request does involve a lot of changes and requires two extra crates as
a dependency, but so far it works well in our project. Perhaps it can be useful
for others as well.


danburkert:
Hi @fdeantoni, thanks for the PR.  Unfortunately this is too large of a change to be reviewed or considered in one chunk.  From a quick skim I see some additions that can be likely be split apart and considered independently:

* `prost-build` support for outputting serde annotations.
* The addition of the `package` annotation.
* The changes to the `Message` trait to allow downcasting, including adding `mopa` as a public dependency.
* Changes to `prost-types`, particularly to `Value` to allow better transformation to/from Rust stdlib types.

It would be very useful if you could write up an issue detailing the design of how the pieces work individually, and how they are useful in combination.  Somewhat like a Rust RFC, if you will.  Even then I think it would be better to land the pieces of the implementation in isolation, since it's significantly easier to review.

fdeantoni:
Hi @danburkert, many thanks for the initial review! You are right that this can be cut up into smaller parts. The division you mentioned is more or less correct. As you suggest I will open up an issue detailing the changes there with separate PRs covering the items there. 

The easiest one to start with are the convenience methods for the Well-Known-Types as the changes are localized to one file (`prost-types/lib.rs`). From there we can move on to the addition of the package annotation and downcasting capability for the `Any` type. Once that is done we can add the Serde stuff so that an Any can be properly serialized and deserialized to/from JSON.

fdeantoni:
Closed in favour of #277.

Issue #267, opened 2020-01-16
Stack overflow when parsing message


dbrgn:
When parsing certain messages, the process aborts with a stack overflow.

I made a [reproducer](https://github.com/dbrgn/prost-stackoverflow-reproducer):

```
prost-stackoverflow-reproducer$ cargo run
     Running `target/debug/prost-stackoverflow-reproducer`
Hello, world!

thread 'main' has overflowed its stack
fatal runtime error: stack overflow
Aborted (core dumped)
```

Potentially the data requests allocation of a buffer that's larger than the actually available data? (23 KiB input data is not *that* large...)

(Found through [afl.rs](https://github.com/rust-fuzz/afl.rs))

danburkert:
Wow nice, thanks for the report.  Getting a stack trace, if possible, would be super great if you are still looking at this (otherwise no worries, it should be easy to repro thanks to your repo).  We have some fuzzers in-tree, but they haven't dug this up.  That's something to be investigated as well, or maybe we can just upstream you're test into `prost`.

danburkert:
I found it, I believe it's this recursive call within `skip_field`: https://github.com/danburkert/prost/blob/master/src/encoding.rs#L402.  It should be straightforward to restructure that as a loop instead of recursion.  Interestingly it doesn't manifest in release mode.

danburkert:
>  Interestingly it doesn't manifest in release mode.

It likely would with longer input.

danburkert:
@dbrgn thanks again for the report, and thorough repro.  Fix has been released in 0.6.1; more details are in the release notes: https://github.com/danburkert/prost/releases/tag/v0.6.1.  Additionally I filed #270 to track better fuzz testing through afl.

dbrgn:
Great, thank you for the very fast response! :slightly_smiling_face: 

I will also file a RUSTSEC advisory, since this could be used for DoS if untrusted input is fed to the parser.

tarcieri:
Note that in absence of [stack probes on architectures like ARM](https://github.com/rust-lang/rust/issues/43241), stack overflow is unsound and can result in potential memory corruption (or even RCE)

dbrgn:
Oh, that's a good point! I wasn't aware of that.

I'll file a RUSTSEC advisory. In that case, should I put it both in "denial-of-service" and "memory-corruption" categories?

tarcieri:
Yup, sounds good.

tarcieri:
[...]

PR #272, opened 2020-01-20
Add afl fuzz tests

### LINKED ISSUE #270: afl fuzz tests
#267 shows that the current in-tree fuzzing based on `cargo-fuzz` has some blindspots.  Investigate adding afl-based fuzzing to the in-tree test suite.

dbrgn:
Here's a setup for fuzzing prost with afl.

For details, see `FUZZING.md`.

I ran it against prost 0.6.0 to verify that it works, the stack overflow from #267 was found within about 15 minutes :slightly_smiling_face: 

Fixes #270.


dbrgn:
By the way, I ran 4 fuzzer instances on master for 5.5 hours and they didn't find any crashes.

dbrgn:
Still interested in this?

LucioFranco:
Sorry for the delay on this. @dbrgn I'd love to understand the benefits of afl over the current fuzzing approach? If it makes sense to add then I think we should!

dbrgn:
@LucioFranco AFL and libfuzzer use slightly different approaches, so in my view more fuzzing is always helpful 🙂

LucioFranco (approved):

Issue #277, opened 2020-02-04
WKT Any support with JSON serialization/deserialization


fdeantoni:
This is an umbrella issue to implement the changes needed to achieve proper support for the `Any` type with serialization and deserialization similar to the Protobuf libraries in Go and Java. For example, in the [google documentation about the Any type](https://developers.google.com/protocol-buffers/docs/proto3#any) it states the following:

> Different language implementations will support runtime library helpers to pack and unpack Any values in a typesafe manner – for example, in Java, the Any type will have special pack() and unpack() accessors, while in C++ there are PackFrom() and UnpackTo() methods...

Supporting `pack()` and `unpack()` provides many benefits. One use case is to pack a [Value](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.Value) into an `Any` to store a dynamically typed value (e.g. an arbitrary JSON structure). For example, assume the following message:

```rust
message Request {
    string requestId = 1;
    google.protobuf.Any payload = 2
}
```

This is a generic Request message that can contain a dynamically typed payload, such as:
```rust
package test;

message Foo {
    string data = 1;
    google.protobuf.Timestamp timestamp = 2;
}
```   

When using the protobuf Java or C++ library, you can "pack" this message into the Request message's payload property and serialize it. The receiver of the `Request` message can then "unpack" the payload and retrieve a `Foo` object.

As protobuf also supports JSON serialization/deserialization, the [documentation specifies](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.Any) that the `Request` message containing the `Foo` payload should be serialized as follows:

```json
{
  "requestId": "message1",
  "payload": {
    "@type": "type.googleapis.com/test.Foo",
    "data": "Hello World!",
    "timestamp": "2019-11-29T02:11:40.066Z"
  }
}
```    

To support the above, the following functionality needs to be added to Prost:
1. Add convenience methods to the [Well-Known-Types](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf) to conveniently pack them into an `Any`. 
2. Support "packing" and "unpacking" of an Any. 
3. Support unpacking an Any to a ::prost::Message trait object

## 1. Add convenience methods to WKT
[...]

PR #278, opened 2020-02-05
Additional convenience methods to convert from/to Rust standard types.

fdeantoni:
Part 1 of #277, adding convenience methods to the Well-Known-Types.


fdeantoni:
No need to merge this if #336 is merged instead. With #336 this functionality can be provided by an external crate.

PR #286, opened 2020-02-24
Add no_std support on top of 0.6.1

dflemstr:
This is an extension of https://github.com/danburkert/prost/pull/215 with the addition that the current master branch has been merged in.

I also created https://github.com/cbeck88/prost/pull/1, so either @cbeck88 can merge that change into his PR, or this PR can be merged directly.  It really only depends on who sees this first, I guess :)


cbeck88:
Thank you! I will look at this today!

cbeck88:
I read through the diff, this looks great to me. I'm going to try making our project point at cbeck88/#1 and see what happens, if tests pass, I will merge into my PR. Thank you for this!

cbeck88:
@dflemstr I'm getting `no_std` build errors, I'm gonna try to dig in it and possibly send you a patch to go on top of this soon

cbeck88:
@dflemstr I ran cargo expand in our build, and I see the following:

```
       #[allow(unused_variables)]
        fn merge_field<B>(
            &mut self,
            tag: u32,
            wire_type: ::prost::encoding::WireType,
            buf: &mut B,
            ctx: ::prost::encoding::DecodeContext,
        ) -> ::std::result::Result<(), ::prost::DecodeError>
        where
            B: ::prost::bytes::Buf,
        {
```

So I think I'm going to try to track that down in `prost_derive` code and replace `::std::result::Result` with `::core::result::Result`

cbeck88:
@dflemstr I guess there's some stuff about `alloc` attributes that happened that I didn't follow, I'm gonna read again more carefully and fix the stuff I pushed

danburkert:
I believe my question in https://github.com/danburkert/prost/pull/215#issuecomment-575934428 still stands - why isn't this switching to `alloc` at all times for all collections except HashMap?  If possible I'd greatly prefer an approach that doesn't rely on cargo features.

danburkert:
I suppose the error impls and timestamp helpers will need to rely on a cargo feature, but essentially I want to minimize the occurances of things being cargo feature dependent, because the rules about breaking APIs with features are non-intuitive.

src/message.rs
```diff
@@ -80,16 +82,18 @@ pub trait Message: Debug + Send + Sync {
...
         Self: Default,
     {
         let mut message = Self::default();
-        Self::merge(&mut message, &mut buf).map(|_| message)
+        Self::merge(&mut message, buf).map(|_| message)
     }
 
     /// Decodes a length-delimited instance of the message from the buffer.
+    ///
+    /// The entire buffer will be consumed.
```
danburkert: This addition is not correct

dflemstr:
Looks like the commits in this PR were merged into #215, let's continue discussion there

PR #290, opened 2020-03-15
prost-build: Support generating single include file

SkamDart:
In order to support use-cases where many protocol buffer
packages are parsed and generated, support the configuration
containing an optional `include_file()` which allows the build
script to name a file to be created to automatically `include!()`
all the generated content with the right module structure.

Signed-off-by: Daniel Silverstone <dsilvers@digital-scurf.org>

### DIFF
--- a/prost-build/src/lib.rs
+++ b/prost-build/src/lib.rs
@@ -114,7 +114,7 @@ use std::collections::HashMap;
 use std::default;
 use std::env;
 use std::fs;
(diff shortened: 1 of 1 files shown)


SkamDart:
Thanks to @kinnison for doing the real work here.
Just getting this through the finish line.

https://github.com/danburkert/prost/pull/123#issuecomment-584746018

PR #291, opened 2020-03-23
Fix code generation for code blocks in prost-build

Xaeroxe:
If the generated doc comments include a code block rustdoc will erroneously try and execute them.

That's silly as these blocks almost never contain valid Rust, so this PR marks these blocks as text.

### DIFF
--- a/prost-build/Cargo.toml
+++ b/prost-build/Cargo.toml
@@ -26,3 +26,4 @@ which = { version = "3", default-features = false }
 
 [dev-dependencies]
 env_logger = { version = "0.7", default-features = false }
+pretty_assertions = "0.6"
--- a/prost-build/Cargo.toml
+++ b/prost-build/Cargo.toml
@@ -11,6 +11,7 @@ edition = "2018"
 
 [dependencies]
 bytes = { version = "0.5", default-features = false }
+comrak = "0.7"
 heck = "0.3"
 itertools = "0.9"
 log = "0.4"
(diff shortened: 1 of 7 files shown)


Xaeroxe:
Btw it's worth noting this PR now includes breaking changes for tonic-build crate users. Not sure if that's a use case this project is trying to accomodate or not, but I will point it out.

Xaeroxe:
I'm not sure what's up with CI, unit tests pass locally.

Xaeroxe:
I'm not sure how to update the definitions such that the bootstrap test passes, nor am I sure I should do that within the scope of this PR.

Xaeroxe:
@danburkert This is ready for review

Xaeroxe:
@danburkert Hey, it's been a few months, just want to make sure this didn't slip off your radar.

Xaeroxe:
> Btw it's worth noting this PR now includes breaking changes for prost-build crate users. Not sure if that's a use case this project is trying to accomodate or not, but I will point it out.

Actually this isn't even true.

Xaeroxe:
It just occurred to me that the problem this is trying to fix is sort of a security vulnerability, due to there being arbitrary code execution. I guess the question becomes whether or not the protobuf being compiled is considered "trusted" for the particular use case.

Xaeroxe:
This PR fell to the bottom of the list, so I'm going to try re-opening it so it's grouped with all the other PRs that are passing near the top.

PR #295, opened 2020-04-01
reordered attribute transcription and added docs to resolve #273 and #233

### LINKED ISSUE #273: prost-build define order in type attributes
Hi,

I'm using prost-build with a custom type attribute,
`prost_build.type_attribute(".", "#[pyclass]");`
but when I run it, I get:
`error: macro attributes must be placed before #[derive]`
and when I see the output I get:
```
#[pyclass]
pub struct ...
```
Is there a way to establish the order of the attributes?

Thanks!

dunmatt:


dunmatt:
Made a git mistake, will resubmit

PR #296, opened 2020-04-01
reordered attribute transcription and added docs to resolve #273 and #233

### LINKED ISSUE #273: prost-build define order in type attributes

dunmatt:
In addition to resolving #233 and #273 , this also adds `Debug` to `Config`, because I found myself needing that during some debugging.


prost-build/src/lib.rs
```diff
@@ -144,7 +144,7 @@ type Module = Vec<String>;
 /// implementation of the generated trait in the application code and plugs it into the framework.
 ///
 /// Such framework isn't part of Prost at present.
-pub trait ServiceGenerator {
+pub trait ServiceGenerator: std::fmt::Debug {
```
danburkert, line 147: This is a breaking change, so I'd prefer not to include it.

dunmatt: What does it break?  I mean, I can take it out if need be, but I really don't think it can break anything.

danburkert: If this were released, existing `ServiceGenerator` impls which don't impl `Debug` would fail to compile, thus it's a breaking change.  Yes, please take it out.

dunmatt: Fair enough, would you be open to a manual implementation that doesn't have that problem?

danburkert: What is a manual implementation?

dunmatt: `impl Debug for Config { ... }`

danburkert: The breaking change comment applies only to the highlighted line, which adds a trait superbound.  Adding `#[derive(Debug)]` to `Config` is fine.

dunmatt: It can't be derived because deriving Debug requires the breaking change you're talking about, that's why I'm suggesting doing it manually.

danburkert: Adding a trait superbound and adding a Debug impl for a struct are two different things.  My comments in this thread only apply to the former.

danburkert: ah ok I see, yeah you can add a manual Debug impl for Config, if the derive impl is assuming that supertrait bound.

danburkert: (I thought it just added a where clause to the impl though?)

dunmatt: I don't think it's quite that easy because it's using `dyn` dispatch

dunmatt:
Do we care that the Apple build failed?

dunmatt:
Friendly (totally-not-infected) ping!

danburkert:
Going to re-open momentarily with a new PR to add some tweaks.

danburkert:
Re-opened as #323.

PR #297, opened 2020-04-01
fix Timestamp <-> SystemTime conversions

danburkert:
I must have been under the impression that SystemTime couldn't handle
datetimes before the Unix epoch back when I wrote the conversion code in
2017. That is wrong, and so was the code.

BREAKING CHANGE: this is a breaking change to prost-types, because the
Timestamp TryFrom<SystemTime> impl is now the default impl, with a
different error type.


danburkert:
This also makes the `normalize` methods public, since they may be useful in niche scenarios.

PR #309, opened 2020-04-29
Add feature to prost-build to allow runtime specification of protoc

dfreese:
Opening this as a way of opening the discussion about an issue that I
was running into.  If you would prefer a different approach, I'm happy
to discuss that.

Within multi-language environments, such as bazel, cargo isn't always
able to drive the build process.  However, prost and prost-build
provides an amazing foundation of which to build off.  The general idea
of this commit was to open up prost-build to having tools built off of
it, while still preserving previous behavior.

Currently, if PROTOC or PROTO_INCLUDE are not set at build-time, then
the build fails.  This commit introduces feature, runtime-protoc. This
hides env! invocations when enabled, allowing compilation, under the
assumption that the protoc location will be provided at runtime.  This
wouldn't be my first preference, but I was prioritizing the build-break
if the environment flags were not specified.  I would generally have
preferred grabbing those with option_env! macros and then trying to
gracefully handle the error at runtime protoc has not been provided at
build time or runtime similar, since this is already covered by tests.
That would allow removal of the feature flag.  It would also make the
code a little cleaner, since there's not a great way to do conditionals
with cfg (see rust-lang/rust#71679).

### DIFF
--- a/prost-build/Cargo.toml
+++ b/prost-build/Cargo.toml
@@ -9,6 +9,11 @@ readme = "README.md"
 description = "A Protocol Buffers implementation for the Rust Language."
 edition = "2018"
 
+[features]
+# Modify prost-build to be built and run as an executable outside of the normal
+# build.rs environment.  Requires explicitly setting the protoc path at runtime.
+runtime-protoc = []
+
 [dependencies]
 bytes = { version = "0.5", default-features = false }
 heck = "0.3"
--- a/prost-build/src/lib.rs
+++ b/prost-build/src/lib.rs
@@ -190,6 +190,8 @@ pub struct Config {
     strip_enum_prefix: bool,
     out_dir: Option<PathBuf>,
     extern_paths: Vec<(String, String)>,
+    #[cfg(feature = "runtime-protoc")]
+    protoc: Option<PathBuf>,
 }
 
 impl Config {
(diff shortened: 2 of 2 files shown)


dfreese:
rebased on top of #314

kornholi:
As far as the Bazel integration goes, it'd be better to make `generate()` public and consume `FileDescriptorSet`s from [proto_library](https://docs.bazel.build/versions/master/be/protocol-buffer.html#proto_library) targets instead of calling `protoc` directly. I ran into this recently while setting up Tonic.

dfreese:
@kornholi I didn't realize proto_library provided that.  Thanks for point that out!

dfreese:
@kornholi is any of the work you did with bazel/tonic open source?

kornholi:
It's not, but I think we're very close with your work in #313 and some PRs open in `rules_rust`/`cargo-raze`. Let me see if I can get it working in a clean Bazel environment.

danburkert:
Now that #283 and #324 have landed is this still needed?  I know this PR goes further in that it provides a protoc plugin, but not sure if that's still advantageous for bazel/buck?

danburkert:
Sorry I got my PRs mixed up, that's not what this PR does.  So now it should be possible to specify the protoc through the runtime `PROTOC` env var, is that good enough?

dfreese:
@danburkert , no worries!  Thanks for taking a look at these.

I had continued the discussion [here](https://github.com/danburkert/prost/pull/307#issuecomment-626828916) over on #307, following your reply.

dfreese:
The PROTOC environment variable override in https://github.com/danburkert/prost/pull/324 unblocked my use case.  While I still like the more explicit side of this design, I don't think it would make sense to have both features.

Issue #256, opened 2019-12-26
Custom options?


lquerel:
Are the custom options supported?
I tried options and uninterpreted_option without success.
Thanks 
Laurent

killercup:
I'm into this as well at the moment. I followed [the proto3 docs](https://developers.google.com/protocol-buffers/docs/proto#customoptions) to add a custom option and prost fails to read in the files when I don't import the correct defintion so I know it's doing _something_; but nevertheless `uninterpreted_option` stays empty.

killercup:
So looking at a `test.proto` like this

```proto
syntax = "proto3";
package test;

import "google/protobuf/descriptor.proto";

extend google.protobuf.MessageOptions {
  int32 message_opt1 = 7739036;
}

message Test {
  option (message_opt1) = 420000;
  option no_standard_descriptor_accessor = true;
  
  string content = 1;
}
```

and calling `protoc --json_out=./ ./test.proto` (`libprotoc 3.11.4`, using [protoc-gen-json](https://github.com/sourcegraph/prototools/blob/b54ce590735f46af87984528ed4853827ef190cb/README.json.md)), the only output I get for the `Test` message is this:

```json
{
  "name": "Test",
  "field": [
    {
      "name": "content",
      "number": 1,
      "label": 1,
      "type": 9,
      "json_name": "content"
    }
  ],
  "options": {
    "no_standard_descriptor_accessor": true
  }
}
```

To me, this looks like there are no `uninterpreted_option`s collected? Is this a bug in `protoc`?

- - -

[Full JSON output](https://gist.github.com/killercup/c3e108647811942bc6c16c1554e6e640) -- quite long

danburkert:
`prost` currently doesn't expose message descriptors in any interface except `ServiceBuilder`, so I'm curious what the usecase for options is?  What does supporting them mean in the context of `prost`?

killercup:
Ah, I should've probably explained that. I've patched prost-build locally to also give me a `Vec<prost_types::FileDescriptorProto>` to generate some helper data structures. I'd like to custom options to have some metadata for this on my messages.

danburkert:
I see.  It's a longer-term goal to expose message descriptors in generated code in some way, but it's going to need some design.  Having a PR to go off might be very helpful!

Back to the question at hand - `prost` isn't doing anything that might strip options from the descriptors, afaik.  You can look and see exactly how `prost-build` is invoking `protoc` to try and isolate where they might be getting dropped.

[...]

PR #317, opened 2020-05-07
Unknown fields

wildarch:
As proposed in #256.

This adds an `unknown_fields` flag to `prost-build` that adds unknown field support to specific proto messages. In the structs for these message it adds an extra field:
```rust
struct MyMessage {
  // ...
  protobuf_unknown_fields: Vec<UnknownField>,
}
```

This works both with deserialization and serialization: Any fields not recognized while parsing are put in the unknown fields, and any fields added here will be included when serializing.

I think the main use case for this feature would be to support custom options by enabling this flag for things like [`FieldOptions`](https://docs.rs/prost-types/0.6.1/prost_types/struct.FieldOptions.html).

There is still some work to be done on this PR in terms of tests and documentation, but I'm hoping to get @danburkert's blessing before finishing that 😄.


wildarch:
CC @tiziano88

tests/src/unknown_fields.rs
```diff
@@ -0,0 +1,76 @@
...
+            value: b"extra".to_vec()
+        }]
+    );
+}
+
+#[test]
+fn test_serialize_unknown_field() {
+    let message = unknown_fields::MessageWithUnknownFields {
+        normal_field: "normal".to_string(),
+        protobuf_unknown_fields: vec![UnknownField {
```
tiziano88, line 31: Could you add a test in which the protobuf message has an actual field called `protobuf_unknown_fields`, of whatever type? I think we should clarify what behaviour we expect in that case, which I guess will boil down to one of the following:

- fail early with a friendly error message
- somehow try to escape the field name so that it does not conflict with this; I expect there must be a list of identifiers to avoid (e.g. Rust keywords) that are already escaped by prost -- or does it just ignore them?
- give this field a name that cannot possibly clash with any generated field, e.g. `_protobuf_unknown_fields` or similar.

wildarch: I did a quick test, and it prost removes leading underscores in protobuf field names, so I think we could name the field `_unknown_fields`?

wildarch: As a sidenote, if you define a message like this:
```
message Test {
  optional string _weird_field = 1;
  optional string weird_field = 2;
}
```

Prost fails with:
```
4 |     pub weird_field: ::std::option::Option<std::string::String>,
  |     ----------------------------------------------------------- `weird_field` first declared here
5 |     #[prost(string, optional, tag="2")]
6 |     pub weird_field: ::std::option::Option<std::string::String>,
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ field already declared
```

Frando:
I was testing this today as a step towards custom options (#256). I noted that there's a bug currently for varint fields. They end up as empty because the `len` that is put into the unknown field value is always 0.

I fixed it [here](https://github.com/Frando/prost/commit/488beaab853188d58581fb26239e967dc6207b1c). I don't know though how to "PR against this PR" - @wildarch, can you merge it manually into this PR, or what's the recommended way?

wildarch:
Thanks for fixing that @Frando! I can pull your changes in 😄. But before I do that and spend more time on this PR I would like some feedback from the maintainers of the crate to see if they are interested in merging this eventually.

Frando:
@danburkert how do you feel about this PR, are there things that should be changed for this to be accepted? I'm using this in [hrpc-build](https://github.com/Frando/hrpc-rs) to specify service and method IDs through options. I'd like to publish it on crates.io sooner or later, thus would look into getting this PR merged (plus the application in [this commit](https://github.com/Frando/prost/commit/b0b20b44d38bde54c95c09d14f8426284a0d81b0).

Frando:
Hei @danburkert, do you think this can move forward, or do you recommend another approach?

bfrog:
I'm really interested in seeing this move forward, as I'd like to customize codegen based on protobuf options

yusdacra:
This would be really useful, what's needed to move this forwards?

Frando:
@danburkert do you have a comment on the future for this PR? Is there a way to get it merged, do you want to see further changes? I'm using it in [hrpc](https://github.com/Frando/hrpc-rs) and would like to release on crates.io soonish, so I'm thinking if I will publish a fork of prost on crates.io or if this can be included in mainline (which I'd prefer of course).

hollinwilkins:
What would need to happen to get this merged? Does someone need to go in and resolve conflicts? This would be a very welcome feature update :)

prost-derive/src/lib.rs
```diff
@@ -73,14 +73,19 @@ fn try_message(input: TokenStream) -> Result<TokenStream, Error> {
...
+
     // We want Debug to be in declaration order
     let unsorted_fields = fields.clone();
 
     // Sort the fields by tag number so that fields will be encoded in tag order.
     // TODO: This encodes oneof fields in the position of their lowest tag,
     // regardless of the currently occupied variant, is that consequential?
     // See: https://developers.google.com/protocol-buffers/docs/encoding#order
-    fields.sort_by_key(|&(_, ref field)| field.tags().into_iter().min().unwrap());
+    fields.sort_by_key(|&(_, ref field)| field.tags().into_iter().min().unwrap_or(std::u32::MAX));
```
caspermeijn, line 88: So unknown fields are always sorted last, why is that decision made? I would like that to be more explicit in the code.

prost-derive/src/lib.rs
```diff
@@ -103,20 +108,24 @@ fn try_message(input: TokenStream) -> Result<TokenStream, Error> {
...
-        quote! {
-            #(#tags)* => {
-                let mut value = &mut self.#field_ident;
-                #merge.map_err(|mut error| {
-                    error.push(STRUCT_NAME, stringify!(#field_ident));
-                    error
-                })
-            },
+        if let Field::UnknownFields(_) = field {
+            quote! {}
```
caspermeijn, line 112: Why this exception? The `merge` function actually exists, so this exception is not needed. I expect the compiler to optimize the unused codepaths.

prost-derive/src/lib.rs
```diff
@@ -175,6 +184,16 @@ fn try_message(input: TokenStream) -> Result<TokenStream, Error> {
         quote!(f.debug_tuple(stringify!(#ident)))
     };
 
+    let unknown_field_handler = if has_unknown_fields_accessor {
```
caspermeijn, line 187: I like this! A simple if statement that decides the fate of unknown fields

src/encoding.rs
```diff
@@ -391,6 +392,52 @@ pub fn skip_field<B>(
 ) -> Result<(), DecodeError>
 where
     B: Buf,
+{
+    handle_unknown_field(
```
caspermeijn, line 396: Good to see the code reuse!

src/message.rs
```diff
@@ -164,3 +164,16 @@ where
...
+///
+/// Structs generated by `prost-build` with the `unknown_fields` flag enabled have a field
+/// `protobuf_unknown_fields: Vec<UnknownField>` that gives access to any fields that were not
+/// recognized.
+/// Any values added to this field will also be present in serialized form.
+#[derive(Debug, Default, Clone, PartialEq)]
+pub struct UnknownField {
+    pub tag: u32,
+    /// The raw field value.
+    pub value: Vec<u8>,
```
caspermeijn, line 178: The C++ implementation handles unknown fields differently. It actually parses the fields. Why is this approach chosen? Can you explain the advantages of both approaches? https://protobuf.dev/reference/cpp/api-docs/google.protobuf.unknown_field_set/#UnknownField

tests/src/build.rs
```diff
@@ -85,6 +85,11 @@ fn main() {
         .compile_protos(&[src.join("well_known_types.proto")], includes)
         .unwrap();
 
+    config
+        .unknown_fields_message("unknown_fields.MessageWithUnknownFields")
```
caspermeijn, line 89: This string should start with a dot, as that is the match notation used in other `Config` functions.

tests/src/unknown_fields.rs
caspermeijn: I expect a message with a conflicting field name to give a compiler error.

caspermeijn:
I understand this is an old PR. If the original author is still interested, please rebase the PR. Is someone else want to work on it, please open a new PR with all comments processed.

PR #320, opened 2020-05-10
CI: update to checkout@v2

danburkert:
It appears PR retries were failing due to (presumably)
https://github.com/actions/checkout/issues/23.


danburkert:
This actually appears to speed up CI times, since the test portion still takes ~8m, but there's no longer ~2m dedicated to storing the cache 🤷‍♂️

PR #314, opened 2020-05-05
Remove cache action step that is causing problems for CI

dfreese:
I've noticed a number of github actions reporting failing or cancelled
on the cache step.  This removes them, for now, to provide a better
signal for the project, until why the cache action fails can be
determined.

Builds on PR #312.

PR #322, opened 2020-05-10
CI: disable Cargo target directory caching

danburkert:
This is #314, but disables for all workflows.  I'm interested in getting some timing information; if this doesn't have much of an impact on CI times then I'm happy to merge, otherwise it'd be good to dig deeper into why the caching failures are occurring.


danburkert:
OK well it appears this significantly improves CI times 🤷‍♂️

PR #328, opened 2020-05-11
Validate Timestamp conversion from SystemTimes before UTC Epoch time

joseph-wakeling-frequenz:
The protobuf spec for `Timestamp` states that:

    negative second values with fractions must still have
    non-negative nanos values that count forward in time

... but in v0.6.1 and earlier of `prost_types`, this is handled wrongly when converting to `SystemTime`, with the `nanos` value contributing to the `Duration` that is to be _subtracted_ from the `UNIX_EPOCH`:
https://github.com/danburkert/prost/blob/v0.6.1/prost-types/src/lib.rs#L142-L145

```Rust
let mut duration = Duration {
    seconds: -timestamp.seconds,
    nanos: timestamp.nanos,
};
```

So, for example `Timestamp { seconds: -20, nanos: 999_999_999 }` would result in a `Duration { seconds: 20, nanos: 999_999_999 }` to subtract from `UNIX_EPOCH`, instead of the correct value of 19 seconds + 1 nano before `UNIX_EPOCH`.

The bug was fixed in commit 4ad5e3451fcd2e2192406351671a962e2e2be909 (https://github.com/danburkert/prost/pull/297), which replaces the old `TryFrom` implementation with a full `From<Timestamp> for SystemTime`, and adds `From<SystemTime> for Timestamp` support for pre-UTC Epoch time `SystemTime` values.  However, the existing unittests do not include any specific validation for timestamps earlier than the UTC epoch time.

This patch therefore adds a set of representative test cases which make sure that the protobuf spec is respected.  This should prevent the risk of regressions in future.

Part of https://github.com/danburkert/prost/issues/327.

### DIFF
--- a/prost-types/src/lib.rs
+++ b/prost-types/src/lib.rs
@@ -159,7 +159,7 @@ impl From<Timestamp> for std::time::SystemTime {
 
 #[cfg(test)]
 mod tests {
-    use std::time::SystemTime;
+    use std::time::{Duration, SystemTime, UNIX_EPOCH};
 
     use proptest::prelude::*;
 
@@ -173,5 +173,74 @@ mod tests {
         ) {
             prop_assert_eq!(SystemTime::from(Timestamp::from(system_time)), system_time);
         }
+    }
+
+    #[cfg(feature = "std")]
+    #[test]
+    fn check_timestamp_negative_seconds() {
+        // Representative tests for the case of timestamps before the UTC Epoch time:
+        // validate the expected behaviour that "negative second values with fractions
+        // must still have non-negative nanos values that count forward in time"
+        // https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.Timestamp
+        //
+        // To ensure cross-platform compatibility, all nanosecond values in these
+        // tests are in minimum 100 ns increments.  This does not affect the general
+        // character of the behaviour being tested, but ensures that the tests are
+        // valid for both POSIX (1 ns precision) and Windows (100 ns precision).
+        assert_eq!(
+            Timestamp::from(UNIX_EPOCH - Duration::new(1_001, 0)),
+            Timestamp {
+                seconds: -1_001,
+                nanos: 0
+            }
+        );
+        assert_eq!(
+            Timestamp::from(UNIX_EPOCH - Duration::new(0, 999_999_900)),
+            Timestamp {
+                seconds: -1,
+                nanos: 100
+            }
+        );
+        assert_eq!(
+            Timestamp::from(UNIX_EPOCH - Duration::new(2_001_234, 12_300)),
+            Timestamp {
+                seconds: -2_001_235,
+                nanos: 999_987_700
+            }
+        );
+        assert_eq!(
+            Timestamp::from(UNIX_EPOCH - Duration::new(768, 65_432_100)),
+            Timestamp {
+                seconds: -769,
+                nanos: 934_567_900
+            }
+        );
+    }
+
+    #[cfg(all(unix, feature = "std"))]
+    #[test]
+    fn check_timestamp_negative_seconds_1ns() {
+        // UNIX-only test cases with 1 ns precision
+        assert_eq!(
+            Timestamp::from(UNIX_EPOCH - Duration::new(0, 999_999_999)),
+            Timestamp {
+                seconds: -1,
+                nanos: 1
+            }
+        );
+        assert_eq!(
+            Timestamp::from(UNIX_EPOCH - Duration::new(1_234_567, 123)),
+            Timestamp {
+                seconds: -1_234_568,
+                nanos: 999_999_877
+            }
+        );
+        assert_eq!(
+            Timestamp::from(UNIX_EPOCH - Duration::new(890, 987_654_321)),
+            Timestamp {
+                seconds: -891,
+                nanos: 12_345_679
+            }
+        );
     }
 }


danburkert:
This is excellent, thanks @joseph-wakeling-frequenz .  Are you working on a fix?  If not I'll take a crack at it, at least before doing the next release.

joseph-wakeling-frequenz:
@danburkert thanks!  About fixes: the bug itself is actually fixed in current master (it was addressed by https://github.com/danburkert/prost/pull/297), so it depends on your plans for a next release.  If you'd like a patch to the 0.6.x release series that can be released as v0.6.2, then I'm happy to submit that, but can you advise on how I should best do that given that `master` is now well ahead of 0.6.1?

OTOH the test failures in the current patch appear to be related to platform specific implementations of `SystemTime`.  Specifically on the windows build:

```
---- tests::check_timestamp_negative_seconds stdout ----
thread 'tests::check_timestamp_negative_seconds' panicked at 'assertion failed: `(left == right)`
  left: `Timestamp { seconds: -1, nanos: 100 }`,
 right: `Timestamp { seconds: -1, nanos: 1 }`', prost-types\src\lib.rs:192:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

I'd overlooked that on Windows the smallest `SystemTime` interval is 100 ns rather than 1 ns.  See:
https://github.com/rust-lang/rust/blob/09c817eeb29e764cfc12d0a8d94841e3ffe34023/src/libstd/sys/windows/time.rs#L20-L23
https://docs.microsoft.com/en-us/windows/win32/api/minwinbase/ns-minwinbase-filetime

So these tests can be rewritten to use 100-nsec smallest increments.  It doesn't make any difference to the behaviour we actually want to test.  I'll submit an updated patch shortly.

joseph-wakeling-frequenz:
<details><summary>I've updated the test cases to split between (i) general cases with 100 ns precision and (ii) POSIX-only cases with 1ns precision.  This should resolve the Windows CI failure.</summary>

```diff
diff --git a/prost-types/src/lib.rs b/prost-types/src/lib.rs
index 1f28a00..366d295 100644
--- a/prost-types/src/lib.rs
+++ b/prost-types/src/lib.rs
@@ -178,10 +178,15 @@ mod tests {
     #[cfg(feature = "std")]
     #[test]
     fn check_timestamp_negative_seconds() {
-        // representative tests for the case of timestamps before the UTC Epoch time:
+        // Representative tests for the case of timestamps before the UTC Epoch time:
         // validate the expected behaviour that "negative second values with fractions
         // must still have non-negative nanos values that count forward in time"
         // https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.Timestamp
+        //
+        // To ensure cross-platform compatibility, all nanosecond values in these
+        // tests are in minimum 100 ns increments.  This does not affect the general
+        // character of the behaviour being tested, but ensures that the tests are
+        // valid for both POSIX (1 ns precision) and Windows (100 ns precision).
         assert_eq!(
             Timestamp::from(UNIX_EPOCH - Duration::new(1_001, 0)),
             Timestamp {
@@ -189,6 +194,33 @@ mod tests {
                 nanos: 0
             }
         );
+        assert_eq!(
+            Timestamp::from(UNIX_EPOCH - Duration::new(0, 999_999_900)),
+            Timestamp {
+                seconds: -1,
+                nanos: 100
+            }
+        );
+        assert_eq!(
+            Timestamp::from(UNIX_EPOCH - Duration::new(2_001_234, 12_300)),
+            Timestamp {
+                seconds: -2_001_235,
+                nanos: 999_987_700
+            }
+        );
+        assert_eq!(
+            Timestamp::from(UNIX_EPOCH - Duration::new(768, 65_432_100)),
+            Timestamp {
+                seconds: -769,
+                nanos: 934_567_900
+            }
+        );
+    }
+
+    #[cfg(all(unix, feature = "std"))]
+    #[test]
+    fn check_timestamp_negative_seconds_1ns() {
+        // UNIX-only test cases with 1 ns precision
         assert_eq!(
             Timestamp::from(UNIX_EPOCH - Duration::new(0, 999_999_999)),
             Timestamp {
@@ -197,17 +229,17 @@ mod tests {
             }
         );
         assert_eq!(
-            Timestamp::from(UNIX_EPOCH - Duration::new(2_001_234, 123)),
+            Timestamp::from(UNIX_EPOCH - Duration::new(1_234_567, 123)),
             Timestamp {
-                seconds: -2_001_235,
+                seconds: -1_234_568,
                 nanos: 999_999_877
             }
         );
         assert_eq!(
-            Timestamp::from(UNIX_EPOCH - Duration::new(768, 987_654_321)),
+            Timestamp::from(UNIX_EPOCH - Duration::new(890, 987_654_321)),
             Timestamp {
-                seconds: -769,
-                nanos: 12345679
+                seconds: -891,
+                nanos: 12_345_679
             }
         );
     }
```
</details>

joseph-wakeling-frequenz:
I'm seeing a bunch of annotations in the CI along the following lines:

```
no global memory allocator found but one is required; link to std or add `#[global_allocator]` to a static item that implements the GlobalAlloc trait.
```

... which look like errors but don't seem to be causing CI failure.  Is that expected behaviour or has it been inadvertently introduced by this patch?

joseph-wakeling-frequenz:
> I'm seeing a bunch of annotations in the CI along the following lines

OK, having checked out a few other PRs, looks like it's nothing specific to this one.  So this patch should be ready for review.

joseph-wakeling-frequenz:
@danburkert is there anything that needs updating or clarifying here?  I think this patch is good to go.

joseph-wakeling-frequenz:
... ping?  Any chance of either feedback or a merge?

danburkert:
Great, thanks!  Sorry for slow response.

joseph-wakeling-frequenz:
No worries, busy times happen to us all :-)  Thanks for the merge!

PR #123, opened 2018-08-12
prost-build: Support generating single include file

kinnison:
In order to support use-cases where many protocol buffer
packages are parsed and generated, support the configuration
containing an optional `include_file()` which allows the build
script to name a file to be created to automatically `include!()`
all the generated content with the right module structure.

Signed-off-by: Daniel Silverstone <dsilvers@digital-scurf.org>

PR #333, opened 2020-05-17
prost-build: Support generating single include file

dfreese:
In order to support use-cases where many protocol buffer
packages are parsed and generated, support the configuration
containing an optional `include_file()` which allows the build
script to name a file to be created to automatically `include!()`
all the generated content with the right module structure.

Signed-off-by: Daniel Silverstone <dsilvers@digital-scurf.org>

### DIFF
--- a/prost-build/src/lib.rs
+++ b/prost-build/src/lib.rs
@@ -117,7 +117,7 @@ use std::default;
 use std::env;
 use std::fmt;
 use std::fs;
-use std::io::{Error, ErrorKind, Result};
+use std::io::{Error, ErrorKind, Result, Write};
 use std::path::{Path, PathBuf};
 use std::process::Command;
 
@@ -190,6 +190,7 @@ pub struct Config {
     strip_enum_prefix: bool,
     out_dir: Option<PathBuf>,
     extern_paths: Vec<(String, String)>,
+    include_file: Option<PathBuf>,
 }
 
 impl Config {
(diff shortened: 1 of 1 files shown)


dfreese:
This is just a rebase of #123

giacomocariello:
Bump

jeffparsons:
This is highly desirable behaviour for me. We have many hundreds of interdependent `.proto` files, so including them all is the only way to use any of them.

@danburkert Are there things about the existing attempts (this PR, https://github.com/danburkert/prost/pull/294, etc.) that make you unsure about whether you'd want to merge them? Equivalently, are you able to offer some guidance on what a PR to address this would look like for you to be confident in merging it?

prost-build/src/lib.rs
```diff
@@ -578,9 +616,83 @@ impl Config {
...
+                basepath,
+            )?;
+            written += subwritten;
+            if subwritten != matching.len() {
+                let modname = matching[0][..=depth].join(".");
+                if let Some(buf) = basepath {
+                    self.write_line(
+                        outfile,
+                        depth + 1,
+                        &format!("include!(\"{:?}/{}.rs\");", buf, modname),
```
samschlegel, line 675: Using debug here is 👎 as it leads to bad quoting
```rust
include!(""src/"/google.protobuf.compiler.rs");
```

dfreese:
Closing in favor of #433, which fixes some issues here

PR #336, opened 2020-05-25
Include Message Proto information

fdeantoni:
A new trait `MessageProto` which includes the following info about a message:
 - `package_name` which is taken from the `package` field in the proto file;
 - `message_name` which is the message name itself;
 - `type_url` which is "type.googleapis.com/" + package_name + "." + message_name;

 The prefix "type.googleapis.com" is the default prefix for all protobuf types (see [here](https://developers.google.com/protocol-buffers/docs/reference/java/com/google/protobuf/Any)).

 Adding the MessageProto trait will help with the packing and unpacking of messages in the `Any` type, but may be helpful in other situations as well, and can easily be extended to include more proto information in future (e.g. source file etc).

This pull request can address #277 completely as it will allow an external crate to provide all the functionalities without "polluting" the `prost` crate with additional dependencies. For an external crate that provides all the functionality described in #277, see [prost-wkt](https://github.com/fdeantoni/prost-wkt).

### DIFF
--- a/prost-derive/src/lib.rs
+++ b/prost-derive/src/lib.rs
@@ -12,11 +12,11 @@ use proc_macro2::Span;
 use quote::quote;
 use syn::{
     punctuated::Punctuated, Data, DataEnum, DataStruct, DeriveInput, Expr, Fields, FieldsNamed,
-    FieldsUnnamed, Ident, Variant,
+    FieldsUnnamed, Ident, Variant, Meta,
 };
 
 mod field;
-use crate::field::Field;
+use crate::field::{Field, prost_attrs};
 
 fn try_message(input: TokenStream) -> Result<TokenStream, Error> {
     let input: DeriveInput = syn::parse(input)?;
@@ -28,6 +28,24 @@ fn try_message(input: TokenStream) -> Result<TokenStream, Error> {
         Data::Enum(..) => bail!("Message can not be derived for an enum"),
         Data::Union(..) => bail!("Message can not be derived for a union"),
     };
+
+    let pkg_name = prost_attrs(input.attrs.clone()).unwrap().iter().find( |meta| {
+        meta.path().is_ident("package")
+    }).and_then( |meta| {
+        match meta {
+            Meta::NameValue(v) => match &v.lit {
+                syn::Lit::Str(lit) => Some(lit.value().clone()),
+                _ => None
+            },
+            _ => None
+        }
+    }).unwrap_or_else( || String::from("prost"));
+
+    let type_url = format!(
+        "type.googleapis.com/{}.{}",
+        pkg_name,
+        ident
+    );
 
     if !input.generics.params.is_empty() || input.generics.where_clause.is_some() {
         bail!("Message may not be derived for generic type");
@@ -212,6 +230,18 @@ fn try_message(input: TokenStream) -> Result<TokenStream, Error> {
                 #ident {
                     #(#default)*
                 }
+            }
+        }
+
+        impl ::prost::MessageProto for #ident {
+            fn message_name(&self) -> &'static str {
+                stringify!(#ident)
+            }
+            fn package_name(&self) -> &'static str {
+                #pkg_name
+            }
+            fn type_url(&self) -> &'static str {
+                #type_url
             }
         }
 
--- a/src/message.rs
+++ b/src/message.rs
@@ -165,3 +165,9 @@ where
         (**self).clear()
     }
 }
+
+pub trait MessageProto: Message + 'static {
+    fn message_name(&self) -> &'static str;
+    fn package_name(&self) -> &'static str;
+    fn type_url(&self) -> &'static str;
+}
--- a/prost-build/src/code_generator.rs
+++ b/prost-build/src/code_generator.rs
@@ -117,6 +117,7 @@ impl<'a> CodeGenerator<'a> {
         debug!("  message: {:?}", message.name());
 
         let message_name = message.name().to_string();
+        let package_name = self.package.clone();
         let fq_message_name = format!(".{}.{}", self.package, message.name());
 
         // Skip external types.
@@ -175,6 +176,8 @@ impl<'a> CodeGenerator<'a> {
         self.push_indent();
         self.buf
             .push_str("#[derive(Clone, PartialEq, ::prost::Message)]\n");
+        self.push_indent();
+        self.buf.push_str(format!("#[prost(package=\"{}\")]\n", package_name).as_str());
         self.push_indent();
         self.buf.push_str("pub struct ");
         self.buf.push_str(&to_upper_camel(&message_name));
--- a/prost-types/src/compiler.rs
+++ b/prost-types/src/compiler.rs
@@ -1,5 +1,6 @@
 /// The version number of protocol compiler.
 #[derive(Clone, PartialEq, ::prost::Message)]
+#[prost(package="google.protobuf.compiler")]
 pub struct Version {
     #[prost(int32, optional, tag="1")]
     pub major: ::core::option::Option<i32>,
@@ -14,6 +15,7 @@ pub struct Version {
 }
 /// An encoded CodeGeneratorRequest is written to the plugin's stdin.
 #[derive(Clone, PartialEq, ::prost::Message)]
+#[prost(package="google.protobuf.compiler")]
 pub struct CodeGeneratorRequest {
     /// The .proto files that were explicitly listed on the command-line.  The
     /// code generator should generate code only for these files.  Each file's
@@ -45,6 +47,7 @@ pub struct CodeGeneratorRequest {
 }
 /// The plugin writes an encoded CodeGeneratorResponse to stdout.
 #[derive(Clone, PartialEq, ::prost::Message)]
+#[prost(package="google.protobuf.compiler")]
 pub struct CodeGeneratorResponse {
     /// Error message.  If non-empty, code generation failed.  The plugin process
     /// should exit with status code zero even if it reports an error in this way.
@@ -63,6 +66,7 @@ pub struct CodeGeneratorResponse {
 pub mod code_generator_response {
     /// Represents a single generated file.
     #[derive(Clone, PartialEq, ::prost::Message)]
+    #[prost(package="google.protobuf.compiler.CodeGeneratorResponse")]
     pub struct File {
         /// The file name, relative to the output directory.  The name must not
         /// contain "." or ".." components and must be relative, not be absolute (so,
--- a/prost-types/src/protobuf.rs
+++ b/prost-types/src/protobuf.rs
@@ -1,12 +1,14 @@
 /// The protocol compiler can output a FileDescriptorSet containing the .proto
 /// files it parses.
 #[derive(Clone, PartialEq, ::prost::Message)]
+#[prost(package="google.protobuf")]
 pub struct FileDescriptorSet {
     #[prost(message, repeated, tag="1")]
     pub file: ::prost::alloc::vec::Vec<FileDescriptorProto>,
 }
 /// Describes a complete .proto file.
 #[derive(Clone, PartialEq, ::prost::Message)]
+#[prost(package="google.protobuf")]
 pub struct FileDescriptorProto {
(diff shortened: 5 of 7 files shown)


aljoscha:
Is there any update on this PR? I recently stumbled across this because I need to work with `Any`. I'd really like to use Prost because I think it feels much more idiomatic but right now I'm forced to use `protobuf`.

bfrog:
I'd really like to see this happen, along with perhaps some attribute describing the .proto real field name rather than the more friendly rust one, such as #[prost(package=..., message=..., field=...)] to effectively describe in metadata than prost-derive could then use, if desired, to generate the reflection API

fdeantoni:
Hi @danburkert , as you suggested in your [comment on issue 255](https://github.com/danburkert/prost/pull/255#issuecomment-575929724), I broke up the requirement into smaller parts. This PR now only implements a single trait with three fields (`message_name`, `package_name`, and `type_url`) for each Message struct, and generates those structs from protobuf messages with a new `package` macro attribute (to provide the `package_name` field). It does not pull in any other dependencies, and should be fully backwards compatible with previous releases. Do let me know if this is ok or if you think more work is needed. I really like this library and hope this PR can be included!

fdeantoni:
Hi @danburkert, I was wondering if you have plans of adding some kind of message descriptor to Prost in future? Although this PR implements it, I hope any implementation that allows me to retrieve a message `type_url` can make into the project. Really love Prost and it is the only thing I'm missing from this great library!

jcaden (approved):
These changes look good to me. I was wondering if they could be merged as they allow for serializing/Deserializing into/from Any type that is the only missing functionality for us in prost.

haozibi:
WANT MERGE!!!

jcaden:
Whie this is merged, I've managed to do a workaround without forking prost nor prost build. Basically I ask prost-build to left the files in a known directory and then get the path from the file name, using that information a rewrite the generated code adding package information. I also add an attribute to the generated types in order to include some methods. It's not the best solution ever, but at least I can carry on without forking.

The buidl is changed by this function:

```
/// Generates source from protocol buffer definitions and adds it information about the package
/// That can be used by the proto_package_name macro to generate media information
pub fn generate_interface_source<P>(
    out_file: P,
    protos: &[P],
    includes: &[P],
) -> Result<(), Box<dyn std::error::Error>>
where
    P: AsRef<Path>,
{
    let mut out_dir: PathBuf = env::var_os("OUT_DIR")
        .ok_or_else(|| Error::new(ErrorKind::Other, "OUT_DIR environment variable is not set"))
        .map(Into::into)
        .unwrap();

    out_dir.push("generated");

    // Ignoring create_dir error as it can already exist
    let _ = fs::create_dir(&out_dir);

    prost_build::Config::new()
        .out_dir(&out_dir)
        .type_attribute(".", "#[proto_package_name(PACKAGE_NAME)]")
        .compile_protos(protos, includes)?;

    fs::remove_file(&out_file)?;
    let mut generated_output = OpenOptions::new()
        .read(true)
        .write(true)
        .create(true)
        .open(&out_file)?;

    for entry in fs::read_dir(&out_dir)? {
        let entry = entry?;
        let path = entry.path();
        if !path.is_dir() {
            let mut content = String::new();
            File::open(&path).unwrap().read_to_string(&mut content)?;
            let file_name = path.file_stem().unwrap().to_str().unwrap();

            let tokens = file_name.split(".");
            let mut tokens_len = 0;
            for module in tokens {
                writeln!(generated_output, "pub mod {} {{", module)?;
                tokens_len = tokens_len + 1;
            }

            writeln!(
                generated_output,
                "\nuse proto_package_name::proto_package_name;\n"
            )?;

            writeln!(
                generated_output,
                "static PACKAGE_NAME : &str = \"{}\";\n",
                file_name
            )?;

            generated_output.write_all(content.as_bytes())?;

            for _ in 0..tokens_len {
                writeln!(generated_output, "}}")?;
            }
        }
    }

    generated_output.sync_all()?;

    fs::remove_dir_all(&out_dir)?;

    Ok(())
}
```

The code of the macro looks like this:

```
#[derive(Debug)]
struct ProtocPackageNameParams {
    module_name_constant: Ident,
}

impl Parse for ProtocPackageNameParams {
    fn parse(input: ParseStream) -> Result<Self> {
        Ok(ProtocPackageNameParams {
            module_name_constant: input.parse()?,
        })
    }
}

#[proc_macro_attribute]
pub fn proto_package_name(attr: TokenStream, item: TokenStream) -> TokenStream {
    let item = parse_macro_input!(item as DeriveInput);
    let struct_name = &item.ident;
    let struct_name_string = quote!(#struct_name).to_string();
    let params = parse_macro_input!(attr as ProtocPackageNameParams);
    let module_name_constant = &params.module_name_constant;

    let item = TokenStream::from(quote!(
        #item impl #struct_name {
            pub fn package_name() -> &'static str{
                &#module_name_constant
            }

            pub fn message_name() -> &'static str {
                #struct_name_string
            }

            pub fn type_url() -> String {
                format!(
                    "type.googleapis.com/{}.{}",
                    Self::package_name(),
                    Self::message_name()
                )
            }
        }
    ));

    item
}
```

If there is interest I can create a project to implement this workaround.

Khalilw1:
Hello guys any ETA on this. I really want to use Any without moving to another library (I really like Prost generation)

mkatychev:
@jcaden that looks really interesting, do you mind providing a minimal example of the two snippets in action?

jcaden:
> @jcaden that looks really interesting, do you mind providing a minimal example of the two snippets in action?

Actually I found some cases where this is not working as expected when having multiple packages there were some name collisions, not sure if that's the best solution.

bfrog:
So my start of this solutions is to effectively include and create a static with the message descriptors

https://github.com/openenergysolutions/openfmb-rs/blob/develop/openfmb-descriptors/src/lib.rs#L9

I then codegen from that crate and have all the metadata available at build and runtime that way.

fdeantoni:
That could work too but I think it is still cleaner to have the descriptor as part of the message via a trait. To me adding a  `MessageDescriptor` trait seems like a relatively safe feature to add, or am I overlooking something important? Really hope the maintainers of Prost can help out and give their feedback...

bfrog:
@fdeantoni I'd say implementing the metadata/descriptor bits as functions that pull out data from the static variable is the sanest route, that way you don't end up with circular references which Rust won't like.

The static is just step one, step two might be adding a trait and trait impls being created for each generated message type to get that descriptor data from the static.

fdeantoni:
Thanks for the feedback @bfrog! I'm definitely not against using a static to get the package information, and using a trait impl for each message to get that data. All I want really is just a way to reconstruct the package url from the message :) . However, I figured just placing this data in the trait would be fastest (and safest). In essence the only thing this PR really does is add the following trait:
```rust
pub trait MessageDescriptor: Message + 'static {
    fn message_name(&self) -> &'static str;
    fn package_name(&self) -> &'static str;
    fn type_url(&self) -> &'static str;
}
```

The derive macro then takes care of implementing this trait for each message, for example:
```rust
    impl ::prost::MessageDescriptor for Foo {
        fn message_name(&self) -> &'static str {
            "Foo"
        }
        fn package_name(&self) -> &'static str {
            "com.example.package"
        }
        fn type_url(&self) -> &'static str {
            "type.googleapis.com/com.example.package.Foo"
        }
    }
```

You mention circular references as a potential risk to this implementation though. Can you tell me more how/where this might happen? If there is a weakness in the design then definitely it should be addressed!

bfrog:
@fdeantoni as soon as you start adding field descriptors which themselves may be messages, there's maybe this desire to do the reverse of determining the types that may contain another type as a field. My suggestion is to avoid all that insanity and basically inject the FileDescriptorSet as a static, and reference all things from it. You can easily expand the metadata API that way.

For example the protobuf api itself provides this, https://developers.google.com/protocol-buffers/docs/reference/cpp/google.protobuf.message#Message.GetDescriptor.details

You could copy all the metadata/descriptor bits about a protobuf message type, or you could build that API on top of the static, both are probably reasonable. Note that the Descriptor type there provides a "containing_type", "file", and field iteration API.

A FieldDescriptor also points to a FileDescriptor, and maybe more interesting the "containing_type" call provides the parent message Descriptor.

So you have a circular reference in Descriptor <-> FieldDescriptor reference and know of eachother.

You can avoid that by making everything point to the static ref perhaps.

fdeantoni:
Ok, but I think that is a much bigger change than what I'm proposing for this PR, which is really a very simple and dumb message descriptor trait containing only the package name and message name (to construct a url). The next step indeed could be adding a static reference to get the field descriptors, but that I think should be a separate PR. 

Nonetheless, if the maintainers are willing to go with a much bigger change that implements the full Descriptor API comparable to the cpp version then that's even better, but I was hoping by keeping this PR small and simple (with potential to add features incrementally) it had more chance of being included. Hopefully the maintainers can weigh in here and suggest the best course of action.

neverchanje:
Let's take a look on how C++ implements `Any::PackFrom`:

src/message.rs
```diff
@@ -166,6 +166,12 @@ where
     }
 }
 
+pub trait MessageDescriptor: Message + 'static {
+    fn message_name(&self) -> &'static str;
+    fn package_name(&self) -> &'static str;
+    fn type_url(&self) -> &'static str;
```
neverchanje, line 172: We may not need `type_url` to be included in the trait. Let's take a look on how C++ implements `Any::PackFrom`:

```cpp
bool AnyMetadata::PackFrom(Arena* arena,
                           const Message& message,
                           StringPiece type_url_prefix) {
  type_url_->Set(
      &::PROTOBUF_NAMESPACE_ID::internal::GetEmptyString(),
      GetTypeUrl(message.GetDescriptor()->full_name(), type_url_prefix),
      arena);
  return message.SerializeToString(
      value_->Mutable(ArenaStringPtr::EmptyDefault{}, arena));
}
```

It uses an API called `Descriptor::full_name` which returns "<package>.<message>", and finally generates the type_url via `GetTypeUrl`, which is hard-coded in the library and can not be customized like a trait.

fdeantoni: The `type_url` was just added for convenience really. If this stops the PR from being merged then I'm all for removing it! A third-party library that implements pack and unpack like [prost-wkt](https://github.com/fdeantoni/prost-wkt) can then implement it instead (with the prefix configurable and whatnot).

neverchanje: I really appreciate your work on prost-wkt and I'm actually not one of the prost maintainers, so I couldn't and didn't intend to stop anybody. I left this comment only because I feel a minimum public trait is better and will not lead to incompatibility issues in the future. Anyway, I'm really looking forward to this PR being merged, then we will not have to use a forked version of prost in prost-wkt any longer.

prost-derive/src/lib.rs
```diff
@@ -206,6 +221,18 @@ fn try_message(input: TokenStream) -> Result<TokenStream, Error> {
...
+            fn message_name(&self) -> &'static str {
+                stringify!(#ident)
+            }
+            fn package_name(&self) -> &'static str {
+                #pkg_name
+            }
+            fn type_url(&self) -> &'static str {
+                #type_url
+            }
+        }
```
oblique, line 234: I believe the above methods should be associated functions instead. Someone will be able to use `Foo::type_url()` instead of constructing `Foo` and then calling the method.

For example this is how pack/unpack to `Any` will be implemented with associated functions:

```rust
fn pack_to_any<M>(msg: M) -> prost_types::Any
where
    M: prost::Message + prost::MessageDescriptor,
{
    prost_types::Any {
        type_url: M::type_url().to_owned(),
        value: msg.encode_to_vec(),
    }
}

fn unpack_from_any<M>(msg: prost_types::Any) -> Option<M>
where
    M: prost::Message + prost::MessageDescriptor + Default,
{
    if msg.type_url == M::type_url() {
        Some(M::decode(&msg.value[..]).ok()?)
    } else {
        None
    }
}
```

oblique:
I created a workaround example for packing/unpacking to Any that uses FileDescriptorSet in build.rs: https://github.com/oblique/prost-any-workaround

PR #341, opened 2020-06-14
Optional support for bytes::Bytes type backing bytes field

danburkert:
I made a hash of #337; this is that PR with a few extra commits of my own.

### DIFF
--- a/prost-build/src/code_generator.rs
+++ b/prost-build/src/code_generator.rs
@@ -299,7 +299,7 @@ impl<'a> CodeGenerator<'a> {
 
         self.push_indent();
         self.buf.push_str("#[prost(");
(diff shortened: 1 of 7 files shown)


rolftimmermans:
👍 Looks good!

akiradeveloper:
@danburkert I am looking for an efficient serialization using `bytes::Bytes` and found this thread. But a slight look at the code finds me that encoding is still copying the buffer entirely but since you use Bytes that is internally just a pair of pointer and len, you should want to avoid copy here as commented.

But I wonder, how do you do this? Do you have some idea to solve this problem?

```rust
 buf.put_slice(value.as_slice()); // TODO Avoid copying
```

PR #345, opened 2020-07-16
Remove testing of Rust 1.39

Xaeroxe:
Master branch doesn't build on Rust 1.39, incoming PRs shouldn't be tested against it. (It did build at one point but if you delete your Cargo.lock you'll see it doesn't anymore)

### DIFF
--- a/.github/workflows/continuous-integration-workflow.yaml
+++ b/.github/workflows/continuous-integration-workflow.yaml
@@ -46,7 +46,6 @@ jobs:
       matrix:
         toolchain:
           - stable
-          - 1.39.0
         os:
           - ubuntu-latest
           - macos-latest


danburkert:
Yes `prost` depends on a crate which doesn't respect an MSRV.  The fix is not for `prost` to stop respecting an MSRV, though.

Xaeroxe:
Your master branch doesn't pass CI. If you're just going to close my PR you should at least detail what your plans are to get CI functional for your contributors. The MSRV isn't documented anywhere, it seems to only exist in this CI.

danburkert:
> you should at least detail what your plans are to get CI functional for your contributors

This is the fix: https://github.com/danburkert/prost/pull/347

> The MSRV isn't documented anywhere, it seems to only exist in this CI.

Yes, that's where it's documented and enforced.

PR #352, opened 2020-07-25
Add a reference to snazzy near the top of readme.

### LINKED ISSUE #346: Documentation needs a "Hello World" example
I'm struggling to find a simple, copy-pastable example of simply serializing and serializing a proto.

hjfreyer:
Resolves #346


nashley (approved):
I think this would make getting started much easier 👍

LucioFranco:
I think it would be better to embed an example crate within the main prost repo and workspace for this rather than link out?

LucioFranco (approved):

PR #360, opened 2020-08-27
Fix code generation for code blocks in prost-build

Xaeroxe:
If the generated doc comments include a code block rustdoc will erroneously try and execute them.

That's silly as these blocks almost never contain valid Rust, so this PR marks these blocks as text.


Xaeroxe:
@danburkert I've been trying to get these commits merged to this repo for over 6 months now. If you don't like the PR can we at least talk about why?

aheadmode:
@danburkert

djc:
(Looks like there's now an alternative proposal on the table in #389.)

PR #361, opened 2020-08-29
prost-build: (docs) adds NixOS related hints

amanjeev:
On NixOS, it is better to explicitly define Protobuf related
environment variables `PROTOC` and `PROTOC_INCLUDE`.

### DIFF
--- a/prost-build/src/lib.rs
+++ b/prost-build/src/lib.rs
@@ -112,8 +112,8 @@
 //!
 //! ```bash
 //! $ export PROTOBUF_LOCATION=$(nix-env -q protobuf --out-path --no-name)
-//! $ export PROTOC=$(PROTOBUF_LOCATION)/bin/protoc
-//! $ export PROTOC_INCLUDE=$(PROTOBUF_LOCATION)/include
+//! $ export PROTOC=$PROTOBUF_LOCATION/bin/protoc
+//! $ export PROTOC_INCLUDE=$PROTOBUF_LOCATION/include
 //! $ cargo build
 //! ```
 //!
--- a/prost-build/src/lib.rs
+++ b/prost-build/src/lib.rs
@@ -113,7 +113,7 @@
 //! ```bash
 //! $ export PROTOBUF_LOCATION=$(nix-env -q protobuf --out-path --no-name)
 //! $ export PROTOC=$(PROTOBUF_LOCATION)/bin/protoc
-//! $ export PROTOC_INCLUDE=$($PROTOBUF_LOCATION)/include
+//! $ export PROTOC_INCLUDE=$(PROTOBUF_LOCATION)/include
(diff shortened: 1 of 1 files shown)


prost-build/src/lib.rs
```diff
@@ -105,6 +105,20 @@
 //!
 //! If `PROTOC_INCLUDE` is not found in the environment, then the Protobuf include directory bundled
 //! in the prost-build crate is be used.
+//!
+//! ### NixOS related hints
+//!
+//! On NixOS, it is better to specify the location of `PROTOC` and `PROTOC_INCLUDE` explicitly.
+//!
+//! ```bash
+//! $ export PROTOC=/nix/store/<SHA>-protobuf/bin/protoc
```
Hoverbear, line 114: I noted that these lines were useful in a `.nix` related build file as well:

```
        PROTOC = "${pkgs.protobuf}/bin/protoc";
        PROTOC_INCLUDE = "${pkgs.protobuf}/include";
```

amanjeev: Oh, thank you! :slightly_smiling_face:  
Should I add those instead? in the docs as well or instead?

Hoverbear: Is there a `nix` command users can use to get the path of the protoc package in the store, so they don't need to discover the hash?

amanjeev: Oh I see! I do not know about that. I should check for `nix-env` maybe? But in my `shell.nix` I do have 

```
{ pkgs ? import <nixpkgs> {} }:
with pkgs;
mkShell {
  buildInputs = [
    python38Full
    stdenv
    stdenv.cc.cc.lib
    python38Packages.tox
    protobuf
  ];
  shellHook = ''
    export PYTHONPATH=`pwd`/$VENV/${python.sitePackages}:$PYTHONPATH;
    export LD_LIBRARY_PATH="${stdenv.cc.cc.lib}/lib64:$LD_LIBRARY_PATH";
    export PROTOC="${pkgs.protobuf}/bin/protoc";
    export PROTOC_INCLUDE="${pkgs.protobuf}/include";
  '';
}
```

amanjeev: Ah! I think this is it 

```
$ nix-env -q protobuf --out-path

protobuf-3.7.1  /nix/store/5kl6rzlxa6g86ks9ldl9hm69fd6fl963-protobuf-3.7.1
```

danburkert:
I beefed up the docs and failure modes when protoc is not executable as part of #410 which should make the error more self evident.  I'm not going to accept docs specific to nixos, since it's an extremely niche system.  Additionally, the specific verbiage included in this PR ('... the failure is not obvious until ...') is not obviously correct to me, and I have no way to test the assertion.

PR #367, opened 2020-09-10
WIP: make deriving Debug optional

TheVova:
addresses #334 by allowing users to opt-out of having prost derive Debug for its structs.
1) `prost_build::Config` gains a `no_derive_dbg(bool)` method for users to use. this inserts a `#[prost(no_dbg)]` attribute on the generated struct.
2)  prost-derive checks for that attribute and skips deriving Debug if it is present.
3) The bound on `Message` change from `Debug + Send + Sync` to just `Send + Sync`.
The default behavior is opt-out, so it breaks nothing for current users. 
WIP because still need to adjust docs, but wanted to see if this is acceptable.


TheVova:
Ok, so this breaks some tests atm because of the hidden assumption that Message implies Debug, which probably means it will break some user code that relies on this property. One solution would be to restore the Debug bound to Message and just rely that every type is gonna have either prost-debug, `#[derive(Debug)]` or a manual `impl Debug` done for it. 
alternatively, could also make this PR more granular and apply to single messages rather than all...

garbageslam:
fwiw:

if this were coupled with a version bump to 0.8 it seems not unreasonable to tell users, there is a breaking change now deriving message does not also derive debug. it is very easy for users to adapt to the change, they just derive Debug as well as Message if they need it. ofc that's just my 2 cents

AlJohri:
related: https://github.com/tokio-rs/prost/pull/493 (this PR also makes debug configurable)

caspermeijn:
I think this has been fixed in main branch in the mean time.

- If your code is generated by `prost_build`, then you can configure that to not generate a `impl Debug` with https://docs.rs/prost-build/latest/prost_build/struct.Config.html#method.skip_debug
- If your code uses `prost_derive`, then the `skip_debug` attribute exists: https://docs.rs/prost-derive/0.13.3/src/prost_derive/lib.rs.html#25
- Currently an `impl Debug` is required for each `prost::Message`, but that requirement is relaxed in this PR: https://github.com/tokio-rs/prost/pull/1147

If this comment helps you, please consider updating the crate documentation to make these features more findable.

PR #368, opened 2020-09-14
Do not force an unnecessary Sized bound

nagisa:
This allows users to call `Message` methods on `&(dyn Message)`.

### DIFF
--- a/src/encoding.rs
+++ b/src/encoding.rs
@@ -369,6 +369,7 @@ pub fn merge_loop<T, M, B>(
 where
     M: FnMut(&mut T, &mut B, DecodeContext) -> Result<(), DecodeError>,
     B: Buf,
+    T: ?Sized,
 {
     let len = decode_varint(buf)?;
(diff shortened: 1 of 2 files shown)


danburkert:
Wow, I didn't realize this was possible.  Thanks!

danburkert:
@nagisa it looks like even with this change, `Message` is not object safe.  As such, what's the advantage of dropping the trait bounds?

danburkert:
err, I should say this change makes `Message` non object safe.

nagisa:
Ah, well. I didn't look into this avenue after submitting this PR any further and went on to do something entirely different.

I guess there isn't much actual value in this PR in that case, entirely up to you to decide what you want to do with it.

danburkert:
Yep, I'm going to close this out accordingly.  If there is a clever way to to do something like this but still retain object safety I'm all for it.  In the meantime I landed #385 which adds a regression test that Message is object safe.

PR #377, opened 2020-10-26
Use different type parameters for compile_protos arguments

nikclayton-dfinity:
Both type parameters are still AsRef<Path>, but this allows for the first
and second arguments to be different types, as long as they both implement
AsRef<Path>.

Previously, they had to be the same type. So you couldn't, for example,
construct the protos arg as a Vec<String> and the includes arg as a
[&str].


nikclayton-dfinity:
The CI failures do not appear to be related

caspermeijn:
This seems to be fixed on latest master

PR #378, opened 2020-10-26
Encode to Vec<u8>

### LINKED ISSUE #154: Add an into vec implementation?
Hi there. Like Prost a lot. I find that by far the most common case for me when encoding is

1. Make a new empty vector
2. Immediately encode into it.

I absolutely understand why the encode function works the way it does, but given how common the case I just mentioned is, seems worth making this even easier on the caller. Namely:

```rust
let encoded: Vec<u8> = message.into()
```

This would be super. If you like the idea, I'm happy to contribute a PR (admittedly, not a whole lot to do here :smile: )

Thanks!

rubdos:
Adds `Message::encode_to_vec(&self) -> Vec<u8>` and `Message::encode_length_delimited_to_vec() -> Vec<u8>`  on `std`-enabled configurations.

This is one line shorter than writing

```rust
let mut buf = Vec::new();
msg.encode(&mut buf).unwrap();
```
and replaces it by
```rust
let buf = msg.encode_to_vec();
```

It also omits the extra capacity checking of `encode` (since `Vec` is pre-allocated because this patch uses `Vec::with_capacity()`).

Fixes #154 in spirit, if you agree that `Into<Vec<u8>> for Message` is a slightly worse idea (i.e., `.into` doesn't specify whether to have length encoding or not, so it might yield slightly more unreadable code).


rubdos:
Rebased on master, subtle *nod nod* to the owners ;-)

Jasperav:
I guess this fixes https://github.com/danburkert/prost/issues/461, what do you think @rubdos?

rubdos:
I agree that it'd fix #461 if it's added to the README. I'd have hoped to have some feedback from the maintainers though by now :/

Jasperav:
@danburkert please review

rubdos:
This is a relatively simple QoL patch; would any of the people that have commit rights please take a look?

LucioFranco (approved):
sorry for the delay on this!

Issue #279, opened 2020-02-14
Investigate potential x86 varint optimization


danburkert:
https://www.reddit.com/r/rust/comments/f36j05/comment/fhhwqp9

danburkert:
see also https://github.com/gnzlbg/bitintr for safe and cross platform wrappers over the intrinsics

danburkert:
https://news.ycombinator.com/item?id=25183811

danburkert:
https://www.reddit.com/r/rust/comments/klck6a/i_published_my_first_crate_varintsimd/

as-com:
So I did some quick and dirty prototyping with varint-simd v0.3.0, and here's what I found:

*   Microbenchmark varint performance is only improved for encoding and decoding larger numbers
*   Encoding performance generally fares better than decoding performance, depending on where I place the branch that uses varint-simd
*   Macrobenchmark performance is mostly a wash (tested on Coffee Lake), with some larger wins and some smaller losses

This is probably because the only encode/decode function is for single u64's, which is currently a weak point for varint-simd (it's not that much faster than other implementations when decoding/encoding tiny u64's).

I suspect there will need to be some larger-scale refactoring to take full advantage of varint-simd. For example, protobuf tags are up to 32 bits long, so a lot of cycles can be saved when encoding/decoding those. 

My library also just added support for quickly decoding two, four, and eight adjacent varints in parallel (subject to size limitations), with some really good throughput figures - most of the time, protobufs will be a 32 bit tag followed by a 32 bit number or length, and decode requests can be shrunk based on how large the data field is in the .proto file. So there's likely a lot more gains to be had.

PR #381, opened 2020-11-07
Update bytes to 0.6

olix0r:
This change updates the `bytes` crate dependencies to version 0.6.

Fixes https://github.com/danburkert/prost/issues/380

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,7 +42,7 @@ no-recursion-limit = []
 std = []
 
 [dependencies]
-bytes = { version = "0.5", default-features = false }
+bytes = { version = "0.6", default-features = false }
 prost-derive = { version = "0.6.1", path = "prost-derive", optional = true }
 
 [dev-dependencies]
--- a/conformance/Cargo.toml
+++ b/conformance/Cargo.toml
@@ -6,7 +6,7 @@ publish = false
 edition = "2018"
 
 [dependencies]
-bytes = "0.5"
+bytes = "0.6"
 env_logger = { version = "0.7", default-features = false }
 log = "0.4"
 prost = { path = ".." }
--- a/prost-build/Cargo.toml
+++ b/prost-build/Cargo.toml
@@ -10,7 +10,7 @@ description = "A Protocol Buffers implementation for the Rust Language."
 edition = "2018"
 
 [dependencies]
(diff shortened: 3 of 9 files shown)


olix0r:
I don't have much context on the prior implementations to be confident that the ported version is absolutely correct, but at least the tests all pass locally.

olix0r:
This CI failure appears to be unrelated to my change?
```
error[E0658]: the `#[non_exhaustive]` attribute is an experimental feature
224
Error:    --> /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/curl-0.4.34/src/easy/handler.rs:395:1
225
    |
226
395 | #[non_exhaustive]
227
    | ^^^^^^^^^^^^^^^^^
228
    |
229
    = note: for more information, see https://github.com/rust-lang/rust/issues/44109
230
```

wchargin:
CI failure is indeed unrelated; #382 should fix.

danburkert:
Took a look, this unfortunately results in a very significant perf hit during encoding.  I switched `encode_varint` back to something closer to the `master` version:

```rust
    unsafe {
        let mut i;
        'outer: loop {
            i = 0;

            let uninit_slice = buf.bytes_mut();
            let ptr = uninit_slice.as_mut_ptr();
            for offset in 0..uninit_slice.len() {
                i += 1;
                let ptr = ptr.offset(offset as isize);
                if value < 0x80 {
                    ptr.write(value as u8);
                    break 'outer;
                } else {
                    ptr.write(((value & 0x7F) | 0x80) as u8);
                    value >>= 7;
                }
            }

            buf.advance_mut(i);
            debug_assert!(buf.has_remaining_mut());
        }

        buf.advance_mut(i);
    }
```

but that still is ~40-50% slower in the in-tree benchmarks (both the synthetic `varint` and the more real world protobuf benchmarks).  Will keep looking into this as I have time, #279 may be a good way to sidestep the issue.  My early thoughts are that this is related to slowness in the new `UninitSlice` API, perhaps there is an `#[inline]` tag missing, or it just doesn't optimize as well.

danburkert:
It was indeed missing `#[inline]` tags in `bytes`.  Closing in favor of #387.  Thanks @olix0r !

PR #326, opened 2020-05-11
Add `include_file_descriptor_set` to `prost-build`

jen20:
This pull request adds support for writing an encoded version of the appropriate FileDescriptorSet to a file named `file_descriptor_set.bin` in the output directory. This can be used with the `include_bytes` macro by consuming applications or libraries.

The rationale for this pull request is to support the gRPC Server Reflection Protocol, which operates in terms of `FileDescriptorProto`, in Tonic. `FileDescriptorProto` instances are contained inside a
`FileDescriptorSet`.

This supersedes #311 following discussion on that issue.

Issue #43, opened 2017-10-14
Add #[non_exhaustive] option to enum codegen options


danburkert:
It would be nice to be able to have a way to tag generated enums with `#[non_exhaustive]`.  The feature is still [in development](https://github.com/rust-lang/rust/issues/44109), but once it's implemented and unstable, `prost` can in theory take advantage of it without requiring nightly.

per-gron:
Is the plan to just add `#[non_exhaustive]` to the generated enums as they are today and keep the i32 storage, or to only have the enum directly in the generated structs?

It would be nice to use enums directly but the [RFC](https://github.com/rust-lang/rfcs/blob/master/text/2008-non-exhaustive.md) says things that imply that it wouldn't be safe to do so: "Outside the crate that defines the enum, users should be required to add a wildcard arm to ensure forward-compatibility, like so: [match expression with default arm] And it should not be marked as dead code, even if the compiler does mark it as dead and remove it."

vlisivka:
#[non_exhaustive] is stabilized. Can we switch back to enum's now, please?

danburkert:
To be clear, this issue is not about changing the representation of enum message fields.  The current strategy of using `i32`s is here to stay.  This issue is about applying the `#[non_exhaustive]` tag to the generated Rust `enum` types, and only on an opt-in basis via code gen builder arguments.

PR #382, opened 2020-11-07
deps: require `curl < 0.4.34` to fix MSRV

wchargin:
Builds are failing on unrelated pull requests (e.g., #326) because the
`curl` crate has pushed a new version that depends on Rust 1.40, while
our CI tests against 1.39. We could probably bump to require 1.40 as
well, especially since we want to use the same `#[non_exhaustive]`
attribute, per #43. But a minimal change to unblock development and keep
the current minimum stable Rust version is to just pin `curl < 0.4.34`.

Test Plan:
Running `cargo +1.39.0 test --workspace --all-targets` passes on my
Linux machine.

wchargin-branch: deps-curl-pre-0.4.34

### DIFF
--- a/protobuf/Cargo.toml
+++ b/protobuf/Cargo.toml
@@ -13,7 +13,9 @@ prost-types = { path = "../prost-types" }
 [build-dependencies]
 anyhow = "1"
 cfg-if = "0.1"
-curl = "0.4"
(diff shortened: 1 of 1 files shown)


danburkert:
Thanks!

wchargin:
Thank you!

PR #387, opened 2020-11-09
Update bytes to 0.6

danburkert:
This change updates the `bytes` crate dependencies to version 0.6.

Fixes https://github.com/danburkert/prost/issues/380

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,7 +38,7 @@ no-recursion-limit = []
 std = []
 
 [dependencies]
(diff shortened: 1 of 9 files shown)


danburkert:
This is #380 but with `encode_varint` switched back to the faster unsafe version.  When pointing to a `bytes` version which include tokio-rs/bytes#442 and tokio-rs/bytes#443 benchmarks are the same or slightly better than bytes 0.5.

PR #389, opened 2020-11-12
Add prost-build::Config::disable_comments option

danburkert:


prost-build/src/lib.rs
```diff
@@ -410,6 +411,33 @@ impl Config {
...
+    ///
+    /// Occasionally `.proto` files contain code blocks which are not valid Rust. To avoid doctest
+    /// failures, annotate the invalid code blocks with an [`ignore` or `no_run` attribute][1], or
+    /// disable doctests for the crate with a [Cargo.toml entry][2]. If neither of these options
+    /// are possible, then omit comments on generated code during doctest builds:
+    ///
+    /// ```rust,ignore
+    /// let mut config = prost_build::Config::new();
+    /// cfg_if! {
+    ///     if #[cfg(not(doctest))] {
```
danburkert, line 426: I think this should be doctest, not not(doctest)?

danburkert: So this trick of using `#[cfg(doctest)]` doesn't actually do the right thing.  As such, I think it'd be good to be able to disable docs on a per-symbol basic, like we allow with e.g. btree_map

PR #390, opened 2020-11-12
Added Config::field_visibility

daladim:
Some programs may want some fields coming from a Protobuf not to be `pub`, e.g. in case they want to provide their own getters, that add some logic or value-checking from the raw Protobuf value.

This PR provides a `field_visibility` function for this purpose.

Note that i have only properly tested this commit on the 0.6.1 release (it looks like my test program is not yet compatible with the "master" version of prost-build). Please tell me in case you need me to test it more thoroughly.

### DIFF
--- a/prost-build/src/code_generator.rs
+++ b/prost-build/src/code_generator.rs
@@ -19,6 +19,7 @@ use crate::extern_paths::ExternPaths;
 use crate::ident::{match_ident, to_snake, to_upper_camel};
 use crate::message_graph::MessageGraph;
 use crate::Config;
+use crate::Visibility;
 
 #[derive(PartialEq)]
 enum Syntax {
@@ -388,7 +389,11 @@ impl<'a> CodeGenerator<'a> {
         self.buf.push_str("\")]\n");
         self.append_field_attributes(msg_name, field.name());
         self.push_indent();
-        self.buf.push_str("pub ");
+        let pub_modifier = match self.get_field_visibility(msg_name, field.name()) {
+            Visibility::Public => "pub ",
+            Visibility::Private => "",
+        };
+        self.buf.push_str(pub_modifier);
         self.buf.push_str(&to_snake(field.name()));
         self.buf.push_str(": ");
         if repeated {
@@ -452,8 +457,13 @@ impl<'a> CodeGenerator<'a> {
         ));
         self.append_field_attributes(msg_name, field.name());
         self.push_indent();
+        let pub_modifier = match self.get_field_visibility(msg_name, field.name()) {
+            Visibility::Public => "pub",
+            Visibility::Private => "",
+        };
         self.buf.push_str(&format!(
-            "pub {}: {}::{}<{}, {}>,\n",
+            "{} {}: {}::{}<{}, {}>,\n",
+            pub_modifier,
             to_snake(field.name()),
             lib_name,
             rust_ty,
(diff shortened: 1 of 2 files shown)


danburkert:
So is the intention here that you would flip this flag, and then hand implement getters for the fields you do want to be public?  That seems pretty cumbersome.  Typically if your `.proto` isn't considered a public API you shouldn't be exposing `prost`-generated types in a public interface.

daladim:
Well, consider the following use case: a server sends a long message with many fields (all of them being public APIs, and one of them being an hexadecimal color) that must be handled in Rust. There is no way to enforce such a thing as an "24-bit hexadecimal color" in a `.proto`, but I'd like to benefit from Rust's strong-typing to enforce it anyway, so I'd like to have a custom getter for this field.

As you suggest, I could indeed have my own Rust type `MessageSentByServer` that implements (such a getter and) `std::convert::From<StructFromProst>`, but since all the other fields are useful to my app, this `from` will basically copy every field, and fiddle with the color only. This is cumbersome as well. 
Plus, any change in the `.proto` will require editing this `from` function as well to keep them in sync (and since this will not cause compilation error, this requires a CI step to check it). Also, since the `.proto` might be changed by the backend team that does not know Rust and that won't fix my `MessageSentByServer`, this could cause extra cumbersomeness.

Having the ability to make a single field private, and then just add a getter in a `impl StructFromProst` block seems simpler imho, and resolves the two above problems.
Maybe this behaviour could be even improved by having an ability to write, in the `.proto`, that this field should be private in Rust (e.g. by using [protobuf field-level `option`s](https://developers.google.com/protocol-buffers/docs/proto#options), although deciding whether extending protobuf options is righteous is a matter of opinions)

davidli2010:
hi all, is there a plan to merge this PR?
I really need this `field_visibility` feature in my project.

davidli2010:
Currently I use the [visible](https://crates.io/crates/visible) crate to override the visibility of fields. But I hope prost can support `field_visibility` directly.

guilload:
Hello,

I also could use such a feature. Same use case as @daladim:

Proto:
```proto
message Foo {
}

message Bar {
  Foo foo = 1;
  google.protobuf.Timestamp ts = 2;
}
```

Generated code (approximately):
```rust
struct Bar {
  pub foo: Option<Foo>,
  pub ts: Option<prost_types::Timestamp>
}
```

Bar impl:

```rust
impl Bar {
  pub fn new() -> Self { ... }
  
  pub foo(&self) -> &Foo {
      self.foo.as_ref().expect("The field `foo` is required and should always be set.")
  }
  
  pub ts(&self) -> OffsetDateTime {
      self.ts.map(Into::into).expect("The field `ts` is required and should always be set.")
  }
}
```

I'm basically trying to make working with my generated structs easier. The API would be better if I could hide some fields in favor of getters that optionally perform some light transformations.

In reality, my structs have many fields, I'd rather not maintain almost identical structs and a bunch of `From` conversions to achieve the same goal.

LucioFranco:
I think I still stand by this

> Typically if your .proto isn't considered a public API you shouldn't be exposing prost-generated types in a public interface.

Since this feels like it aligns better with prost's design decisions, I am leaning to not add this additional complexity. That said, I don't have time to do this work so if you're interested I would take a proposal via an issue before a PR and discuss there.

PR #394, opened 2020-11-17
Add bundled support for Mac aarch64 through Rosetta

### LINKED ISSUE #393: Bundled Mac aarch64 binary support
Would be great to be able to support bundled protoc builds for native compiles on the new Mac's with aarch64

repi:
Enable usage of bundled `protoc` when building for Mac on Apple Silicon (aarch64) by simply re-using the existing x86_64 OS X binary and have Rosetta seamlessly translate it.

One could later replace this with a natively compiled OS X aarch64 binary, but this was a quick easy solution and Rosetta is really quite efficient so likely is good enough for most.

Fix #393

### DIFF
--- a/prost-build/build.rs
+++ b/prost-build/build.rs
@@ -42,6 +42,7 @@ fn bundled_protoc() -> Option<PathBuf> {
         ("linux", "x86_64") => "protoc-linux-x86_64",
         ("linux", "aarch64") => "protoc-linux-aarch_64",
         ("macos", "x86_64") => "protoc-osx-x86_64",
+        ("macos", "aarch64") => "protoc-osx-x86_64", // will be translated to aarch64 by Rosetta
         ("windows", _) => "protoc-win32.exe",
         _ => return None,
     };


danburkert:
I don't have hardware to test this on, nor am I aware of any way to test through CI - can you confirm it's working on the new hardware?  Other than that PR LGTM.

danburkert:
WRT shipping a native binary we are beholden to upstream protobuf to start publishing that, and it looks like the latest release doesn't include it.  Should be easy to add here when it does get done upstream: https://github.com/danburkert/prost/blob/master/prost-build/third-party/update-bundled-protobuf.sh#L18

repi:
@danburkert yes confirmed that this worked when running on the Apple DTK which unblocked us. but would be good to have a second party confirmation on the real Apple M1 final hardware and software. 

@hrydgard could you on your M1 try building this branch and see if it works on your new M1 Mac if you have the time?

hrydgard:
Works just fine on M1. Tested with `cargo test`.

kali:
I can also confirm it's working on M1.

hrydgard:
Let's just get this trivial change in? It's a good interim solution.

danburkert:
Thanks all.

PR #397, opened 2020-11-28
Update `seanmiddleditch/gha-setup-ninja` to v3

jen20:
This commit updates the use of seanmiddleditch/gha-setup-ninja in the CI pipeline to v3 from v1, which should remove the use of set-path, which was deprecated by GitHub.

This should allow builds to proceed in GitHub Actions.


danburkert:
Thanks!

Issue #75, opened 2018-02-13
JSON Mapping


cretz:
Are there any plans to support the official [proto3 JSON mapping](https://developers.google.com/protocol-buffers/docs/proto3#json) in prost-build? When building https://github.com/cretz/prost-twirp I have found a need for it. Even if it is not supported directly, the field/type attributes in the prost-build config are still difficult to use for this purpose. I suppose I could parse the protobuf myself w/ the other lib here and then create the field/types though it would be nice if prost-build gave me the AST.

adeschamps:
I configure prost to derive `Serialize` in my `build.rs`:

```rust
let mut config = prost_build::Config::new();
config.type_attribute(".", "#[derive(Serialize)]");
config.type_attribute(".", "#[serde(rename_all = \"camelCase\")]");
```

Are there ways in which the proto3 JSON mapping differs from what serde can do?

cretz:
@adeschamps - I haven't tested, but there is at least one obvious problem with that approach. What if your protobuf fields used underscores in the `.proto` file? You would always change them back to camel case which is wrong. There are surely other reasons you'd either need the AST or need to be built in to the code generator which is making decisions a type/field at a time.

adeschamps:
Doesn't protobuf convert field names to camel case for JSON anyway?

cretz:
Ah yes, you appear to be correct. Maybe I will give this a whirl and see if I can find a proto3 JSON conformance test suite to see which edge cases the two type attributes don't work.

trainman419:
I'm interested in doing something similar and I would love to know if this works too!

danburkert:
Hey all, this has been a really good discussion.  I don't have any plans to use or work on proto3 JSON support, but if someone can figure out a way to do it in a straightforward way, ideally by leveraging Serde, then I'm all for it.

@cretz hopefully you've already seen it already, but there is an in-tree conformance test crate which [skips the JSON](https://github.com/danburkert/prost/blob/master/conformance/src/main.rs#L74) tests.  You should be able to hook that up, and add the JSON/Serde data to the [protobuf](https://github.com/danburkert/prost/tree/master/protobuf) crate.

jroesch:
I am considering doing this as something I'm trying to support at work, is anyone else actively working on this? I was planning on building a new derive which will generate a `serde_json` compatible `Serialize` and `Deserialize` implementation.

danburkert:
[...]

PR #415, opened 2020-12-30
Add a 'serde-derive' feature

daladim:
These changes add a way to have `#[derive(serde::Serialize)]` and `#[derive(serde::Deserialize)]` added to types from prost-types.

They may or may not be useful depending on the user needs, and they add a dependency, so they are gated behind a `serde-derive` feature.

This MR will conflict with #414 , but if you agree to merge both of them I can fix the conflicts if you want

### DIFF
--- a/prost-types/Cargo.toml
+++ b/prost-types/Cargo.toml
@@ -15,10 +15,13 @@ doctest = false
 [features]
 default = ["std"]
 std = ["prost/std"]
+# Derive serde traits for the structs defined in this crate (unless they contain optional fields)
+serde-derive = ["serde"]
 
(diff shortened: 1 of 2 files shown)


danburkert:
Hi @daladim, thanks for the PR.  I agree with the motivation for this change, but unfortunately this implementation will not be able to land.  The files you have modified are auto-generated (see bootstrap.rs).  I would like to have a prost-build option that would result in the correct serde derive annotations being added to generated code.  Critical to this is that the serde format be compatible with the protobuf json serialization rules.  See #75 for the tracking issue.

PR #419, opened 2021-01-11
Add optional serde to prost-types

enfipy:
Adds optional `serde` to `prost-types`. If `serde` feature provided - prost-types will be with `serde::Deserialize, serde::Serialize`.

Additionally, I ran rustfmt. That fixed some `tag="<some>"` stuff.


danburkert:
Hi @enfipy, thanks for the PR.  This looks quite similar to #415, and appears to not be viable for the same reasons.

enfipy:
@danburkert Thank you for the answer. Are there any other blockers to make this in prost-build?

danburkert:
Yes, see my comment here: https://github.com/danburkert/prost/pull/415#issuecomment-752740067

PR #421, opened 2021-01-12
Make DecodeError::new public

sphw:
With the release of prost 0.7 `DecodeError::new` is marked as `pub(crate)`. This makes custom implementations of `Message` difficult to create. Since `DecodeError` is already used in a public trait (`Message`) it seems reasonable to be able to construct it externally

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -30,8 +30,9 @@ impl DecodeError {
     /// Creates a new `DecodeError` with a 'best effort' root cause description.
     ///
     /// Meant to be used only by `Message` implementations.
+    #[doc(hidden)]
     #[cold]
-    pub(crate) fn new(description: impl Into<Cow<'static, str>>) -> DecodeError {
+    pub fn new(description: impl Into<Cow<'static, str>>) -> DecodeError {
         DecodeError {
             inner: Box::new(Inner {
                 description: description.into(),


danburkert:
Just curious, what’s the use case for calling this method?   `Decode` isn’t really designed to be implemented by hand, I think that’s why I switched it to be private.  Note that nothing in the `prost::encoding` module is covered under server guarantees, it’s only public because of technical limitations in how visibility in macros works.

sphw:
We have a few custom protocol buffer implementations. One of them is a dynamic protocol buffer implementation that is zero copy where it's reasonable. Most of the code is based on Prost, and is interoperable with Prosts traits, hence wanting create DecodeError from another crate. 

In the future we will probably open source these implementations (or upstream them into Prost if there is interest), but for now we get enormous value out of Prost as a toolkit for encoding and decoding the protocol buffer spec. It allows us to piggyback on the great optimization work that has been done in Prost.

danburkert:
OK I'm happy to make this public, with the understanding that it's an 'at your own risk' style API, similar to the public APIs in `prost::encoding`.  Toward that end, @sphw will you also add a `#[doc(hidden)]` annotation on the method to make it clear that it's not part of the public API?

sphw:
Done! Thanks for the quick response. Is there anyway we could get a minor version pushed with these changes? These changes are backwards compatible so it should be safe for a minor release.

sphw:
@danburkert Just bumping to see if this can get merged

garbageslam:
oh, i just reported a duplicate of this as an issue -- thank you for making this PR!

https://github.com/danburkert/prost/issues/432

danburkert:
Thanks!

PR #436, opened 2021-02-11
Remove the downloading of big test data that's no longer available

carols10cents:
Hi! It looks like the build's broken because the benchmarking test data at https://storage.googleapis.com/protobuf_opensource_benchmark_data/datasets.tar.gz isn't there anymore. 

I've [opened an issue with protobuf](https://github.com/protocolbuffers/protobuf/issues/8283) to see if it's been moved somewhere (that URL is still referenced in their repo too), but I figured in the meantime removing its usage from prost would be a good idea.


danburkert:
Awesome, thanks for the thorough writeup and filing the upstream issue!

PR #439, opened 2021-02-14
Fix #438: Check for overflow in `Duration` and `Timestamp` processing

### LINKED ISSUE #438: Conversion from `prost_types::Timestamp` to `SystemTime` may panic
In prost-types 0.7.0, the conversion from `Timestamp` to `SystemTime` uses the `+` and `-` operators on `UNIX_EPOCH`. [This can overflow and panic](https://github.com/rust-lang/rust/blob/e1884a8e3c3e813aada8254edfa120e85bf5ffca/library/std/src/time.rs#L543), creating a denial-of-service vulnerability if the input `Timestamp` is untrusted.

Also, the conversion involves calling `Timestamp::normalize`, which uses the `+` and `-` operators. This may panic or wrap around (depending on compiler settings), also creating a denial-of-service vulnerability if the application is compiled to panic on overflow.

# Proof of concept

```rust
use prost_types::Timestamp;
use std::time::SystemTime;

SystemTime::from(Timestamp {
    seconds: i64::MAX,
    nanos: 0,
}); // panics on i686-unknown-linux-gnu (but not x86_64) with default compiler settings

SystemTime::from(Timestamp {
    seconds: i64::MAX,
    nanos: i32::MAX,
}); // panics on x86_64-unknown-linux-gnu with default compiler settings
```

# Fixing

`Timestamp::normalize` should probably use `saturating_{add,sub}`. This can silently change the timestamp by up to about 3 seconds if its `nanos` field is out of range, but such a timestamp is arguably invalid anyway, so that's probably okay.

`SystemTime` does not have `saturating_{add,sub}` methods, nor `MIN` and `MAX` constants. Therefore, unlike `Timestamp::normalize`, there is no way to silently clamp the input timestamp to the valid range of `SystemTime`s. That's probably a bad idea anyway, since `SystemTime` could be based on a 32-bit `time_t` or Windows `FILETIME`, in which case clamping it could change the timestamp by much more than a mere few seconds. I recommend making the conversion fallible again and using `SystemTime::checked_{add,sub}`.

argv-minus-one:
This PR fixes the issues described in #438.

**Breaking change:** Because conversion from `Timestamp` to `SystemTime` can overflow by a platform-dependent, non-trivial amount, this PR removes `impl From<Timestamp> for SystemTime` and replaces it with a `TryFrom` implementation that fails on overflow.


prost-types/src/lib.rs
```diff
@@ -243,4 +339,156 @@ mod tests {
...
         );
     }
+
+    #[test]
+    fn check_duration_normalize() {
+        #[rustfmt::skip] // Don't mangle the table formatting.
+        let cases = [
+            // --- Table of test cases ---
+            //        test seconds      test nanos  expected seconds  expected nanos
+            (line!(),            0,              0,                0,              0),
```
danburkert, line 349: I've never seen this `line!()` trick before - it is absolutely brilliant.

danburkert:
Great PR, thanks for the thorough test cases (and for teaching me about the `line!()` trick!).  See https://github.com/danburkert/prost/issues/438#issuecomment-803671538 for question about approach - if possible I'd like to retain a deprecated version of the `From` impl for now (which can just call the TryFrom and unwrap).

argv-minus-one:
I don't believe that's possible. [Trait implementations cannot be deprecated.](https://github.com/rust-lang/rust/issues/39935)

olix0r:
This will fix https://github.com/danburkert/prost/issues/423 as well.

olix0r:
@danburkert @argv-minus-one is there anything I can do to help get this into a released version of prost-types? This bug has a security impact for services that can handle durations & timestamps from untrusted clients.

argv-minus-one:
@olix0r: I've rebased against current `master`, but that's about all I can think of. It's up to @danburkert to decide whether to merge this and whether there's anything else I need to do first.

danburkert:
Couple of thoughts on this PR:

1. Overall, I think the `TryFrom` impl in this PR should definitely land.
2. It may make sense to continue to provide a 'comment deprecated' `From` impl until the 0.8.0 release, even if it can't be tagged as deprecated through the normal rust annotation.
3. Long term, I think we should look into whether it would make sense to have `Timestamp` and `Duration` uphold the range contract, documented upstream to be [`0001-01-01T00:00:00Z to 9999-12-31T23:59:59Z`](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#timestamp) for timestamp and [` -315,576,000,000 to +315,576,000,000 `](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.Duration) for duration.  This will require a hand-rolled `impl Message`, but if 'simply' defers to a derived private type it's not too bad.  It will also require making the fields private and providing a ctor fn which checks the range.  I haven't checked against all platforms, but hopefully these ranges are tight enough that the non-fallible `From` impl can remain?

argv-minus-one:
>It may make sense to continue to provide a 'comment deprecated' From impl until the 0.8.0 release, even if it can't be tagged as deprecated through the normal rust annotation.

Okay. Is there a particular style or HTML pattern or something that you'd like me to use for that?

>I haven't checked against all platforms, but hopefully these ranges are tight enough that the non-fallible From impl can remain?

That doesn't help, no.

* On platforms that use 64-bit Unix time, the conversion is infallible, but it doesn't need that restriction. `prost_types::Timestamp`'s native representation of time *is* 64-bit Unix time, so the conversion is infallible anyway.
* On platforms that use 32-bit Unix time, time begins in 1902 (long after year 1) and ends in 2038 (long before year 9999), so the conversion is fallible.
* On Windows, time begins in 1601 and ends in 60055. Protobuf's upper bound is within that range, but the lower bound is not, so the conversion is fallible.
* Other platforms may have their own idea of when time begins and ends, which the Protobuf range may or may not fit into.

olix0r:
> Okay. Is there a particular style or HTML pattern or something that you'd like me to use for that?

Perhaps something like [this comment in tracing](https://github.com/tokio-rs/tracing/blob/eaaf8fdb89f9b8f79d4341061e0d4c63d3c3d8ed/tracing-core/src/collect.rs#L347-L355)?

```rust
/// **This implementation is deprecated.**
///
/// ... explanation ...
```

LucioFranco (approved):

PR #442, opened 2021-02-27
Add option to prost-build to skip the run of protoc

dfreese:
Our build system ends up generating all of the relevant protobuf files
already, so this ends up being a redundant step for us.  This allows the
running of the protoc step to be skipped.

There are some API concerns with this.  The inputs to compile_protos are
now ignored if that flag is set.  I didn't know to balance that with
Config::file_descriptor_set_path, which effectively allows for an input
of a filename into the config.  I could create a separate entry, similar
to compile_protos, that takes either:

1. Nothing and uses the config function input.  This, however, can be
   left unspecified and error out, which isn't super user-friendly.
2. A path, but this would now ignore the existing config input function
   if it was passed in.

I'm happy to put up a change that does that, but it wasn't clear either
was great, so I ended up going with the simplest approach.


LucioFranco:
@dfreese is this something you all are still interested in?

dfreese:
> @dfreese is this something you all are still interested in?

Yes, it would be.  I think it's still something that's broadly useful (see [here with grpc-rs](https://github.com/tikv/grpc-rs/blob/master/compiler/src/prost_codegen.rs#L66-L69)).  Given a chance to break APIs, this wouldn't be the exact one I'd chose, but the base feature is still of interest.

LucioFranco (approved):

PR #443, opened 2021-02-27
Fix clippy lints with 1.50.0

dfreese:
1.50.0 introduced a new clippy lint that has one false positive that
this disables and one that was actually valid.  Given a refactor in that
file, prost_attrs was made into a private function.

Note that #436, or similar, is also needed to get CI green again.


danburkert (approved):
Thanks!

PR #444, opened 2021-03-01
Update from deprecated criterion::Benchmark to BenchmarkGroup

dfreese:
This gets rid of the warnings generated by the benchmark code.  This
does not completely replecate the old behavior, which created three
benchmarks, all with their own throughput, but given the same name.  I
don't think that same thing is possible anymore with the non-deprecated
APIs.


danburkert:
Looks great, thanks @dfreese. Will merge shortly.

PR #445, opened 2021-03-06
Remove Cargo.lock from gitignore

SCOTT-HAMILTON:
Cargo.lock is required to make prost build in NixOS with standard build tools, I could make it myself and include it in nixos' repo but this would be a pain to maintain at each prost update. And it also seems like a good practice to include it in your rust projects, see : https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,1 @@
 target
-Cargo.lock


danburkert:
Hey @SCOTT-HAMILTON, the link you included says

danburkert:
I'm going to close this out since I don't think it's appropriate for a library-only crate, but please feel free to re-open if I'm overlooking something.

PR #446, opened 2021-03-12
Documentation improvements

### LINKED ISSUE #69: Document Enumerations
README says:

*All .proto enumeration types convert to the Rust i32 type. Additionally, each enumeration type gets a corresponding Rust enum type, with helper methods to convert i32 values to the enum type. The enum type isn't used directly as a field, because the Protobuf spec mandates that enumerations values are 'open', and decoding unrecognized enumeration values must be possible.*

It took me some (rather cheerless) time to discover that I can convert an enumeration like the one in the README example

```rust
    #[derive(Clone, Copy, Debug, PartialEq, Eq, Enumeration)]
    pub enum PhoneType {
        Mobile = 0,
        Home = 1,
        Work = 2,
    }
```

to the value this way:

```rust
PhoneType::Mobile as i32
```

However, for the opposite direction, I'm lost for good. How do I convert an i32 code to PhoneType? Where do I find those "helper methods to convert i32 values to the enum type"? I'd really appreciate an example in the README. Thank you.

Furthermore, what is the `#[derive(Enumeration)]` ? Can't find the Enumeration trait anywhere..

carols10cents:
This is mostly to fix #69 by documenting generated enum conversion functions in the README. 

I also wanted to make sure the generated code example was up to date, and it wasn't, so there's a commit for that. 

And while setting up a project to generate the code from the `tutorial` example, I hit a small hiccup in `prost-build`'s documentation, so there's a fix for that :)

Thank you! I'm happy to make any changes ❤️

### DIFF
--- a/README.md
+++ b/README.md
@@ -215,27 +215,29 @@ message AddressBook {
 and the generated Rust code (`tutorial.rs`):
 
 ```rust
-#[derive(Clone, Debug, PartialEq, Message)]
+#[derive(Clone, PartialEq, ::prost::Message)]
 pub struct Person {
     #[prost(string, tag="1")]
-    pub name: String,
+    pub name: ::prost::alloc::string::String,
     /// Unique ID number for this person.
     #[prost(int32, tag="2")]
     pub id: i32,
     #[prost(string, tag="3")]
-    pub email: String,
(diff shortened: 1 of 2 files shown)


danburkert:
This is fantastic, thank you @carols10cents!

PR #449, opened 2021-03-19
Enable zero-copy support for bytes::Bytes fields

pluth:
I just noticed this while poking through the code. Fields of type `Bytes` can now be slices of a greater `Bytes` that is passed to Message::decode

The referenced issue https://github.com/tokio-rs/bytes/issues/374 was implemented as Buf::copy_to_bytes and `Bytes` features the zero-copy implementation https://docs.rs/bytes/1.0.1/src/bytes/bytes.rs.html#526-560


danburkert:
Thanks!

PR #455, opened 2021-04-11
Fix messages with optionals and oneofs

### LINKED ISSUE #430: proto3 optional
In documentation you write:

`The table of field modifiers above applies to message fields, except that proto3 message fields without a modifier (the default) will be wrapped in an Option`

But that's not true in my case. Leaving all modifiers will not create an option for me. The only way to do so is use proto2 and optional modifier

sfackler:
Optional scalar fields create a synthetic oneof, but we don't add those fields to the oneof_fields map since we just want to treat them as optional fields. However, if a message containing an optional scalar field also contained a real oneof, the code would enter the oneof processing logic and panic trying to load the fields of the synthetic oneof. We now just treat a lack of oneof fields as an indication that this oneof was synthetic and skip it.

Closes #430

### DIFF
--- a/prost-build/src/code_generator.rs
+++ b/prost-build/src/code_generator.rs
@@ -237,12 +237,12 @@ impl<'a> CodeGenerator<'a> {
 
             for (idx, oneof) in message.oneof_decl.into_iter().enumerate() {
                 let idx = idx as i32;
-                self.append_oneof(
-                    &fq_message_name,
-                    oneof,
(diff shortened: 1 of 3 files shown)


sfackler:
Semi-related, but it might be nice to update the bundled protoc version to 3.15 so proto3 optional support doesn't require an experimental flag anymore. I can make another PR for that if it seems reasonable to you.

PR #459, opened 2021-04-12
Bundled protoc only if the interpreter exists

nagisa:
Otherwise on systems such as NixOS the prost-build attempts to execute
the bundled protoc, hits a very obscure error about file not existing
and exits, without trying the other options (such as the protoc from
path, if it exists there).

This also happens to cover the musl case, which has its interpreters as
`/lib/ld-musl-...` instead, as well as various instances where glibc might
want to bump the so version for its interpreter (has recently happened in
the past for x86_64 IIRC).


LucioFranco (approved):
Sorry for the delay on this! Thanks!

Issue #207, opened 2019-08-01
prost-build: keep protobuf's enum variant name as is


n4to4:
It seems that enum variant names are generated with `to_upper_camel`, and there is no option to customize that.

I'd like to use the same variant names both for protobuf and the generated Rust code, because
- to serialize/deserialize the enum values to/from plain text, as well as protobuf
- if I have two variant names like `Code1_2` and `Code12` in an enum, they will get conflict when converted with `to_upper_camel`

Would it be possible to add a configuration option, for instance, to the `prost_build::Config` builder?

danburkert:
Yes, I think that should be possible.  Sounds like a good potential feature for `prost`.

Voronar:
Until this https://github.com/danburkert/prost/pull/336 changes will be merged there is no good type safe way for casting `Any` to concrete message. I have workaround like this:
```rust

/* test.proto
syntax = "proto3";
package proto3.test;

message DNAExample {
  uint32 uint = 1;
  string str = 2;
}
*/

pub mod protobuf {
  include!(concat!(env!("OUT_DIR"), "/protoapi.rs"));

  #[derive(Clone, PartialEq, ::prost::Message)]
  pub struct DnaExample {
      #[prost(uint32, tag="1")]
      pub uint: u32,
      #[prost(string, tag="2")]
      pub str: ::prost::alloc::string::String,
  }
}

#[macro_export]
macro_rules! cast_any {
[$castin_type:ident] => {
    |attr: &prost_types::Any| {
    use prost::{Message, bytes::Bytes};

      let type_url =
        format!("type.googleapis.com/gobgpapi.{}", stringify!($castin_type));
      if attr.type_url == type_url {
        let value = protobuf::$castin_type::decode(Bytes::from(attr.value.clone()));
        Some(value)
      } else {
        None
      }
    }
  };
}

fn main() {
  // type.googleapis.com/gobgpapi.DNAMessage !== type.googleapis.com/gobgpapi.DnaMessage
  let attr = cast_any![DnaExample](&any_message); // None
}
```
But it doesn't work correctly all the time because protobuf messages renames during codegen process by default as @n4to4 has mentioned above in subject.

Voronar:
One things to note.
Then we use `type_attribute` function we need to set original message name from proto-file.

caspermeijn:
The given examples names are advices against in the protobuf style guide. So from my point of view these names should be improved. Generators for other languagues will have the same problem. https://protobuf.dev/programming-guides/style/

PR #467, opened 2021-04-28
Preserves original *.proto message and enum names

Voronar:
Feature request (https://github.com/tokio-rs/prost/issues/207) implementation.

### DIFF
--- a/prost-build/src/code_generator.rs
+++ b/prost-build/src/code_generator.rs
@@ -39,6 +39,13 @@ pub struct CodeGenerator<'a> {
 }
 
 impl<'a> CodeGenerator<'a> {
+    fn to_upper_camel(&self, value: &str) -> String {
+        if self.config.preserve_original_names {
+            value.into()
+        } else {
+            to_upper_camel(value)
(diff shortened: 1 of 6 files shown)


LucioFranco:
If I understand correctly this change would allow prost-build to generate code with warnings?

Voronar:
@LucioFranco Yes. But one can set warning disabling attribute like this:
```rust
pub mod protobuf {
  #![allow(non_camel_case_types)]
  include!(concat!(env!("OUT_DIR"), "/protoapi.rs"));
}
```

caspermeijn:
The intention of this project is to generate idiomatic code. The naming convention in Rust for structs is UpperCamelCase. If we ignore that conversion, the rust compiler will generate a warning and that is unwanted.

The examples in the test case are not very convincing, because the protobuf style guide advices against that spelling: https://protobuf.dev/programming-guides/style/

I feel like the `prost` code will become more complicated with this change, and it doesn't align with the description in the  `prost` README. Therefore I am closing the PR.

PR #475, opened 2021-05-09
Fix code generation where enum variant name is a number

### LINKED ISSUE #132: Prepend underscore to numerical enum variants
After stripping of enum prefix from enum variant name, some constants may have invalid name because they start with number. e.g. `Time3s = 0,` -> `3s = 0,` . As workaround, prepend `_` to name of such constants, e.g. `Time3s = 0,` -> `_3s = 0,` .

Arkanayan:
Prepend underscore to the enum variant name

Fixes #132


LegNeato:
I believe this is fixed more generally in https://github.com/danburkert/prost/pull/466.

PR #477, opened 2021-05-15
Add attributes just to messages.

jozn:
This add attributes just to Rust structs.

### DIFF
--- a/prost-build/src/code_generator.rs
+++ b/prost-build/src/code_generator.rs
@@ -171,6 +171,7 @@ impl<'a> CodeGenerator<'a> {
             });
 
         self.append_doc(&fq_message_name, None);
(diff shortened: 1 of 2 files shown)


LucioFranco:
Im not totally following how this approach is different than the type attributes but it would also be good to add some tests.

jozn:
Sorry for late reply.
Type attributes adds them to both rust enums and structs, this one adds attributes just to struct.  
This PR is for a long time ago, at the time I needed to add some attributes to rust struct only, after this PR I have changed things and didn't need this functionality anymore. Anyway as it's not a priority for my and i'm busy with some other projects feel free to close this PR.

LucioFranco:
sorry for the delay, thanks for your PR! Closing this in favor of #522

Issue #431, opened 2021-01-26
Allow adding custom derive macros


cemoktra:
Not sure if the proto options allow some per file or per package options. But there are some java specific options for example. And in my case it would be nice to add a derive macro to the generated struct in order to avoid having the struct again and writing conversion

cemoktra:
Also it may not also be interesting to not only add derive, but other macros

ogoes:
Hi, I'm new here.

I really like this feature!

I´ll try submit some code.

danburkert:
Hi, please add more specific details about what this feature request.  Are `prost_build::Config::type_attribute` and `prost_build::Config::field_attribute` not sufficient for the usecase?

cemoktra:
> Hi, please add more specific details about this feature request. Are `prost_build::Config::type_attribute` and `prost_build::Config::field_attribute` not sufficient for the usecase?

Sorry i just came back to this. So `type_attribute` works actually but it feels kind of strange to put this information into the `build.rs` instead of the proto file while other language put these kind of information directly to the proto file. Anyway thank you

cemoktra:
@danburkert i had some glimpse into the source code. Before i start working on a PR let me try to discuss this with you.

In order to run `protoc -o` i will need to create a .proto file that extends the MessageOptions. After that i will be able to handle the the option MessageOption struct from which i could add options to field_attributes just as the build config does. Correct?

So the main question for me would be, how to add a `prost.proto` file to the includes that defines prost specific options that can be used

cs-clarence:
Are there any updates on this? This approach seems less error prone than specifying them in the build.rs file

davo417:
Hello, I'm trying to implement this functionality for prost+[influxdb](https://docs.rs/influxdb/latest/influxdb/) with the following config:

```
fn main() -> Result<(), Box<dyn std::error::Error>> {
    tonic_build::configure()
        .type_attribute("sensor.SensorReading", "#[derive(InfluxDbWriteable)]")
        .field_attribute("sensor.SensorReading.id", "#[influxdb(tag)]")
        .compile(&["./protos/sensor.proto"], &["./protos/"])?;
    Ok(())
}
```

But when compiling I encounter the following error:

```
error: cannot find derive macro `InfluxDbWriteable` in this scope
 --> /home/david/projects/rustwp/ecs/target/debug/build/ecs-f5fecc71596c1f28/out/sensor.rs:4:10
  |
4 | #[derive(InfluxDbWriteable)]
[...]

PR #482, opened 2021-06-02
Add struct attributes from proto file

cemoktra:
Issue #431


raptros:
this same approach could also add support for field attributes as well, right?

cemoktra:
Yes could be done the same way

PR #488, opened 2021-06-18
add configuration for prost and prost_type paths

cab:
## Motivation

Allowing users to configure the `prost` and `prost_types` paths lets crates "wrap" those crates as an implementation detail. This is particularly useful for crates that attempt to expose a "framework" for rpc services. For instance, `tonic` could reexport `prost` and set this to use `::tonic::prost`, and users would not need to depend on `prost` directly.

## Solution

Added two new configuration methods and fields to `Config`: `prost_path` and `prost_types_path`. These default to `::prost` and `::prost_types`, respectively. 

If `prost_path` is changed from `::prost`, a new attribute is added to structs: `#[prost(path = "the_new_path")]`. This is used in `prost-derive` in order to use the correct path.


danburkert:
Is this what `tonic` wants to do?  My experience with building applications with `tonic` and `prost` is that custom code generation options are usually required (using `prost_build::Config`), and at that point it's going to be an impediment that the API split is duplicated and obscured. I can see this being a very slight win for 'hello world' minimal setup examples, but personally I think that's not a useful thing to optimize for.

prost-types/src/compiler.rs
```diff
@@ -67,7 +67,7 @@ pub struct CodeGeneratorResponse {
 pub mod code_generator_response {
     /// Represents a single generated file.
     #[derive(Clone, PartialEq, ::prost::Message)]
-    pub struct File {
+        pub struct File {
```
danburkert, line 70: This indentation change is a regression.

cab:
> Is this what `tonic` wants to do? My experience with building applications with `tonic` and `prost` is that custom code generation options are usually required (using `prost_build::Config`), and at that point it's going to be an impediment that the API split is duplicated and obscured. I can see this being a very slight win for 'hello world' minimal setup examples, but personally I think that's not a useful thing to optimize for.

No, sorry! I didn't mean to speak for tonic here, I just thought it was a good example.

> I can see this being a very slight win for 'hello world' minimal setup examples, but personally I think that's not a useful thing to optimize for.

That's a fair point. This might be optimizing too much for app setup instead of realistic long-term use cases.

evergreen-trading-systems:
I have a related issue with this using `tonic` https://github.com/hyperium/tonic/issues/699#issuecomment-875975780. 

`::prost` expects `prost` to be in the same package that compiles protos. In my case, `prost` is imported from another package, and this causes the module not to be found.
image

I read the pr (https://github.com/tokio-rs/prost/pull/147) that addresses this, but Rust-2018 equivalent of `::prost` would be `crate::prost`?

wycats:
Any chance of getting this addressed?

danburkert:
@wycats can you give a bit of color around your usecase/motivation?  My thoughts are the same as what I wrote in https://github.com/tokio-rs/prost/pull/488#issuecomment-864586915, ie this is a nice-to-have for the simplest of getting started usecases, but ultimately users will need to pull in `prost` on non-trivial projects anyway.  If there's something I'm overlooking I'm glad to hear it, though.

caspermeijn (changes requested):
This does clutter the code generation a bit, but I think makes `Config::prost_path` actually usable.

Please rebase to latest master.

caspermeijn:
Fixes #939

chris-zen:
Hi, any plans to make progress here or #1274?

PR #494, opened 2021-06-30
allow matching fields by attributes

cab:
## Motivation

Allowing for more granular field matching enables placing attributes that only make sense or are only valid in certain contexts. For example, matching on `optional` fields allows placing a `#[serde(default)]` attribute.

## Solution

Update `PathMap` to be parameterized by the attribute type. This means that different attributes can be used for different types. For example, this PR only adds `FieldAttributes`, but things like `MessageAttributes` could be added.

Add a new struct, `Matcher`, which matches against a given path and attribute set. Matcher attributes must implement a crate-visible trait, `MatcherAttributes`, which indicates whether the attributes in a matcher match or not. I didn't use `Eq` because I wanted to do things like have `None` match any value.

Add a new `FieldAttributes` struct which is used to match against proto field attributes (`optional` and `repeated`).

Add a new method to `Config` named `field_attribute_matcher` that allows passing in a `Matcher`.

### DIFF
--- a/prost-build/src/code_generator.rs
+++ b/prost-build/src/code_generator.rs
@@ -18,12 +18,71 @@ use crate::ast::{Comments, Method, Service};
 use crate::extern_paths::ExternPaths;
 use crate::ident::{to_snake, to_upper_camel};
 use crate::message_graph::MessageGraph;
+use crate::path::MatcherAttributes;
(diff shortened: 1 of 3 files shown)


cab:
Closing, likely no longer needed!

PR #496, opened 2021-07-06
Remove path type generic for better ergonomic

aquarhead:
This could be especially useful when one parameter (e.g. `protos`) is dynamically collected while the other (`includes`) is statically defined. Splitting the type should make `compile_protos` easier to use, I also changed a few instances where type conversion was used to show the impact.

### DIFF
--- a/prost-build/src/lib.rs
+++ b/prost-build/src/lib.rs
@@ -698,10 +698,11 @@ impl Config {
     ///   Ok(())
     /// }
     /// ```
-    pub fn compile_protos<P>(&mut self, protos: &[P], includes: &[P]) -> Result<()>
-    where
-        P: AsRef<Path>,
-    {
+    pub fn compile_protos(
+        &mut self,
+        protos: &[impl AsRef<Path>],
+        includes: &[impl AsRef<Path>],
+    ) -> Result<()> {
         let target: PathBuf = self.out_dir.clone().map(Ok).unwrap_or_else(|| {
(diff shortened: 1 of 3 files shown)


prost-build/src/lib.rs
```diff
@@ -698,9 +698,10 @@ impl Config {
     ///   Ok(())
     /// }
     /// ```
-    pub fn compile_protos<P>(&mut self, protos: &[P], includes: &[P]) -> Result<()>
+    pub fn compile_protos<P, Pi>(&mut self, protos: &[P], includes: &[Pi]) -> Result<()>
```
LucioFranco, line 701: Can we change these both to be `impl AsRef<Path>` instead of generics on the fn?

LucioFranco (changes requested):

LucioFranco (approved):
Thanks!

LucioFranco:
@aquarhead looks like there is a test failure, could you fix that then we can merge this?

aquarhead:
I'll take a look

aquarhead:
So seems this will break if `includes` is empty, it previously works because it uses the type of  `protos` to infer the type of `&[]`. I'm not sure there's a way to fix this (the type hint requirement)?

It seems this change would improve ergonomics when I have `includes` but would have the reverse effect when it's empty. Though using `&[""]` seems to be a workaround.

LucioFranco (approved):

Issue #502, opened 2021-07-09
prost-build 0.8.0 only adds final `type_attribute`


tommilligan:
In prost-build 0.7.0, it was possible to add any number of type attributes, and all would be added to the generated struct. In the following minimal example:

```proto
// src/example.proto
syntax = "proto3";
package example;

message Example {
  string example_field = 1;
}
```

```rust
// build.rs
use prost_build::Config;

fn main() {
    let mut config = Config::new();
    config.type_attribute(".", "#[derive(Display)]");
    config.type_attribute(".", "#[derive(PartialOrd)]");
    config
        .compile_protos(&["./src/example.proto"], &["./src"])
        .unwrap();
}
```

Generates the following on prost-build 0.7.0:

```rust
#[derive(Display)]
#[derive(PartialOrd)]
#[derive(Clone, PartialEq, ::prost::Message)]
pub struct Example {
    #[prost(string, tag="1")]
    pub example_field: ::prost::alloc::string::String,
}
```

Note that all configured type_attributes are added in the order they are configured. However, with prost-build 0.8.0, the generated output is:

```rust
#[derive(PartialOrd)]
#[derive(Clone, PartialEq, ::prost::Message)]
pub struct Example {
    #[prost(string, tag="1")]
    pub example_field: ::prost::alloc::string::String,
}
```

Note that only the most recent `type_attribute`, `PartialOrd`, has been added.

This appears to be an unintentional breaking change, I expect introduced here: https://github.com/tokio-rs/prost/commit/79f0dfd8ed703fcce27f053c3b29c360c7ead5cb#diff-5283c009c22e833ce65e3b8a6d7f57c026e0562f79225a9375f8e738ad4080c2R252-R260

I'll investigate further and see if I can submit a fix today.

tommilligan:
After investigation, there appear two be two distinct changes introduced by the `PathMap` refactoring in https://github.com/tokio-rs/prost/commit/79f0dfd8ed703fcce27f053c3b29c360c7ead5cb. They are:

- calls to `type_attribute` and `field_attribute` are no longer cumulative. Calling again with the same pattern key will overwrite any previously inserted attributes. This is in contradiction to the docs [which state](https://docs.rs/prost-build/0.8.0/prost_build/struct.Config.html#method.type_attribute):

> The calls to this method are cumulative. They don’t overwrite previous calls and if a type is matched by multiple calls of the method, all relevant attributes are added to it.


[...]

Issue #490, opened 2021-06-21
Attribute configuration no longer allows multiple attributes per pattern


cab:
The `type_attributes`, `field_attributes`, etc methods are described as cumulative, but after this change there can only be one attribute per path.

It's not clear to me if this new behavior is the correct behavior or if the old behavior was unintentionally allowing this.

I _think_ this is a regression from https://github.com/tokio-rs/prost/commit/79f0dfd8ed703fcce27f053c3b29c360c7ead5cb. Previously, [multiple values could exist](https://github.com/tokio-rs/prost/commit/79f0dfd8ed703fcce27f053c3b29c360c7ead5cb#diff-702d1c883416266db565804e8243178a5e3b59e9c22dff7d3161247d13aad208L401) with the same matcher, but [now only one entry](https://github.com/tokio-rs/prost/commit/79f0dfd8ed703fcce27f053c3b29c360c7ead5cb#diff-5c6fd8964f3a708b4b4ae0231791a3119fc994c82c498c8a69af388567dbdd41R14) can exist in the `PathMap`'s `matchers` hashmap.

### Example

```rust
.type_attributes(".", "[derive(::serde::Serialize)]")
.type_attributes(".", "[serde(rename = \"camelCase\")]")
```

Before the regression (e.g. on `0.7`), this would output

```rust
#[derive(::serde::Serialize)]
#[serde(rename = "camelCase")]
struct MyProto {}
```

After the regression, this outputs
```rust
#[serde(rename = "camelCase")]
struct MyProto {}
```

cab:
if this is a regression, one possible fix might be making `Config#type_attributes` a `PathMap<Vec<String>>` and then adding something like this to `PathMap`:
```rust
pub(crate) fn extend<A>(&mut self, matcher: String, value: T) where T: Extend<A> + IntoIterator<Item = A> + Default {
    self.matchers.entry(matcher).or_default().extend(value);
}
```

LucioFranco:
Fixed in #522

PR #503, opened 2021-07-09
Support multiple attributes

pruthvikar:
Fixes https://github.com/tokio-rs/prost/issues/502 and https://github.com/tokio-rs/prost/issues/490


tommilligan:
This looks like a step in the right direction, but ref #502  it only fixes the issue of applying multiple attributes using the same key, not using different keys (which was also allowed in `0.7.0`).

See the second part of this comment: https://github.com/tokio-rs/prost/issues/502#issuecomment-877141665

LucioFranco:
#522 supersedes this. Thanks for your contribution!

PR #509, opened 2021-07-17
[prost-type] Impl `Hash + Eq` traits on `Timestamp`.

lytefast:
Implement `Hash` and an unstable/naive version of `Eq` for `Timestamp

## Eq
A basic equality check on the internal fields of the `Timestamp`.
This implies that `normalized_ts != non_normalized_ts` even if `normalized_ts == non_normalized_ts.normalized()`.

This is acceptable to as timestamps across the network are already unstable.

The edge case is when the `Timestamp` is not normalized yet and would have mutated during the `normalize()` call.

- [x] add tests

### DIFF
--- a/prost-types/src/lib.rs
+++ b/prost-types/src/lib.rs
@@ -165,18 +165,19 @@ impl Timestamp {
     }
 }
 
+/// Implements the unstable/naive version of `Eq`: a basic equality check on the internal fields of the `Timestamp`.
+/// This implies that `normalized_ts != non_normalized_ts` even if `normalized_ts == non_normalized_ts.normalized()`.
 #[cfg(feature = "std")]
+impl Eq for Timestamp {}
+
+#[cfg(feature = "std")]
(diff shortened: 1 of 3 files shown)


sfackler:
What is the purpose of implementing Hash but not Eq?

It seems like we could just implement both by making manual implementations that operate on a normalized version internally.

lytefast:
I decided to not go that route because I do not believe having eq or hash should do extra overhead calculations. In these cases I wasn't sure whether it was intended that a non-normalized timestamp should equal a normalized timestamp without the need to call `normalize`.

For our use case it's perfectly fine that `Timestamp` is `PartialEq` the only issue was the missing `Hash` impl since we were trying to insert a proto message into a hashmap. The `Timestamp`s in our cases should always be "normalized".

LucioFranco (approved):

lytefast:
unfortunately this hash collections also require `Eq` not sure how I missed that.

lytefast:
Updated PR Description, tests, and impl.

PR #510, opened 2021-07-19
Escape comments

### LINKED ISSUE #495: Escape brackets in the doc
I have a source file containing `[EXPERIMENTAL]` markers in the comments. This gets translated to doc comment without escaping leading to false links and warnings. Escaping it (`\[`, `\]`) would solve the issue.

nbaztec:
The PR fixes the problem of escaping URLs and Brackets in comments which fail on `clippy` lints.

Fixes #495

### DIFF
--- a/prost-build/src/ast.rs
+++ b/prost-build/src/ast.rs
@@ -95,8 +95,8 @@ impl Comments {
     ///     - escape `[` & `]`
     fn sanitize_line(line: &str) -> String {
         lazy_static! {
-            static ref RULE_URL: Regex = Regex::new(r"https?://\S+").unwrap();
-            static ref RULE_BRACKETS: Regex = Regex::new(r"(\[)([^\]]+)(])").unwrap();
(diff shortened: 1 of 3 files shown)


prost-build/src/ast.rs
```diff
@@ -83,10 +85,24 @@ impl Comments {
...
+            buf.push_str(&Self::sanitize_line(line));
             buf.push('\n');
         }
     }
+
+    /// Sanitizes the line for rustdoc by performing the following operations:
+    ///     - escape urls as <http://foo.com>
+    ///     - escape `[` & `]`
+    fn sanitize_line(line: &str) -> String {
+        lazy_static! {
```
LucioFranco, line 97: Not totally sure if we really need lazy static but don't think it hurts that much.

LucioFranco (approved):

LucioFranco:
@nbaztec looks like there is a test failure, can you fix that then we can merge?

Kixunil:
Thanks!

PR #513, opened 2021-07-20
Clarify docs for includes arg

adamchalmers:
I spent a lot of time confused by why prost wasn't finding my .proto import files. Turns out I had falsely assumed that imports were searched recursively. This is definitely my bad, I'm very new to protobuf, but I think it would help other new users to clarify this.


adamchalmers:
OK, nevermind, I had the wrong mental model, the search _is_ recursive.

PR #515, opened 2021-08-02
Fix escape for default string

ldm0:
The `default` is already unescaped.


danburkert:
> The `default` is already unescaped.

Interesting, I definitely didn't realize that!  Please also add a corresponding regression test to the `tests` crate.  It looks like we don't have an existing .proto which really fits, but you can add one.  See `default_enum_value.proto` / `test_default_enum()` for an example.

danburkert (approved):
Awesome, thanks!

Issue #507, opened 2021-07-10
prost-build 0.8.0 regression: `field_attribute` on a oneof field gets added to enum variants


appaquet:
Adding field attributes on oneof fields in prost-build 0.8.0 also adds the attributes to the enum variants. In this example, adding `#[serde(flatten)]` to `TestMessage.some_field` correctly adds the attribute to the `TestMessage.some_field` field of the struct, but also adds the attribute to each enum variant.

proto
```proto
syntax = "proto3";

package test;

message TestMessage {
    oneof some_field {
        string a_field = 1;
        string b_field = 2;
    }
}
```

build.rs
```rust
    prost_build::Config::new()
        .type_attribute("TestMessage", "#[derive(Serialize, Deserialize)]")
        .type_attribute(
            "TestMessage.some_field",
            "#[derive(Serialize, Deserialize)]",
        )
        .field_attribute("TestMessage.some_field", "#[serde(flatten)]")
        .compile_protos(&vec!["./protos/test.proto"], &["./protos/"])
        .expect("prost error");
```

Outputs:
```rust
#[derive(Serialize, Deserialize)]
#[derive(Clone, PartialEq, ::prost::Message)]
pub struct TestMessage {
    #[prost(oneof="test_message::SomeField", tags="1, 2")]
    #[serde(flatten)]
    pub some_field: ::core::option::Option<test_message::SomeField>,
}
/// Nested message and enum types in `TestMessage`.
pub mod test_message {
    #[derive(Serialize, Deserialize)]
    #[derive(Clone, PartialEq, ::prost::Oneof)]
    pub enum SomeField {
        #[prost(string, tag="1")]
        #[serde(flatten)]
        AField(::prost::alloc::string::String),
        #[prost(string, tag="2")]
        #[serde(flatten)]
        BField(::prost::alloc::string::String),
    }
}
```
Notice the extra `#[serde(flatten)]` on each variant. This results in `unknown serde variant attribute flatten` since serde macro gets wrongly added to the enum variants.

LucioFranco:
Fixed in #522

wagenet:
This issue does not appear to be fixed. It still happens to me in master.

markmmm:
Is this an example of the same issue:
```protobuf
syntax="proto3";

package api;
message Init {
    string msg_id = 1;
    message SetOs {
        enum OS {
            OS_UNKNOWN = 0;
            OS_WIN     = 1;
            OS_MAC     = 2;
        }
[...]

PR #522, opened 2021-08-27
Fix multiple attributes support

Guiguiprim:
Improves `PathMap`:
- Add iterator over all values matching a path (allow multiple attributes, with the same path and with sub-paths)  #490 #502
- Keep insertion order to avoid warnings/errors https://doc.rust-lang.org/rustc/lints/listing/warn-by-default.html#legacy-derive-helpers
- Correctly handle field attributes on one-of #507

### DIFF
--- a/prost-build/src/code_generator.rs
+++ b/prost-build/src/code_generator.rs
@@ -38,6 +38,12 @@ pub struct CodeGenerator<'a> {
     buf: &'a mut String,
 }
 
+fn push_indent(buf: &mut String, depth: u8) {
+    for _ in 0..depth {
+        buf.push_str("    ");
+    }
+}
+
 impl<'a> CodeGenerator<'a> {
     pub fn generate(
         config: &mut Config,
@@ -256,25 +262,22 @@ impl<'a> CodeGenerator<'a> {
 
     fn append_type_attributes(&mut self, fq_message_name: &str) {
         assert_eq!(b'.', fq_message_name.as_bytes()[0]);
-        // TODO: this clone is dirty, but expedious.
-        if let Some(attributes) = self.config.type_attributes.get(fq_message_name).cloned() {
-            self.push_indent();
-            self.buf.push_str(&attributes);
+        for attribute in self.config.type_attributes.get(fq_message_name) {
+            push_indent(&mut self.buf, self.depth);
+            self.buf.push_str(&attribute);
             self.buf.push('\n');
         }
     }
 
     fn append_field_attributes(&mut self, fq_message_name: &str, field_name: &str) {
         assert_eq!(b'.', fq_message_name.as_bytes()[0]);
-        // TODO: this clone is dirty, but expedious.
-        if let Some(attributes) = self
+        for attribute in self
             .config
             .field_attributes
             .get_field(fq_message_name, field_name)
-            .cloned()
         {
-            self.push_indent();
-            self.buf.push_str(&attributes);
+            push_indent(&mut self.buf, self.depth);
+            self.buf.push_str(&attribute);
             self.buf.push('\n');
         }
     }
@@ -315,7 +318,7 @@ impl<'a> CodeGenerator<'a> {
             let bytes_type = self
                 .config
                 .bytes_type
-                .get_field(fq_message_name, field.name())
+                .get_first_field(fq_message_name, field.name())
                 .copied()
                 .unwrap_or_default();
             self.buf
@@ -427,7 +430,7 @@ impl<'a> CodeGenerator<'a> {
         let map_type = self
             .config
             .map_type
-            .get_field(fq_message_name, field.name())
+            .get_first_field(fq_message_name, field.name())
             .copied()
             .unwrap_or_default();
         let key_tag = self.field_type_tag(key);
@@ -570,10 +573,10 @@ impl<'a> CodeGenerator<'a> {
         let append_doc = if let Some(field_name) = field_name {
             self.config
                 .disable_comments
-                .get_field(fq_name, field_name)
+                .get_first_field(fq_name, field_name)
                 .is_none()
         } else {
-            self.config.disable_comments.get(fq_name).is_none()
+            self.config.disable_comments.get(fq_name).next().is_none()
         };
         if append_doc {
             Comments::from_location(self.location()).append_with_indent(self.depth, &mut self.buf)
@@ -712,9 +715,7 @@ impl<'a> CodeGenerator<'a> {
     }
 
     fn push_indent(&mut self) {
-        for _ in 0..self.depth {
-            self.buf.push_str("    ");
-        }
+        push_indent(&mut self.buf, self.depth);
     }
 
     fn push_mod(&mut self, module: &str) {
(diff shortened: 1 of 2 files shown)


prost-build/src/code_generator.rs
```diff
@@ -251,26 +251,35 @@ impl<'a> CodeGenerator<'a> {
...
             .field_attributes
             .get_field(fq_message_name, field_name)
-            .cloned()
         {
-            self.push_indent();
-            self.buf.push_str(&attributes);
-            self.buf.push('\n');
+            for _ in 0..self.depth {
+                buf.push_str("    ");
+            }
```
LucioFranco, line 277: Why not call `self.push_indent` here?

prost-build/src/code_generator.rs
```diff
@@ -251,26 +251,35 @@ impl<'a> CodeGenerator<'a> {
...
+            buf.push_str(&attribute);
+            buf.push('\n');
+        }
+        if !buf.is_empty() {
```
LucioFranco, line 281: Is this check needed? Appending a empty str should be a noop

prost-build/src/code_generator.rs
```diff
@@ -251,26 +251,35 @@ impl<'a> CodeGenerator<'a> {
...
+            self.buf.push_str(&buf);
         }
     }
 
     fn append_field_attributes(&mut self, fq_message_name: &str, field_name: &str) {
         assert_eq!(b'.', fq_message_name.as_bytes()[0]);
-        // TODO: this clone is dirty, but expedious.
-        if let Some(attributes) = self
+        let mut buf = String::new();
```
LucioFranco, line 269: What is the benefit of this extra String? I don't see why you can't just push directly into the buf.

prost-build/src/path.rs
```diff
@@ -1,61 +1,107 @@
 //! Utilities for working with Protobuf paths.
 
-use std::collections::HashMap;
 use std::iter;
 
 /// Maps a fully-qualified Protobuf path to a value using path matchers.
 #[derive(Debug, Default)]
 pub(crate) struct PathMap<T> {
-    matchers: HashMap<String, T>,
+    // insertion order might actually matter (to avoid warning about legacy-derive-helpers)
```
LucioFranco, line 8: Nice catch 👍

prost-build/src/path.rs
```diff
@@ -1,61 +1,107 @@
...
-            // Then, try matching path suffixes.
-            .chain(suffixes(fq_path))
-            // Then, try matching path prefixes.
-            .chain(prefixes(fq_path))
-            // Then, match the global path. This matcher must never fail, since the constructor
-            // initializes it.
-            .chain(iter::once("."))
-            .flat_map(|path| self.matchers.get(path))
-            .next()
+    pub(crate) fn get<'a>(&'a self, fq_path: &'_ str) -> Iter<'a, T> {
```
LucioFranco, line 19: I think you can elide this lifetime

`pub(crate) fn get(&self, fq_path: &str) -> Iter<'_, T>`

prost-build/src/path.rs
```diff
@@ -1,61 +1,107 @@
...
+    pub(crate) fn get_field<'a>(&'a self, fq_path: &'_ str, field: &'_ str) -> Iter<'a, T> {
+        Iter::new(self, format!("{}.{}", fq_path, field))
+    }
+
+    #[allow(unused)]
+    pub(crate) fn get_best<'a>(&'a self, fq_path: &'_ str) -> Option<&'a T> {
+        self.find_best_matching(fq_path)
+    }
+
+    pub(crate) fn get_best_field<'a>(&'a self, fq_path: &'_ str, field: &'_ str) -> Option<&'a T> {
```
LucioFranco, line 32: would a better name here be `get_first_field()` and might be worth it to document this behavior.

prost-build/src/path.rs
```diff
@@ -1,61 +1,107 @@
...
+                }
+                None => return None,
+            }
+        }
+    }
+}
+
+impl<'a, T> std::iter::FusedIterator for Iter<'a, T> {}
+
+/// Given a fully-qualified path, retruns a sequence of paths:
```
LucioFranco, line 89:
```suggestion
/// Given a fully-qualified path, returns a sequence of paths:
```

LucioFranco (changes requested):
Overall this looks great, I left a few inline comments/questions.

prost-build/src/code_generator.rs
Guiguiprim: The extra string is adressing the removed TODO.

It avoids the need for cloning the map (lifetime issue between `&self` for itering the map and `&mut self` for editing the buffer).

prost-build/src/code_generator.rs
Guiguiprim: Probably not

prost-build/src/code_generator.rs
Guiguiprim: Because I'm using a local buffer ...

prost-build/src/code_generator.rs
LucioFranco: Ah got it didn't catch that borrowing issue.

prost-build/src/code_generator.rs
LucioFranco: Right xD many buf variables

prost-build/src/code_generator.rs
Guiguiprim: I improved it to remove the local buffer, only the call to `self.push_ident()` was problematic.

prost-build/src/code_generator.rs
Guiguiprim: Removed

prost-build/src/code_generator.rs
Guiguiprim: Reworked.

prost-build/src/path.rs
Guiguiprim: Done

prost-build/src/path.rs
Guiguiprim: Done

prost-build/src/path.rs
Guiguiprim: Done

prost-build/src/path.rs
Guiguiprim: Done

prost-build/src/code_generator.rs
```diff
@@ -38,6 +38,12 @@ pub struct CodeGenerator<'a> {
     buf: &'a mut String,
 }
 
+fn _push_indent(buf: &mut String, depth: u8) {
```
LucioFranco, line 41: Not following why this has a `_`?

Guiguiprim: Fixed

My C++ brain made me bug, I needed a local private free function which name would not conflict with the class method...

LucioFranco (approved):
Thanks!

Logarithmus:
@LucioFranco new release maybe? We can't use this fix until it ends up in a new release.

u0xy:
A new release would be awesome for this fix.

PR #527, opened 2021-09-05
Protect encoding::decode_varint from overflow

ajguerrer:
Fixes the infamous #528.

### DIFF
--- a/src/encoding.rs
+++ b/src/encoding.rs
@@ -66,7 +66,8 @@ where
 /// Decodes a LEB128-encoded variable length integer from the slice, returning the value and the
 /// number of bytes read.
 ///
-/// Based loosely on [`ReadVarint64FromArray`][1].
+/// Based loosely on [`ReadVarint64FromArray`][1] with a varint overflow check from 
+/// [`ConsumeVarint`][2].
 ///
 /// ## Safety
 ///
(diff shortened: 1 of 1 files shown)


src/encoding.rs
```diff
@@ -167,7 +169,12 @@ where
         let byte = buf.get_u8();
         value |= u64::from(byte & 0x7F) << (count * 7);
         if byte <= 0x7F {
-            return Ok(value);
+            // check for u64::MAX overflow
+            if count == 9 && byte >= 2 {
```
ajguerrer, line 173: Decided to add this check here in case moving it outside the first `if` caused some performance impact.

LucioFranco:
Do you know if the C++ code protects against this? From what I can tell it does not.

ajguerrer:
> Do you know if the C++ code protects against this? From what I can tell it does not.

@LucioFranco, your right! The C++ code will happily parse and overflow. I found this while studying the [Go code](https://github.com/protocolbuffers/protobuf-go/blob/master/encoding/protowire/wire.go#L358).

I understand that fixing this bug has debatable usefulness. It will not hurt my feelings at all if you decide not to merge this. I just thought it was interesting.

LucioFranco:
I appreciate it! @ajguerrer do you mind updating the code with a reference to that go code, both the function doc and the line you changed? I think if we have that then I am fine to merge, I just wanted to check if there was precedence.

LucioFranco:
Oh and also rebase against master

ajguerrer:
@LucioFranco hows this look?

LucioFranco (approved):

LucioFranco:
@ajguerrer there is one rustfmt error if you can fix that then we can merge

ajguerrer:
@LucioFranco Once more!

PR #519, opened 2021-08-15
Improve encode_var/decode_varint performance

quininer:
* using `put_u8` simplifies the `encode_varint` code. this slightly changed behavior, but it should have no effect on prost. It improve performance while reducing unsafe code.
* inline `decode_varint` and remove some unnecessary unsafe.

bench
```
varint/small/encode     time:   [214.82 ns 214.91 ns 215.00 ns]
                        change: [-69.573% -69.350% -69.126%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 15 outliers among 100 measurements (15.00%)
  4 (4.00%) low mild
  3 (3.00%) high mild
  8 (8.00%) high severe

varint/small/decode     time:   [231.53 ns 231.84 ns 232.21 ns]
                        change: [-17.817% -16.309% -14.947%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 14 outliers among 100 measurements (14.00%)
  6 (6.00%) high mild
  8 (8.00%) high severe

varint/small/encoded_len
                        time:   [144.18 ns 144.37 ns 144.61 ns]
                        change: [-3.0551% -2.0177% -1.0496%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 15 outliers among 100 measurements (15.00%)
  1 (1.00%) low mild
  2 (2.00%) high mild
  12 (12.00%) high severe

varint/medium/encode    time:   [1.1107 us 1.1183 us 1.1284 us]
                        change: [-18.071% -17.222% -16.488%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 9 outliers among 100 measurements (9.00%)
  4 (4.00%) high mild
  5 (5.00%) high severe

varint/medium/decode    time:   [414.66 ns 414.86 ns 415.12 ns]
                        change: [-9.6812% -9.1417% -8.4957%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 10 outliers among 100 measurements (10.00%)
  1 (1.00%) low mild
  1 (1.00%) high mild
  8 (8.00%) high severe

varint/medium/encoded_len
                        time:   [143.71 ns 143.79 ns 143.89 ns]
                        change: [-3.8550% -2.6239% -1.6465%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 7 outliers among 100 measurements (7.00%)
  4 (4.00%) high mild
  3 (3.00%) high severe

varint/large/encode     time:   [2.2581 us 2.2611 us 2.2657 us]
[...]

PR #530, opened 2021-09-06
increase decode_varint performance

ajguerrer:
I managed to increase the performance of encoding::decode_varint by removing some branching and getting rid of the slow decode loop. This sounds concerning, but I also added `bytes.len()` checks every step of the way.

This PR mostly came about though trial-and-error, but I suspect it has something to do with a branching/inline interaction. The result, at least on my machine, is sizable performance gains for small varints (I imagine the most common case, at least as far as tags go) and diminishing performance gains as the varints get larger.

Note: My machine is a old AMD FX-8150 running Linux WSL2. I would be interested to see how the performance is on a fancy new cloud machine.

```
varint/small/encode     time:   [254.83 ns 256.70 ns 259.09 ns]
                        change: [-0.2769% +3.2613% +9.8484%] (p = 0.28 > 0.05)
                        No change in performance detected.

varint/small/decode     time:   [224.79 ns 226.81 ns 229.44 ns]
                        change: [-39.192% -36.446% -33.858%] (p = 0.00 < 0.05)
                        Performance has improved.

varint/small/encoded_len
                        time:   [203.60 ns 205.40 ns 207.53 ns]
                        change: [-2.5341% +1.7540% +8.6075%] (p = 0.68 > 0.05)
                        No change in performance detected.


varint/medium/encode    time:   [1.3112 us 1.3206 us 1.3317 us]
                        change: [-10.751% -3.7931% +1.7344%] (p = 0.33 > 0.05)
                        No change in performance detected.

varint/medium/decode    time:   [580.60 ns 584.50 ns 589.62 ns]
                        change: [-30.263% -27.334% -22.984%] (p = 0.00 < 0.05)
                        Performance has improved.

varint/medium/encoded_len
                        time:   [205.49 ns 207.27 ns 209.46 ns]
                        change: [-6.0480% -0.5540% +3.8425%] (p = 0.86 > 0.05)
                        No change in performance detected.

varint/large/encode     time:   [2.5997 us 2.6160 us 2.6387 us]
                        change: [-2.7384% -0.3883% +1.9255%] (p = 0.77 > 0.05)
                        No change in performance detected.

varint/large/decode     time:   [1.0150 us 1.0234 us 1.0337 us]
                        change: [-9.7888% -3.3455% +1.8321%] (p = 0.35 > 0.05)
                        No change in performance detected.

varint/large/encoded_len
                        time:   [204.44 ns 206.36 ns 208.69 ns]
                        change: [-4.0713% -0.7775% +1.9785%] (p = 0.68 > 0.05)
                        No change in performance detected.

varint/mixed/encode     time:   [1.7585 us 1.7739 us 1.7949 us]
                        change: [-2.8892% -0.4886% +1.9897%] (p = 0.70 > 0.05)
                        No change in performance detected.

varint/mixed/decode     time:   [1.2819 us 1.2915 us 1.3041 us]
                        change: [-20.716% -14.960% -10.392%] (p = 0.00 < 0.05)
                        Performance has improved.

varint/mixed/encoded_len
                        time:   [203.70 ns 204.81 ns 206.03 ns]
                        change: [-8.2241% -4.1701% -0.9167%] (p = 0.02 < 0.05)
                        Change within noise threshold.
```

For reference the "before" decode measurements:

```
varint/small/decode     time:   [357.51 ns 360.22 ns 364.29 ns]
varint/medium/decode    time:   [811.81 ns 817.69 ns 823.87 ns]
varint/large/decode     time:   [1.0463 us 1.0545 us 1.0654 us]
varint/mixed/decode     time:   [1.4749 us 1.4832 us 1.4923 us]
```

Taking a page from @quininer (#519), I also tried out `buf.get_u8()`. Unfortunately, its still quite a bit slower than `bytes.len()` + `bytes.get_unchecked(index)` + `buf.advance(len)`. This is likely because of the difference between calling `buf.advance(len)` 1 time vs. calling `buf.advance(1)` `len` times.

### DIFF
--- a/src/encoding.rs
+++ b/src/encoding.rs
@@ -62,7 +62,7 @@ where
     }
     part0 -= 0x80;
 
-    if bytes.is_empty() {
+    if bytes.len() < 2 {
         return Err(DecodeError::new("invalid varint"));
     }
     let b = unsafe { *bytes.get_unchecked(1) };
@@ -73,7 +73,7 @@ where
     }
     part0 -= 0x80 << 7;
 
-    if bytes.is_empty() {
+    if bytes.len() < 3 {
         return Err(DecodeError::new("invalid varint"));
     }
     let b = unsafe { *bytes.get_unchecked(2) };
@@ -84,7 +84,7 @@ where
     }
     part0 -= 0x80 << 14;
 
-    if bytes.is_empty() {
+    if bytes.len() < 4 {
         return Err(DecodeError::new("invalid varint"));
     }
     let b = unsafe { *bytes.get_unchecked(3) };
@@ -96,7 +96,7 @@ where
     part0 -= 0x80 << 21;
     let value = u64::from(part0);
 
-    if bytes.is_empty() {
+    if bytes.len() < 5 {
         return Err(DecodeError::new("invalid varint"));
     }
     let b = unsafe { *bytes.get_unchecked(4) };
@@ -107,7 +107,7 @@ where
     }
     part1 -= 0x80;
 
-    if bytes.is_empty() {
+    if bytes.len() < 6 {
         return Err(DecodeError::new("invalid varint"));
     }
     let b = unsafe { *bytes.get_unchecked(5) };
@@ -118,7 +118,7 @@ where
     }
     part1 -= 0x80 << 7;
 
-    if bytes.is_empty() {
+    if bytes.len() < 7 {
         return Err(DecodeError::new("invalid varint"));
     }
     let b = unsafe { *bytes.get_unchecked(6) };
@@ -129,7 +129,7 @@ where
     }
     part1 -= 0x80 << 14;
 
-    if bytes.is_empty() {
+    if bytes.len() < 8 {
         return Err(DecodeError::new("invalid varint"));
     }
     let b = unsafe { *bytes.get_unchecked(7) };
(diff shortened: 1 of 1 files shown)


ajguerrer:
Initial commit was wrong and could invoke UB with `get_unchecked` (forgot I was working with `&[u8]` and not `Bytes`, which act like a cursor). Updated the PR description with new performance numbers. Unfortunately it made the perf improvements drop a bit, but hopefully still worth merging.

I thought about writing a unit test to capture my mistake, but the problem with invoking UB is that my test was passing for all the wrong reasons, so it ended up not being very useful.

jaskij:
You can not get rid of the slow decode loop, unfortunately.

`bytes::buf::Buf` documentation states that:
> The underlying storage may or may not be in contiguous memory.

Also, for `chunk()`:
> Returns a slice starting at the current position and of length between 0 and `Buf::remaining()`.

So you cannot rely on all bytes of the varint being in the slice returnd by `chunk()`. They could be spread among multiple chunks, depending on the actual implementation of the trait. This would be a subtle bug, which could or could not come up in the future, depending on what actual `Buf` implementations are used.

(For this same reason the default implementation of `Buf::chunks_vectored` seems buggy to me).

ajguerrer:
@jaskij Thanks for pointing that out. The naming behind `chunk()` makes more sense to me now too. I am going to go ahead and close this PR since it would not survive the tests added by #559.

jaskij:
I'm a stickler for adhering to specs, so there's that. You're welcome @ajguerrer

PR #534, opened 2021-09-09
explicitly use core::default::Default to avoid collisions

cab:
## Motivation

Currently, `Default` is implemented for proto messages by relying on the prelude scope. This can lead to name collisions. (For example, [this discord thread](https://discord.com/channels/500028886025895936/628706823104626710/885500518582079519)).

## Implementation

Use the fully qualified path to `Default` (`::core::default::Default`).


cab:
This was already done [here](https://github.com/tokio-rs/prost/blob/bb986cd78fd03b40f16845ab46865cd3be9a9f6c/prost-derive/src/lib.rs#L307).

LucioFranco (approved):

prost-derive/src/lib.rs
```diff
@@ -204,7 +204,7 @@ fn try_message(input: TokenStream) -> Result<TokenStream, Error> {
             }
         }
 
-        impl #impl_generics Default for #ident #ty_generics #where_clause {
+        impl #impl_generics ::core::default::Default for #ident #ty_generics #where_clause {
```
DerekStrickland, line 207: @cab 

This does not behave as expected in my use case

for this proto

```protobuf
message Block {
 string name = 1;
 bool required = 2;
 Spec nested = 3;
}
```

it generates this struct which at first glance seems reasonable

```rust
pub struct Block {
    #[prost(string, tag = "1")]
    pub name: ::prost::alloc::string::String,
    #[prost(bool, tag = "2")]
    pub required: bool,
    #[prost(message, optional, boxed, tag = "3")]
    pub nested: ::core::option::Option<::prost::alloc::boxed::Box<Spec>>,
}
```

But generates this compile time error

```
error[E0271]: type mismatch resolving `<fn() -> hclspec::Default {<hclspec::Default as std::default::Default>::default} as FnOnce<()>>::Output == Box<Spec>`
   --> src/proto/hashicorp.nomad.plugins.shared.hclspec.rs:106:28
    |
106 | #[derive(Clone, PartialEq, ::prost::Message)]
    |                            ^^^^^^^^^^^^^^^^ expected struct `hclspec::Default`, found struct `Box`
    |
    = note: expected struct `hclspec::Default`
               found struct `Box<Spec>`

Other structs that  fail in a similar manner.
```

LucioFranco: Are you replacing the Default used? Seems like `hclspec::Default` is a diff trait?

DerekStrickland: @LucioFranco  I'm not trying to replace it. Rather, @cab is trying to help me solve the name collision that occurs because the `hcl_spec.proto` defines its own `Default` message.

```protobuf
message Default {
 Spec primary = 1;
 Spec default = 2;
}
```

DerekStrickland: FYI when fixed, closes #533

cab: I initially only updated the trait impl. I just updated the `Default::default` calls to `::core::default::Default::default`, which should fix this other error.

LucioFranco (approved):

PR #548, opened 2021-10-14
Make it possible to only generate the services

Guiguiprim:
This make it possible to generate only the messages in one crate (with dependencies only on `prost`) and the services in another crate (with dependencies on the messages crate and some networking crate like `tonic`). This offers more possibilities in workspace organization (separation of concerns and better build parallelization).

For example:
- crate `Messages` (use prost-build with no `service_generator`)
- crate `Business-A` depends on `Messages`
- crate `Services` depends on `Messages` (use tonic-build with `service_only(true)`)
- crate `Binary` depends on `Services` and `Business-A`

### DIFF
--- a/prost-build/src/code_generator.rs
+++ b/prost-build/src/code_generator.rs
@@ -123,21 +123,23 @@ impl<'a> CodeGenerator<'a> {
             code_gen.package
         );
 
-        code_gen.path.push(4);
-        for (idx, message) in file.message_type.into_iter().enumerate() {
-            code_gen.path.push(idx as i32);
-            code_gen.append_message(message);
+        if !code_gen.config.service_only {
+            code_gen.path.push(4);
+            for (idx, message) in file.message_type.into_iter().enumerate() {
+                code_gen.path.push(idx as i32);
+                code_gen.append_message(message);
+                code_gen.path.pop();
+            }
             code_gen.path.pop();
-        }
-        code_gen.path.pop();
 
-        code_gen.path.push(5);
-        for (idx, desc) in file.enum_type.into_iter().enumerate() {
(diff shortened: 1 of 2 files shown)


danburkert:
Can you not get the same behavior by registering an `extern_path` mapping all of the symbols to the remote crate?  Practically speaking I think you'd need to do that to get the types to line up.

Guiguiprim:
`extern_path` only makes it possible for imported files to be generated in an other crate.

Here I want the content of one proto file to be split in two crates. The messages in one crate and the services in an other.
(In order for it to work I do need to re-export the messages in the services crate [so that the services implementation find the messages, using `super::`).

As an attached file, a proof-of-concept, I should have added it sooner.

[tonic-prost-test.tar.gz](https://github.com/tokio-rs/prost/files/7503351/tonic-prost-test.tar.gz)

But you are right I could split my proto file: `echo.messages.proto` and `echo.services.proto`, but I would prefer to avoid this.

Guiguiprim:
Hello,
Coming back on this.

For dependencies tree and build performance, I want to split in two crates the types and the clients declared in some protobuf that I can not split. Even more, I actually need to because one of my crates is used in some python binding and it can not bring tonic as a dependency but needs to bring some proto types.

This new API make it possible to have one crate with only the types definitions, then a second crate depending one the first one with only the tonic client (and doing so without altering the protobuf definitions).

Thank you for your consideration.

caspermeijn:
I understand your need. I feel you could do this with `extern_path`. The `services` crate can define `extern_path("testing.echo", "::messages")` and I expect it to generate the service definition without the messages. I have not tested this. Am I right?

PR #557, opened 2021-11-03
Change `decode_length_delimiter` to a mutable borrow

jaskij:
While it can make sense in some cases to use a clone here, let's not force users to do it since the function does *not need* it.


danburkert:
There is a blanket `impl Buf for &mut T where T: Buf` [provided in the bytes crate](https://docs.rs/bytes/1.1.0/bytes/trait.Buf.html#impl-Buf-for-%26mut%20T), so a clone is never necessary.  The API is set up to be maximally flexible in it's current format.

PR #560, opened 2021-11-03
Add test for split varint decoding

jaskij:
An additional test to force testing non-contguous chunks in `Buf`. This is necessary, since this edge case is, theoretically, supported by the `Buf` trait.

Additionally, this also serves to test the slow path decoding.

Although, as https://github.com/tokio-rs/bytes/issues/474 shows, this is probably a low-priority case.

### DIFF
--- a/src/encoding.rs
+++ b/src/encoding.rs
@@ -1776,4 +1776,27 @@ mod test {
         (String, string),
         (Vec<u8>, bytes)
     ]);
+
+
+    #[test]
+    fn split_varint_decoding() {
+        let mut test_values = Vec::<u64>::with_capacity(10 * 2);
+        test_values.push(128);
+        for i in 2..9 {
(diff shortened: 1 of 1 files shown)


src/encoding.rs
```diff
@@ -1776,4 +1776,33 @@ mod test {
         (String, string),
         (Vec<u8>, bytes)
     ]);
+
+
+    #[test]
+    fn split_varint_decoding() {
```
caspermeijn, line 1782:
```suggestion
    /// `decode_varint` accepts a `Buf`, which can be multiple concatinated buffers. 
    /// This test ensures that future optimizations don't break the 
    /// `decode_varint` for non-continuous memory.
    fn split_varint_decoding() {
```

caspermeijn (changes requested):
Looks good to me. Can you rebase to latest master?

Issue #575, opened 2021-12-22
Supply chain security with embedded protoc binaries


benesch:
We're thinking about using prost over at @MaterializeInc, but the supply chain security of the embedded `protoc` binaries has me very nervous.

Specifically: what sort of validation are you doing before you merge PRs to check that the binaries supplied by third parties are identical to the binaries published by google/protobuf? I did some spot checks and they look identical, but I didn't see anything in your CI pipeline that would assert this. 

We'd honestly probably want to opt out of the bundled protoc deps entirely. I know that's already possible if you specify `PROTOC`, but I'd feel better if there were a way to not even bring the binaries into our dependency tree. Otherwise it seems too easy for accidents to happen; e.g., if that env var somehow gets unset. 

How would y'all feel about adopting an approach similar to rust-protobuf, where the bundled binaries are kept in a separate crate (e.g., `prost-build-vendored`)? Using that crate could be enabled by default, but we could keep it off @MaterializeInc and lint for its absence using `cargo deny`, and I'd sleep a good bit easier at night.

I've also been toying with the idea of building something like `protobuf-src` for Rust (a la `openssl-src`) that builds libprotobuf from source and makes its protoc available to downstream libraries. That could be a nice alternative for `prost-build` to expose.

LucioFranco:
Hey @benesch sorry for the delay on this I must have missed it somehow. Are you still interested in getting this working? I think the separate crate idea is good. 

cc @jonhoo I think you've been interested in some of this stuff too.

jonhoo:
I think doing something along what's proposed here is a great idea!

benesch:
No problem, @LucioFranco! Thanks for taking a look now. I actually whipped up a proposal in https://github.com/tokio-rs/prost/pull/576, but we need you or another maintainer to weigh in on how to expose the toggle of which `protoc` to use. The lack of mutually exclusive features in Cargo means there’s no great solution.

I’m also maintaining a fork of prost that _forces_ compiling protoc from source: https://github.com/materializeinc/prost. But I figure you probably don’t want to force a C++ dependency on all users of prost.

benesch:
By the way, if it’s easiest to sync on Discord or something, happy to find the time!

LucioFranco:
Yeah, let me try to catch up on everything so may be a bit slow then we can sync. I'd love for yall to not have to maintain a fork so I'd like to get this solved :)

LucioFranco:
[...]

Issue #562, opened 2021-11-05
Security: avoid storing binary artifacts for `protoc` binary


tmfink:
Thanks for your great work with this crate.

# Situation

The `prost-build` crate seems to use cached, pre-compiled binaries for the `protoc` by default:
https://github.com/tokio-rs/prost/tree/master/prost-build/third-party/protobuf

Users can opt out of this behavior by setting the `PROTOC` environment variable:
https://github.com/tokio-rs/prost/blob/c3b7037a7f2c56cef327b41ca32a8c4e9ce5a41c/prost-build/build.rs#L102-L108

# Problem

It's surprising that this crate would use a pre-compiled binary artifact by default.
Those concerned with supply-chain security don't want to unwittingly rely on binary artifacts

In fact, in the software security space, the [Software Bill of Materials (SBOM)](https://www.ntia.gov/SBOM) movement has gained traction, where vendors show where all code comes from. In the United States, there was an [executive order](https://www.whitehouse.gov/briefing-room/presidential-actions/2021/05/12/executive-order-on-improving-the-nations-cybersecurity/) mandating that vendors providing software to the government provide an SBOM. It would not look good if a vendor unwittingly relied on a binary artifact.

The Open Source Security Foundation (OpenSSF) [classifies storing binary artifacts in the repo as a high-risk security concern](https://github.com/ossf/scorecard/blob/af594d39210f97029cd207e32b703939cf42b177/docs/checks.md#binary-artifacts). I don't want to re-iterate every point, but the high-level point is that while not impossible to review, binaries are much more difficult to validate compared to source code. This makes it easier for an attacker to insert malicious code.

For example, in a recent [breach of the JavaScript NPM package `ua-parser-js`](https://us-cert.cisa.gov/ncas/current-activity/2021/10/22/malware-discovered-popular-npm-package-ua-parser-js), attackers inserted malware into the package. This is especially bad since (as of writing) that package has about 6 million weekly downloads.

As of writing, crates.io shows that  [`prost` has 368 public reverse dependencies](https://crates.io/crates/prost/reverse_dependencies). That does not include unpublished/proprietary code or transitive reverse dependencies. There are probably many users of `prost` who don't know they are users of `prost` (due the `prost` existing somewhere in the dependency graph) :wink:.

# Solutions

Note that either of these is a backwards-incompatible change and would require a version bump.

## Minimum Fix

[...]

PR #576, opened 2021-12-22
proto-build: optionally unbundle protoc

benesch:
This is a proposed solution to #575. I expect there will be lots of opinions here, but I at least wanted to whip something up as a show of good faith! With something like this in place we'd be able to take the dependency on prost in [MaterializeInc/materialize](https://github.com/MaterializeInc/materialize).

----

This commit makes it possible to opt out of downloading the bundled
pre-compiled protoc binaries. This is useful for downstream users with
strict supply chain security requirements (see #575 for details).

The approach taken is to move the pre-compiled binaries to a new crate
called `protoc-bin`. prost-build depends on this crate only if the
`protoc-vendored-bin` feature is enabled (which it is by default).
While it was always possible to avoid the vendored binaries by
specifying the `PROTOC` environment variable, the feature is a bit more
foolproof; security-concious users can audit their Cargo.lock to ensure
that the protc-bin crate is never even downloaded.

This commit also adds a new feature called `protoc-vendored-src`, which
uses the new `protobuf-src` crate to compile a copy of protoc from
source on the fly. This feature makes `cargo build` seamless on all
platforms that Protobuf supports, not just those that protoc-bin
distributed pre-compiled binaries for, but has the downside of requiring
a C++ toolchain and increasing compile times.

If neither of the new vendored features are enabled, `prost-build` fails
unless `PROTOC` and `PROTOC_INCLUDE` are explicitly specified.


tmfink:
This PR is a good first step to solving the supply chain security concerns in #562 and #575!

I have some comments below.

## Environment variables instead of cargo features

Only the "root" of the dependency of graph (AKA the executable being built) should have control over how `protoc` is acquired. As such, controlling via environment variables is preferable to cargo features. 

Each crate `A` that depends on prost (even transitively) would need to "forward" cargo features. This would allow any crate `B` that depends on `A` to itself expose (and forward to `A`) the necessary "prost" features (and so on). This adds more work for each crate that includes prost in the dependency graph. 

Also, there is no good way to resolve conflicting features. Due to how cargo resolves [feature unification](https://doc.rust-lang.org/cargo/reference/features.html#feature-unification), the prost crate may be compiled with with the features to use source *and* binary `protoc`.

The end-user is the one who can make the best security decisions as to how `protoc` is acquired. There is no reason to allow the transitive dependencies to have influence on the decision. The best way I know to allow that is by controlling the functionality via environment variables.

## Fail closed

This would be a breaking change, but I vote that this feature "fails closed" (be secure-by-default).

For example:
1. If `PROST_BUILD_PROTOC_FROM_SRC` is set, then build from source
2. If `PROTOC` is set, use the binary.
3. If `protoc` is found on the system (maybe in `PATH`?) then use that
4. If `PROST_USE_BUNDLED_BINARIES` is set, then use the supplies binaries.
5. Otherwise, panic with a descriptive error message.

The openssl-sys crate does a good job of emitting a descriptive error message when OpenSSL cannot be found:
https://github.com/sfackler/rust-openssl/blob/b3dde76ea28ca5e54ce6341927c96679b9921153/openssl-sys/build/find_normal.rs#L98-L112

On a side note, I think it's reasonable to expect that a user has a C++ compiler available to compile protobuf from source. If the user is already compiling Rust, then they probably have a C++ compiler. A C compiler is essentially required to link Rust binaries (which is probably accompanied by a C++ compiler).

benesch:
> Only the "root" of the dependency of graph (AKA the executable being built) should have control over how `protoc` is acquired.

I agree wholeheartedly with this. 

Unfortunately it's a nonstarter for us to move to a model where environment variables are _required_ for a successful compile. We maintain the invariant for Materialize that a `cargo build` is all you need to compile (assuming you've got CMake, a C++ toolchain, and a Rust toolchain installed). We don't want our users and developers to have to type `PROST_BUILD_PROTOC_FROM_SRC=1 cargo build`.  It's a shame that Cargo doesn't allow root crates to specify env vars to set before invoking any build scripts, because that would neatly solve the problem.

> The end-user is the one who can make the best security decisions as to how `protoc` is acquired. There is no reason to allow the transitive dependencies to have influence on the decision. The best way I know to allow that is by controlling the functionality via environment variables.

I think using features is unavoidable, though, because otherwise you'll be stuck downloading both the binary blobs _and_ compiling protobuf from source, even if you know you only want one or the other (or neither). I.e., you can't depend on the `protobuf-src` crate only if an environment variable is set; you're always going to compile the thing from source so that it's available in case you need it. And then you end up with a dependency on a C++ compiler even if you were planning to use the binaries. 

I think it makes sense to expose environment variables as _additional_ layers of control, though, on top of the features that are selected. That could allow you to enable both the `protoc-vendored-bin` and `protoc-vendored-src` features (sidestepping Cargo's mutually-exclusive feature morass) and toggle which one you want via an environment variable.

I do wonder if the Prost maintainers and users would be happy enough with the default always being to compile protobuf from source. I agree that a C++ compiler is not terribly much to ask; indeed you've probably already got one if you've got a C compiler.

Somewhat relatedly, I'm working on [C++ bindings for libprotobuf](https://github.com/MaterializeInc/rust-protobuf-native/tree/master/protobuf-native/src) that will make it possible to "invoke" protoc by just calling into the appropriate C++ APIs. That could prove useful here.

tmfink:
> It's a shame that Cargo doesn't allow root crates to specify env vars to set before invoking any build scripts, because that would neatly solve the problem.

I think you can set both [cfg flags](https://doc.rust-lang.org/cargo/reference/build-scripts.html#rustc-cfg) and  [env vars](https://doc.rust-lang.org/cargo/reference/build-scripts.html#rustc-env) from build scripts.

> I think using features is unavoidable, though, because otherwise you'll be stuck downloading both the binary blobs _and_ compiling protobuf from source, even if you know you only want one or the other (or neither). I.e., you can't depend on the `protobuf-src` crate only if an environment variable is set; you're always going to compile the thing from source so that it's available in case the env var says you need it.

By setting cfg flags in a build script, you can at least avoid *compiling* `protobuf-src` (but not downloading the dependency).

> I do wonder if the Prost maintainers and users would be happy enough with the default always being to compile protobuf from source. I agree that a C++ compiler is not terribly much to ask; indeed you've probably already got one if you've got a C compiler.

I agree. If you are already in a mode of "compiling code" (AKA `prost`), then I think you should be okay compiling dependencies.

Could you update your PR to use `protobuf-src` by default? That way, there does not need to be a `PROST_BUILD_PROTOC_FROM_SRC` variable/feature. The only feature would be the non-default "protbuf-bin" feature.

benesch:
> By setting cfg flags in a build script, you can at least avoid compiling protobuf-src (but not downloading the dependency).

I don’t think it works like that. The cfg flags and env vars you set in a build script only apply to the one crate that is being built. 

> Could you update your PR to use protobuf-src by default? That way, there does not need to be a PROST_BUILD_PROTOC_FROM_SRC variable/feature. The only feature would be the non-default "protbuf-bin" feature.

I’m happy to do that, but I’d like @LucioFranco or another maintainer could weigh in on their desired approach before I take the time!

protoc-bin/README.md
```diff
@@ -0,0 +1,16 @@
...
+a Cargo build. See the crate [documentation](https://docs.rs/prost-build/) for examples
+of how to integrate `prost-build` into a Cargo project.
+
+## License
+
+`prost-build` is distributed under the terms of the Apache License (Version 2.0).
+
+See [LICENSE](../LICENSE) for details.
+
+Copyright 2017 Dan Burkert
```
LucioFranco, line 16: I think this whole file needs to be updated for  the `protoc-bin` crate?

protoc-bin/src/lib.rs
```diff
@@ -0,0 +1,59 @@
+//! `protoc-bin` vendors `protoc` binaries for common architectures.
+//!
+//! It is intended for use as a build dependency for  libraries like
```
LucioFranco, line 3:
```suggestion
//! It is intended for use as a build dependency for libraries like
```

protoc-bin/src/lib.rs
```diff
@@ -0,0 +1,59 @@
...
+//! [`prost-build`](https://docs.rs/prost_build).
+//!
+//! Binaries are currently provided for the following platforms:
+//!
+//!   * Linux x86
+//!   * Linux x86_64
+//!   * Linux aarch64
+//!   * macOS x86_64
+//!   * macOS aarch64
+//!   * Windows 32-bit
```
LucioFranco, line 13: No windows 64-bit?

LucioFranco (commented):
Sorry for the delay on reviewing this! I'd like to get something like this in the next breaking release. I left a few questions nothing blocking, I think this approach looks pretty good.

---

> >    Could you update your PR to use protobuf-src by default? That way, there does not need to be a PROST_BUILD_PROTOC_FROM_SRC variable/feature. The only feature would be the non-default "protbuf-bin" feature.

> I’m happy to do that, but I’d like @LucioFranco or another maintainer could weigh in on their desired approach before I take the time!

Honestly, on a pure user experience end I would _not_ want to default to compiling c++ code since this is usually quite a hit to the build experience. That said, in practice I am not sure if it would be that big of a deal. 

I'd love to hear more from people more integrated in the build process to chime in. 

cc @jonhoo

LucioFranco:
I wrote an update here https://github.com/tokio-rs/prost/issues/575#issuecomment-1078153519 I think for now we can close this PR and start with a new one. Thanks for taking the start on this @benesch.

Issue #553, opened 2021-10-25
Fix clippy warnings introduced with 2021 edition


LucioFranco:

chetaldrich:
So I looked into this just because I was curious, and noticed that there were a decent number of warnings already showing up without changing the edition. Would you be willing to take a look at a PR that addresses existing warnings?

chetaldrich:
I gave it a whirl. If you have a chance to take a look at some point I'd appreciate it!

chetaldrich:
That being said, we can fix clippy errors with 2021 in stable, we just can't set the library to 2021.

LucioFranco:
Yeah I am good to fix clippy errors, we can follow up on the edition update.

chetaldrich:
I went ahead and followed up with that edition update so we can close this out.

PR #581, opened 2022-01-07
fix clippy linting recommendations (edition 2018)

chetaldrich:
Figured I'd start with recommendations in the current codebase as is, and that'll make handling https://github.com/tokio-rs/prost/issues/553 a bit easier.

### DIFF
--- a/prost-build/src/code_generator.rs
+++ b/prost-build/src/code_generator.rs
@@ -263,8 +263,8 @@ impl<'a> CodeGenerator<'a> {
     fn append_type_attributes(&mut self, fq_message_name: &str) {
         assert_eq!(b'.', fq_message_name.as_bytes()[0]);
         for attribute in self.config.type_attributes.get(fq_message_name) {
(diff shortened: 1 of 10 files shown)


LucioFranco:
@chetaldrich looks like you may need to run rustfmt, ci is failing.

chetaldrich:
@LucioFranco I ran rustfmt, so should be good to give CI another run.

chetaldrich:
It turns out you get a compilation error on Rust 1.51.0 if you use the `split_once` and `rsplit_once` methods since they were unstable at that version.

In Rust stable (currently 1.58.1), there are clippy recommendations that
say to use those methods instead of `rsplitn` and `splitn` if you're only splitting once.

I figured the best way to handle this since we're targeting 1.51 in addition to stable is to just set specific ignore in that case so the old version still compiles successfully.

Should be good for another CI run @LucioFranco, thanks for being prompt about the last one!

PR #591, opened 2022-02-05
Support for custom options and proto2 extensions

### LINKED ISSUE #100: Are proto2 extensions supported?
https://developers.google.com/protocol-buffers/docs/proto#extensions

I have existing proto2 protocol definition which uses extensions. And I'm unable to understand how to work with such protocol using `prost`. Am I looking into wrong direction or this feature is not supported?

### LINKED ISSUE #256: Custom options?

nswarm:
Hi, this PR adds support for [extensions](https://developers.google.com/protocol-buffers/docs/proto#extensions) (proto2 only) and [custom options](https://developers.google.com/protocol-buffers/docs/proto#customoptions) (proto2 and 3).

Custom options are built on top of extensions, so I'll refer to them collectively as "extensions".

Please see the included documentation and tests for examples of how `prost` users would use these features.

## Overview

This code is roughly patterned after how extensions are handled in other languages generated by protoc itself.

Here's a whiteboard which hopefully helps explain the flow a bit better. This is from when I was planning so the names don't perfectly line up to the code.

protobuf-extensions-discovery

The core functionality lives in `src/extension.rs`:
- Extendable
	- Applied to proto messages marked with `extensions n to m;` via code generation.
	- Provides functions for working with the underlying ExtensionSet.
- ExtensionSet
	- Contains a set of decoded values in a type-opaque manner through ExtensionValue.
	- Its methods take `ExtensionImpl<T>` which can be used to manage the data of an ExtensionValue by downcasting it to an ExtensionValueImpl<T>. 
- Extension/ExtensionImpl
	- const definitions are generated in rust code from the proto `extend MessageName {...}` defintions.
	- These definitions are used with the Extendable/ExtensionSet to provide type information.
- ExtensionRegistry
	- This is a temporary structure that you load the generated const `Extensions` into before parsing.
	- It provides the information to the parser necessary to decode values into the type-erased `ExtensionSets`

The other major file `src/generic.rs` provides traits `Merge`, `MergeRepeated`, `Encode`, and `EncodeRepeated`, which exist exclusively to map rust types to the various functions in `prost::encoding`. This allows merging and encoding to work for the `ExtensionSet` fields that won't have prost derive attributes for each of field it needs to decode.

The remaining changes are mostly implementing the above traits, tests, documentation.

## Notes
- This PR makes additive changes to `prost_types::protobuf.rs` which adds to all `*Options` types:
	- derive ::prost::Extendable.
	- a `pub extension_set` field. :warning: This will break for anyone creating these types directly through e.g. `FileOptions{...}`.
	- an impl that defines a const EXTENDABLE_TYPE_ID.
- This PR has almost zero runtime impact on anyone not using extensions.
	- The only effect is the various `*Options` take an additional Option<Box<_>> bytes.
- I ran into a build issue on which seems to have been encountered by others (#556). That is the change to `prost-build/Cargo.toml`.
- I am happy to provide maintenance for this area of code in the future.

A nice side effect of this change is prost could define its own custom options for use in compile_protos e.g. the ability to tag specific fields to generate using small_vec.

### Related issues

Resolves #256 
Resolves #100


README.md
```diff
@@ -343,6 +343,116 @@ during the build & code generation step. When used in conjunction with the `std:
...
+
+```protobuf
+import "google/protobuf/descriptor.proto";
+extend google.protobuf.FileOptions {
+  int32 my_opt = 50000;
+  repeated string my_repeated_opt = 50001;
+}
+```
+
+You need to compile these definitions alone before your other protos. The above will generate const `Extension` fields `MY_OPT` and `MY_OPT_REPEATED` that you'll need to register in an `ExtensionRegistry` as well as request data from an `Extendable` message.
```
neoeinstein, line 375: Mismatch in code and description?

```suggestion
You need to compile these definitions alone before your other protos. The above will generate const `Extension` fields `MY_OPT` and `MY_REPEATED_OPT` that you'll need to register in an `ExtensionRegistry` as well as request data from an `Extendable` message.
```

README.md
```diff
@@ -343,6 +343,116 @@ during the build & code generation step. When used in conjunction with the `std:
...
+```rust
+// The generated code of `Extensions` declared previously.
+mod custom_options {
+    include!("path/to/my/custom_options.rs");
+}
+
+fn generate() {
+    let mut extension_registry = ExtensionRegistry::new();
+    extension_registry.register(custom_options::MY_OPT); // const Extension definitions.
+    extension_registry.register(custom_options::MY_OPT_REPEATED);
```
neoeinstein, line 400:
```suggestion
    extension_registry.register(custom_options::MY_REPEATED_OPT);
```

README.md
```diff
@@ -343,6 +343,116 @@ during the build & code generation step. When used in conjunction with the `std:
...
+    let descriptor_set: FileDescriptorSet = Message::decode_with_extensions(DESCRIPTOR_SET_BYTES, extension_registry);
+    for file in descriptor_set.file {
+        println!("file {} my_opt: {:?}", file.name.unwrap(),
+                 // Returns an Option<i32> (based on our Extension declared above).
+                 file.options.extension_data(&custom_options::MY_OPT));
+        println!("file {} my_opt: {:?}", file.name.unwrap(),
+                 // Returns an Option<Vec<String>> (based on our Extension declared above).
+                 file.options.extension_data(&custom_options::MY_OPT_REPEATED));
```
neoeinstein, line 408:
```suggestion
                 file.options.extension_data(&custom_options::MY_REPEATED_OPT));
```

neoeinstein (commented):
Minor documentation changes that looked to be mismatched.

Would be interested to see how this could be used to, for instance, add an attribute (such as a proc-macro) to a message or a field to modify the code generation behavior.

nswarm:
Thanks @neoeinstein for some great usability changes!

nswarm:
> neoeinstein

Operating inside the prost lib, we could define prost-level custom options that could be explicitly loaded and referenced during code generation and add the relevant rust attribute. That could even be a user defined macro, for instance we could directly inject the name into the attr list:
```
option (prost_proc_attr) = "::mycrate::MyAttr";
```

This would be akin to the protoc plugin [insertion points](https://developers.google.com/protocol-buffers/docs/reference/cpp-generated#plugins) in google's generators, although wouldn't scale very well for code blocks.

Is that the kind of thing you mean?

neoeinstein:
That's part of it. Inside prost, I'm thinking about things like indicating a message or field should be a foreign type, providing specified names, etc. Below is a series of potential things off the top of my head. Not all of these would need to be implemented; some will be more useful than others.

```
// This one would be more complex due to impacts to type resolution
// in generated code. It's also a bit less useful due to the way we use `include!`.
option (prost.package).path = "::my_crate::placement";

message Example {
    option (prost.type).type_name = "ExampleRust";
    option (prost.type).attrs = "#[::aliri_braid::braid]";

    uint64 id = 1 [(prost.field).field_name = "identifier"];
}

message External {
    option (prost.type).external = "::my_external::Type";

    sint32 value = 1;
}

message MessageWithUuid {
    string uuid = 1 [(prost.field).type = "::uuid::Uuid"];
}

message OwnedAndBorrowed {
    option (prost.type).gen_borrowed = true;

    bytes data = 1 [
        (prost.field).type = "::bytes::Bytes",
        (prost.field).borrowed_type = "[u8]"
    ];
    string my_typed_string = 2 [
        (prost.field).type = "::my_crate::MyStronglyTypedString",
        (prost.field).borrowed_type = "::my_crate::MyStronglyTypedStringRef"
    ];
}
```

A borrowed mode would make it possible to reduce unnecessary copies, as you could deserialize an owned buffer into values referring to that data or construct a message for serialization without needing to take ownership of all of the parts.

neoeinstein:
I posted up a _really_ quick example for the renaming of a field option I mentioned above. Should probably move this discussion to a separate issue or such, but it all builds on top of this PR right now. Renaming of types and packages likely requires a two-pass process so that we can collect all of the aliases _first_ and then use the correct aliases and internally-defined externs when doing code generation pass.

https://github.com/neoeinstein/prost/compare/extensions-usability...neoeinstein:prost-ext

prost-build/src/code_generator.rs
```diff
@@ -403,7 +403,7 @@ impl<'a> CodeGenerator<'a> {
         self.push_indent();
         self.buf.push_str("pub const ");
         self.buf.push_str(&extension.name().to_ascii_uppercase());
-        self.buf.push_str(": &::prost::ExtensionImpl<");
+        self.buf.push_str(": &'static ::prost::ExtensionImpl<");
```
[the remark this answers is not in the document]
nswarm: > I actually did this initially, and then removed it based on a `clippy` recomendation.

Really? It wasn't even compiling for me with errors like:

```
error[E0106]: missing lifetime specifier
   --> <snip>\prost\target\debug\build\protobuf-56feaaf2b10845f1\out/protobuf_unittest.rs:893:40
    |
893 | ...   pub const TEST_EXT_ORDERINGS3: &::prost::ExtensionImpl<TestExtensionOrderings3> = &::prost::ExtensionImpl::<TestExtensionOrderings3...
    |                                      ^ expected named lifetime parameter
    ```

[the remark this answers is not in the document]
Comment (neoeinstein) [reply prost-build/src/code_generator.rs:406]: This sounds like a compiler version issue. What version of the rust compiler are you using locally?

This is the specific [clippy lint regarding redundant static lifetimes](https://rust-lang.github.io/rust-clippy/master/index.html#redundant_static_lifetimes). According to [this comment](https://github.com/rust-lang/rust-clippy/issues/6097#issuecomment-731080791), it expects a Rust compiler to be version 1.17 or later, which is almost 5 years old now.

The Prost minimum supported version of the Rust compiler right now is, I belive, 1.51.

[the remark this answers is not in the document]
Comment (nswarm) [reply prost-build/src/code_generator.rs:406]: I've built/tested on the following compilers using `cargo test --workspace`:

1.51.0-x86_64-unknown-linux-gnu
stable-x86_64-unknown-linux-gnu (1.58.1)

1.51.0-x86_64-pc-windows-msvc
stable-x86_64-pc-windows-msvc (1.58.1)

And I get the errors on all of them. Although I did find another unsupported 1.51 issue...

I do notice if I only use `cargo test` I won't get the errors. By chance are you not using --workspace?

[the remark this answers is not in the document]
Comment (neoeinstein) [reply prost-build/src/code_generator.rs:406]: Ooooh. After using `cargo test --all`, I see where it is. These are the nested extensions that end up inside `impl` blocks. For the module-level consts, the static modifier is redundant, but inside of an `impl` block they aren't considered redundant. I think that makes it possible for us to split the difference, either adding `'static` always and throwing in the `allow`, or only adding in `'static` when we're dealing with a nested extension.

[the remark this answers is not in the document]
Comment (neoeinstein) [reply prost-build/src/code_generator.rs:406]: Threw you one more PR. It adds the above mentioned clippy lint as a kind of least-common-denominator. Beyond that, I found some errors when building on no-std, so I added the appropriate `use` references to the `alloc` crate to make sure that path compiles as well.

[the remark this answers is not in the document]
Comment (nswarm) [reply prost-build/src/code_generator.rs:408]: Awesome, thanks for digging into that!

**Comment by vriesk** [comment]
Hi! Is this change abandoned now? It is somewhat similar to my feature request here: https://github.com/tokio-rs/prost/issues/658

I'd be particularly eager to see a way to inject support for PGV, preferably directly into `validator` tags.

**Comment by neoeinstein** [comment]
> Hi! Is this change abandoned now? It is somewhat similar to my feature request here: https://github.com/tokio-rs/prost/issues/658
> 
> 
> 
> I'd be particularly eager to see a way to inject support for PGV, preferably directly into `validator` tags.

This change is not abandoned, and indeed is planned on the way to Prost 1.0. However, we may not land this in the exact form provided in the PR for general public use. There is an underlying need to support unrecognized tags that this should be built upon, and that would eliminate the need to do special work up-front with a registry in order to get to these extensions.

Nonetheless, there are several, including myself, who are using this PR branch as a git dependency and a bridge for doing supplemental codegen. I have some plans to create a generator similar to PGV in my line of `protoc-gen-prost` crates.

The key thing is that, we want to have an API where you can just use `decode` rather than `decode_with_registry`.

It may be that we _could_ land this for an intermediate release and then have a further breaking change that removes the need for a registry, but if we start publishing a release that includes support for it, we'd need to make sure the functionality isn't broken even across breaking API changes.

**Comment by vriesk** [comment]
Thanks for the update. Yeah, I was even thinking about forking `prost-build` privately for now, however the problem I see is that it is non-trivial to integrate it with `tonic` for the full service generation (but I have not tried hard enough TBF).  Or is there?

**Comment by nswarm** [comment]
> There is an underlying need to support unrecognized tags that this should be built upon, and that would eliminate the need to do special work up-front with a registry in order to get to these extensions.
> (snip)
> The key thing is that, we want to have an API where you can just use `decode` rather than `decode_with_registry`.

I'm not sure that is possible without a registry or some form of additional information up front.

Extensions are fundamentally data that cannot be parsed by the default parser, so the user has to provide additional information i.e. the registry which tells it how to parse that data. Similarly, unregonized tags / uninterpreted options are just serialized data the parser was not provided a schema for handling.

In either case you still need a way to parse it that's not available by default. If you don't provide a registry or schema up front, you get a list of "sorry couldn't parse this" data that you have to manually parse yourself.

In the case of extensions, the registry step could be eliminated by having a static, global registry that automatically contains all extensions in generated code at compile time, that the default parser would know to use. This is how [google's c++ does it](https://groups.google.com/g/protobuf/c/BwhsxPjPplE/m/gSwtjUamOZ4J). That could be additive on top of this PR and even feature gated for anyone who wants to specify registry manually. Other implementations like [java](https://developers.google.com/protocol-buffers/docs/reference/java/com/google/protobuf/ExtensionRegistry) and [c#](https://cloud.google.com/dotnet/docs/reference/Google.Protobuf/latest/Google.Protobuf.ExtensionRegistry) use the extension registry pattern as well.

Unrecognized tags are a different beast because it means you have a different or out of date schema from what the data was serialized with. I think it's a subtle but important difference from extensions.

For that reason I _think_ support for unrecognized tags is a separate code path from extensions, but fwiw I have not thought deeply on that.

**Comment by LucioFranco** [comment]
@nswarm I would be interested in landing some of your work but because I am not the original author of this crate we are gonna need go through a proper proposal process w/ feedback etc in a github issue. Then I think we can start to land smaller PRs that are easier to review etc. Sorry for the major delay on reviewing I've had a huge backlog to get through. Thanks for your patience.

**Comment by nswarm** [comment]
> @nswarm I would be interested in landing some of your work but because I am not the original author of this crate we are gonna need go through a proper proposal process w/ feedback etc in a github issue. Then I think we can start to land smaller PRs that are easier to review etc. Sorry for the major delay on reviewing I've had a huge backlog to get through. Thanks for your patience.

That's fine. Let me know how you'd like me to proceed.

I'll need to spend some time figuring out how to break it up as well.

File: tests/src/extensions.proto [hunk 5]
```diff
@@ -0,0 +1,49 @@
...
+  repeated sint32 ext_sint32_repeated = 122;
+  repeated sint64 ext_sint64_repeated = 123;
+  repeated fixed32 ext_fixed32_repeated = 124;
+  repeated fixed64 ext_fixed64_repeated = 125;
+  repeated sfixed32 ext_sfixed32_repeated = 126;
+  repeated sfixed64 ext_sfixed64_repeated = 127;
+  repeated bool ext_bool_repeated = 128;
+  repeated string ext_string_repeated = 129;
+  repeated bytes ext_bytes_repeated = 130;
+  repeated CustomMessageType ext_message_repeated = 131;
```
andrewhickman, line 48: Its a bit of an edge case, but it would be nice to test a `group` field here as well

src/extension.rs
```diff
@@ -0,0 +1,881 @@
...
+use core::option::Option;
+
+type ExtendableTypeId = &'static str;
+type FieldTag = u32;
+
+/// Marks a message as having able to have extensions.
+/// Extension data must be retrieved and modified through use of static Extensions in generated code.
+pub trait Extendable: 'static {
+    /// A static type id associated with this Extendable.
+    fn extendable_type_id() -> ExtendableTypeId;
```
andrewhickman, line 21:
```suggestion
    fn extendable_type_id(&self) -> &'a str;
```
Could this return a borrowed string to allow implementing it for something like [`DynamicMessage`](https://docs.rs/prost-reflect/0.8.1/prost_reflect/struct.DynamicMessage.html)? Most consumers can just call `T::default().extendable_type_id()`

src/extension.rs
```diff
@@ -0,0 +1,881 @@
...
+
+    fn proto_int_type(&self) -> ProtoIntType {
+        self.proto_int_type
+    }
+}
+
+/// A generic container for an Extension that can delegate to the implementation for type-specific work.
+pub trait Extension {
+    /// Fully-qualified type name of the message this extension is for.
+    fn extendable_type_id(&self) -> ExtendableTypeId;
```
andrewhickman, line 523: This also could return a borrowed string to allow implementing for something like [`ExtensionDescriptor`](https://docs.rs/prost-reflect/0.8.1/prost_reflect/struct.ExtensionDescriptor.html)

src/extension.rs
```diff
@@ -0,0 +1,881 @@
...
+/// Extension data must be retrieved and modified through use of static Extensions in generated code.
+pub trait Extendable: 'static {
+    /// A static type id associated with this Extendable.
+    fn extendable_type_id() -> ExtendableTypeId;
+
+    /// The data structure that stores extension data.
+    fn extension_set(&self) -> &ExtensionSet<Self>;
+
+    /// The data structure that stores extension data.
+    fn extension_set_mut(&mut self) -> &mut ExtensionSet<Self>;
```
andrewhickman, line 27: These two methods basically require the implementation to have an `ExtensionSet` field, which feels unnecessarily restrictive. For example, another valid implementation would be to store unknown fields at decode time (without looking at `ExtensionRegistry` and then decode on-the-fly in methods like `extension_data`

nswarm:
Thanks for the comments @andrewhickman! Note that I probably won't do much work on this for a bit, per status in: https://github.com/tokio-rs/prost/issues/674

README.md
```diff
@@ -357,6 +357,116 @@ during the build & code generation step. When used in conjunction with the `std:
...
+// Using our options declared above.
+option (my_opt) = 1234;
+option (my_repeated_opt) = "hello";
+option (my_repeated_opt) = "goodbye";
+```
+Note that the repeated option is specified multiple times; this is the syntax for specifying multiple values that will be parsed into a `Vec<_>`.
+
+### Reading Custom Options from Descriptors
+
+```rust
```
slinkydeveloper, line 405:
```suggestion
```rust,ignore
```

To disable testing it

README.md
```diff
@@ -357,6 +357,116 @@ during the build & code generation step. When used in conjunction with the `std:
...
+// Using the `extend` keyword will define const `Extensions` in the generated code.
+extend Foo {
+  // This will generate a `pub const EXT_INTEGER: Extension...;` declaration.
+  optional int32 ext_integer = 100;
+}
+```
+
+When working with extensions on a decoded object, you use a reference to the generated `Extension` as a sort of handle.
+
+```rust
```
slinkydeveloper, line 453:
```suggestion
```rust,ignore
```

To disable testing it

slinkydeveloper (commented):
Hi all,
I need this feature for a prototype that I'm building that manipulates `FileDescriptorSet` with options, I'll be more than happy to provide a thorough review and feedback in the next days. 

If it can be any helpful, here is my branch rebased with master: https://github.com/slinkydeveloper/prost/tree/extensions

cgorski:
@nswarm just wanted to let you know I use this PR in a private project with a few code additions. We're replacing a legacy app using old proto2 definitions and it would not be possible to do a drop-in replacement in any reasonable way without this PR.

Houndie:
I know this PR is kind of in limbo right now, and while I'm not volunteering to push it through to the end, I did go and make a fork of it and merge master into it. From reading the comments, it seems like a number of people are using this PR branch, so having an updated one might be helpful

https://github.com/Houndie/prost/tree/extensions

(I also fixed a small bug where `Result` wasn't properly scoped in macro generation)

ozars:
In the same vein, I rebased @Houndie's branch on current master at https://github.com/ozars/prost/tree/extensions, fixing a small issue due to failed doctests in readme file.

jhugard:
I added a few commits to ozars' fork above & merged in master to pick up 0.12.1

https://github.com/jhugard/prost/tree/extensions

* Added `extension_registry` to ServiceGenerator Config & builder, so that that extensions can be used in the `service_generator`.
* Commented-out generation of file-level `register_extensions`, because it creates multiple conflicting instances when extensions are defined in multiple proto files using the same namespace.

jhugard:
As an aside, I would be more in favor of a reflection-based approach applied at runtime.  Feels very clunky to have to bootstrap extensions by copying a generated file into my project, so that the extension definitions are available for use with `prost_build` in `build.rs`.

Kampfkarren:
Should be noted that none of the forks here I've tried have seemed to work with ServiceGenerator, I still have no idea how to extract custom options from those :(

pentatonick:
Please don't let this feature die. It is very interesting for creating protoc plugins in Rust, where we can annotate the messages with Meta data.

Please maintain the same workflow of dealing with options and extensions as C++ and Python for lowering the entry level to the feature as it was proposed by the @nswarm in the original CR.

nswarm:
> It's been 3 years since this pull request was opened, so we can assume this was abandoned?

See  https://github.com/tokio-rs/prost/issues/674 .

I consider this in the hands of maintainers since the last comms were to hold off and await greater design. I personally no longer need the feature as I changed jobs so it's unlikely I'll do more work here as well.

PR #597, opened 2022-02-14
Fix out-of-bounds array index

andrewhickman:
Doesn't make much difference since we're panicking either way, but at least we get a nicer error message


LucioFranco:
@andrewhickman could you by chance provide what the improved error looks like?

andrewhickman:
The error looks like `invalid c-escaped default binary value (\x1): incomplete hex value`

LucioFranco (approved):

PR #598, opened 2022-02-24
Expose generation methods for buf

neoeinstein:
This revives PR #313. With the advent of v1.0 of [buf.build](https://buf.build), the ecosystem for generating polyglot code from protobuf heavily favors the plugin model. It very easily slots into the `buf generate` model. This revival creates a distinct `protoc-gen-prost` crate to host the binary, rather than the prior method of putting the binary in the `prost-build` crate. This makes installation using `cargo install protoc-gen-prost` more convenient.

In my opinion, Prost would benefit from integrating into this model, so I wanted to revive the old protoc plugin. Even with this very basic PR (which doesn't handle options), I have confirmed that puting the binary onto the `PATH` and running `buf generate` results in precisely the Rust code I would expect over a somewhat complicated hierarchy of protobuf files.

### DIFF
--- a/prost-build/src/lib.rs
+++ b/prost-build/src/lib.rs
@@ -847,25 +847,32 @@ impl Config {
             )
         })?;
 
-        let requests: Vec<_> = file_descriptor_set.file.into_iter().map(|descriptor| {
-            (self.module(&descriptor), descriptor)
-        }).collect();
+        let requests: Vec<_> = file_descriptor_set
+            .file
+            .into_iter()
+            .map(|descriptor| (self.module(&descriptor), descriptor))
+            .collect();
 
-        let file_names: HashMap<Module, String> = requests.iter().map(|req| {
-            let mut file_name = if req.0.is_empty() {
-                self.default_package_filename.clone()
-            } else {
-                req.0.join(".")
-            };
+        let file_names: HashMap<Module, String> = requests
+            .iter()
+            .map(|req| {
+                let mut file_name = if req.0.is_empty() {
+                    self.default_package_filename.clone()
+                } else {
+                    req.0.join(".")
+                };
 
-            file_name.push_str(".rs");
+                file_name.push_str(".rs");
 
-            (req.0.clone(), file_name)
-        }).collect();
+                (req.0.clone(), file_name)
+            })
+            .collect();
 
         let modules = self.generate(requests)?;
         for (module, content) in &modules {
-            let file_name = file_names.get(module).expect("every module should have a filename");
+            let file_name = file_names
+                .get(module)
+                .expect("every module should have a filename");
             let output_path = target.join(file_name);
 
             let previous_content = fs::read(&output_path);
@@ -964,7 +971,10 @@ impl Config {
     /// This is generally used when control over the output should not be managed by Prost,
     /// such as in a flow for a `protoc` code generating plugin. When compiling as part of a
     /// `build.rs` file, instead use [`compile_protos()`].
-    pub fn generate(&mut self, requests: Vec<(Module, FileDescriptorProto)>) -> Result<HashMap<Module, String>> {
+    pub fn generate(
+        &mut self,
+        requests: Vec<(Module, FileDescriptorProto)>,
+    ) -> Result<HashMap<Module, String>> {
         let mut modules = HashMap::new();
         let mut packages = HashMap::new();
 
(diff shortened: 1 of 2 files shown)


neoeinstein:
This has now been updated to an even smaller, minimum viable set of changes to support an external plugin executable. I've create that executable now at https://github.com/neoeinstein/protoc-gen-prost. Presuming this looks good, I'll be in a position to push up a `protoc` plugin crate binary.

prost-build/src/lib.rs
```diff
@@ -949,25 +965,32 @@ impl Config {
...
-    fn generate(&mut self, files: Vec<FileDescriptorProto>) -> Result<HashMap<Module, String>> {
+    /// Processes a set of modules and file descriptors, returning a map of modules to generated
+    /// code contents.
+    ///
+    /// This is generally used when control over the output should not be managed by Prost,
+    /// such as in a flow for a `protoc` code generating plugin. When compiling as part of a
+    /// `build.rs` file, instead use [`compile_protos()`].
+    pub fn generate(
+        &mut self,
+        requests: Vec<(Module, FileDescriptorProto)>,
```
LucioFranco, line 976: module isn't a public type def + seems a bit awkward to expose a `Module` that is just `Vec<String>`. What are your thoughts on making this a struct with limited methods?

LucioFranco (changes requested):
This is a good start! Left a question but I think we are moving in the right direction.

prost-build/src/lib.rs
LucioFranco: Yeah I think its fine to keep it in lib.rs

neoeinstein: Added the `Module` type.

prost-build/src/lib.rs
```diff
@@ -1031,6 +1045,68 @@ impl fmt::Debug for Config {
...
 
+/// A Rust module path for a Protobuf package.
+#[derive(Clone, PartialEq, Eq, Hash)]
+pub struct Module {
+    components: Vec<String>,
+}
+
+impl Module {
+    /// Construct a module path from an iterator of parts.
+    pub fn from_parts<I: IntoIterator<Item = S>, S: Into<String>>(parts: I) -> Self {
```
LucioFranco, line 1056: can we move these bounds to a where clause?

prost-build/src/lib.rs
```diff
@@ -949,25 +964,32 @@ impl Config {
...
-    fn generate(&mut self, files: Vec<FileDescriptorProto>) -> Result<HashMap<Module, String>> {
+    /// Processes a set of modules and file descriptors, returning a map of modules to generated
+    /// code contents.
+    ///
+    /// This is generally used when control over the output should not be managed by Prost,
+    /// such as in a flow for a `protoc` code generating plugin. When compiling as part of a
+    /// `build.rs` file, instead use [`compile_protos()`].
+    pub fn generate(
+        &mut self,
+        requests: Vec<(Module, FileDescriptorProto)>,
```
LucioFranco, line 975: I think this should accept some generic `impl Iterator` since we don't really care if its a vec or not. What do you think?

LucioFranco (changes requested):
Almost there! Just two more comments then I think we are good to merge.

prost-build/src/lib.rs
neoeinstein: The issue here is that we need to build up the message graph before we can process each module. That requires stepping through `requests` twice. We can accept an `Iterator` in, but we'd need to stuff it into an intermediate that we can iterate through a second time. Based on my experimentation on the Rust Playground, in release mode, if the input `Iterator` is a `Vec` and it is collected it inside generate, the collection incurs no cost.

We could instead use an iterator that just borrowed the above values (`Item=&(Module, FileDescriptorProto)` or `Item=(&Module, &FileDescriptorProto)`). That would change the return type to `HashMap<&Module, &str>`, but doing so would require cloning the `FileDescriptorProto` (a relatively expensive operation) as `CodeGenerator` wants to consume an owned value (that may be worth looking into changing in the future).

LucioFranco: Oh that seems like a super valid point, lets just keep the Vec then! Thanks for the thorough explanation!

neoeinstein: Sounds good. I'll mark this as resolved as well.

neoeinstein:
I believe that this is ready for a new review.

LucioFranco (approved):

PR #549, opened 2021-10-19
Fine-grained selection of where to add type or field attributes using filters

marjakm:
This does not break the external API, it only adds 2 new functions

```rust

pub fn type_attribute_with_filter<P, A, F>(
    &mut self,
    path: P,
    attribute: A,
    filter: F,
) -> &mut Self;
where
    P: AsRef<str>,
    A: AsRef<str>,
    F: Into<TypeFilter>
    
pub fn field_attribute_with_filter<P, A, TF, LF, FF>(
    &mut self,
    path: P,
    attribute: A,
    type_filter: TF,
    label_filter: LF,
    field_filter: FF,
) -> &mut Self;
where
    P: AsRef<str>,
    A: AsRef<str>,
    TF: Into<TypeFilter>,
    LF: Into<LabelFilter>,
    FF: Into<FieldFilter>
```

and 6 types - 3 pairs of Filter and Selector:
```
TypeFilter, TypeSelector,
LabelFilter, LabelSelector, 
FieldFilter, FieldSelector, 
```
Filters are bitsets (newtypes around u32) and Selectors give meaning to values in that bitset (c-like enums with selected discriminator values). Attribute maps contain these filters and attributes are only applied if all filters match.

```rust
    type_attributes: PathMap<(String, TypeFilter)>,
    field_attributes: PathMap<(String, TypeFilter, LabelFilter, FieldFilter)>,
```
This allows very specific selection of where to add attributes, example:

```rust
        .type_attribute(".", "#[derive(serde::Serialize, serde::Deserialize)]")
        // enums with data should use serde adjacently tagged format
        .type_attribute_with_filter(
            ".",
            "#[serde(tag = \"kind\", content = \"data\")]",
            TypeSelector::RustEnumWithData
        )
        // All c-like enums should have clap::ArgEnum impl
        .type_attribute_with_filter(
            ".",
            "#[derive(clap::ArgEnum)]",
            TypeSelector::RustEnumCLike
        )
        // and structs a clap::Parser impl
        .type_attribute_with_filter(
            ".",
            "#[derive(clap::Parser)]",
            TypeSelector::RustStruct
        )
        // use a trait to find parsers for non-scalar types
        .field_attribute_with_filter(
            ".",
            "#[clap(parse(try_from_str=crate::CustomClapParse::parse))]",
            TypeSelector::RustStruct,
            LabelSelector::Everything,
            !FieldSelector::ProtobufScalar | FieldSelector::Bytes,
        )
        // repeated String also needs the custom parser
[...]

PR #600, opened 2022-02-26
Add a configuration hook to run after protobuf compilation

andrewhickman:
This is intended as a simpler alternative to PRs like #494 and #549. Instead of adding additional ways to filter types, we can just expose the file descriptor set to the user and give them the power to do the filtering.

By allowing the user to inspect the file descriptor set themselves, its possible to do stuff like adding attributes to messages but not enums and oneofs

For example, with [`prost-reflect`](https://crates.io/crates/prost-reflect) it can be used to annotate the messages with the message name to add reflection support

```
config
    .post_compile_hook(|config, file_set| {
        for file in &file_set.file {
            for message in &file.message_type {
                config.type_attribute(
                    &format!(".{}.{}", file.package(), message.name()),
                    &format!("#[derive(::prost_reflect::ReflectMessage)] #[prost_reflect(file_descriptor = \"TEST_FILE_DESCRIPTOR\", message_name = \"test.{}\")]", message.name()),
                );
            }
        }
    })
```


LucioFranco:
Why does this need to be a hook on config and not a free fn that you can call then allow you to call the methods on config?

andrewhickman:
I'm not sure what you mean by a free fn, could you give an example of the api you had in mind?

LucioFranco:
```rust
let file_set = prost_build::get_file_descriptor(...path config);

for file in &file_set.file {
            for message in &file.message_type {
                config.type_attribute(
                    &format!(".{}.{}", file.package(), message.name()),
                    &format!("#[derive(::prost_reflect::ReflectMessage)] #[prost_reflect(file_descriptor = \"TEST_FILE_DESCRIPTOR\", message_name = \"test.{}\")]", message.name()),
                );
            }
        }
```

Something like this maybe? and we can avoid the hook config

andrewhickman:
I think that works too, though there are a couple slight annoyances:
- The user has to provide the protos/include paths twice
- Unless the user calls `skip_protoc_run` on the config, we'd be running `protoc` twice

The main reason for the hook-based approach was so that as a library author, I can just provide an extension method like
`add_reflect_message_impls(config: prost_build::Config)` without the user needing to make any other changes to how they use `prost_build`

LucioFranco:
Yeah, I am just hesitant to adding these sort of "hooks" when I don't think they are 100% needed. I would imagine maybe going a non extension trait way might be better in this case and provide a custom config that kinda handles this. I would really to reduce the verbosity of the config struct as I think its gotten a bit out of hand and am hesitant to adding more to it. I don't think running protoc twice is that bad and as you said its still possible to disable it.

PR #601, opened 2022-03-03
fix edition 2021 clippy errors

chetaldrich:
This handles clippy errors that showed up when switching over
to edition 2021. Note that we can't move to use edition 2021 until
the MSRV is set to 1.56. See https://github.com/tokio-rs/prost/issues/553 for
more details.

### DIFF
--- a/prost-build/src/path.rs
+++ b/prost-build/src/path.rs
@@ -118,6 +118,7 @@ fn sub_path_iter(full_path: &str) -> impl Iterator<Item = &str> {
 /// Example: prefixes(".a.b.c.d") -> [".a.b.c", ".a.b", ".a"]
 fn prefixes(fq_path: &str) -> impl Iterator<Item = &str> {
     std::iter::successors(Some(fq_path), |path| {
+        #[allow(unknown_lints)]
         #[allow(clippy::manual_split_once)]
         path.rsplitn(2, '.').nth(1).filter(|path| !path.is_empty())
     })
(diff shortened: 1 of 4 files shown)


prost-build/src/path.rs
```diff
@@ -118,6 +118,7 @@ fn sub_path_iter(full_path: &str) -> impl Iterator<Item = &str> {
 /// Example: prefixes(".a.b.c.d") -> [".a.b.c", ".a.b", ".a"]
 fn prefixes(fq_path: &str) -> impl Iterator<Item = &str> {
     std::iter::successors(Some(fq_path), |path| {
+        #[allow(unknown_lints)]
         #[allow(clippy::manual_split_once)]
```
olix0r, line 122:
```suggestion
        #[allow(unknown_lints, clippy::manual_split_once)]
```

prost-build/src/path.rs
```diff
@@ -130,6 +131,7 @@ fn prefixes(fq_path: &str) -> impl Iterator<Item = &str> {
 /// Example: suffixes(".a.b.c.d") -> ["a.b.c.d", "b.c.d", "c.d", "d"]
 fn suffixes(fq_path: &str) -> impl Iterator<Item = &str> {
     std::iter::successors(Some(fq_path), |path| {
+        #[allow(unknown_lints)]
         #[allow(clippy::manual_split_once)]
```
olix0r, line 135:
```suggestion
        #[allow(unknown_lints, clippy::manual_split_once)]
```

olix0r (commented):
tioli

prost-build/src/path.rs
chetaldrich: Oh, clever. Thanks for the suggestion!

LucioFranco (approved):

PR #603, opened 2022-03-09
upgrade regex crate to avoid security issue

tisonkun:
```
Crate:         regex
Version:       1.5.4
Title:         Regexes with large repetitions on empty sub-expressions take a very long time to parse
Date:          2022-03-08
ID:            RUSTSEC-2022-0013
URL:           https://rustsec.org/advisories/RUSTSEC-2022-0013
Solution:      Upgrade to >=1.5.5
```


tisonkun:
Closed as fixed by https://github.com/tokio-rs/prost/pull/604. However, I think this is the earlier PR and be confused why a later patch get merged.

cc @LucioFranco

PR #610, opened 2022-03-25
build: Remove bundled protoc and build from source

### LINKED ISSUE #575: Supply chain security with embedded protoc binaries

LucioFranco:
This is an initial attempt to remove bundled binaries from `prost-build` and to instead compile protobuf from source. This is a much safer approach and reduces prost's vulnerability to supply chain attacks.

Reference https://github.com/tokio-rs/prost/issues/575#issuecomment-1078153519 for more information on how this is implemented.

This also adds a `test-vendored` crate which ensures that vendoring works.

Closes #575


prost-build/Cargo.toml
```diff
@@ -27,6 +31,9 @@ regex = { version = "1.5.5", default-features = false, features = ["std", "unico
 
 [build-dependencies]
 which = { version = "4", default-features = false }
+cfg-if = "1"
+autotools = "0.2"
```
LucioFranco, line 35: needs to be removed

test-vendored/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "test-vendored"
+version = "0.1.0"
+edition = "2021"
```
LucioFranco, line 4: should get a publish = false

prost-build/third-party/update-vendored-protobuf.sh
```diff
@@ -0,0 +1,44 @@
...
+mv "$TEMPDIR/linux-x86_64/include" "$DIR/include/"
+
+# # Update the Protocol Buffers license.
+# mkdir "$TEMPDIR/src"
+# curl --proto '=https' --tlsv1.2 -sSfL \
+#   "https://github.com/protocolbuffers/protobuf/archive/v$VERSION.zip" \
+#   -o "$TEMPDIR/src/protobuf.zip"
+# unzip "$TEMPDIR/src/protobuf.zip" -d "$TEMPDIR/src"
+
+# mv "$TEMPDIR/src/protobuf-$VERSION" "$DIR/protobuf-src"
```
LucioFranco, line 34: Remove this since we track from git

Thomasdezeeuw:
I just want to note that this change (I think) requires `cmake` to be install to build prost, which broke our build pipeline.

PR #611, opened 2022-03-27
feat: implement Ord and PartialOrd for Module

neoeinstein:
This allows a consumer to use `Module` as the key in a `BTreeMap` in addition to a `HashMap`.

### DIFF
--- a/prost-build/src/lib.rs
+++ b/prost-build/src/lib.rs
@@ -1061,7 +1061,7 @@ impl fmt::Debug for Config {
 }
 
 /// A Rust module path for a Protobuf package.
-#[derive(Clone, PartialEq, Eq, Hash)]
+#[derive(Clone, PartialEq, Eq, Hash, PartialOrd, Ord)]
 pub struct Module {
     components: Vec<String>,
 }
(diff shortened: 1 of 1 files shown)


prost-build/src/lib.rs
```diff
@@ -1115,6 +1115,26 @@ impl Module {
     }
 }
 
+impl fmt::Debug for Module {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        fmt::Display::fmt(self, f)
+    }
+}
```
LucioFranco, line 1122: Why not derive this?

LucioFranco (approved):
LGTM, just one question and I think you will have to rebase from master and fix CI then we can merge.

prost-build/src/lib.rs
neoeinstein: That's a good point. Let me step that back one and have it derive instead.

neoeinstein: Updated. This was a relic of my own wrapper class where I _couldn't_ derive `Debug` because `Module` didn't derive it either. `Display` impl is also corrected too.

LucioFranco (approved):

PR #558, opened 2021-11-03
to-spec json mapping via serde

konradjniemiec:
First round for draft review.

Adds serde annotations with the `json_mapping` configuration option.

Verified using the Google `conformance` test suite.

Still missing for initial merging:

- [ ] enums
- [ ] maps with proper de/serialization for primitives
- [ ] Timestamp fixes
- [ ] Duration fixes

Not intending on addressing before merging:
- `google.protobuf.Any` support
- `google.protobuf.Value` support
- `google.protobuf.Struct` support
- `google.protobuf.FieldMask` support
- failing serialization when multiple oneof options are specified
- double min/max values being inexact when serializing

All currently failing tests from `conformance/failing_tests.txt` should fall into one of the above categories ^

PR #615, opened 2022-04-01
impl Timestamp::{Display, FromStr} with RFC-3339 format

danburkert:
This PR implements to and from string methods for
`prost_types::Timestamp` following the RFC-3339 timestamp format, as
also specified by the [Protobuf JSON mapping spec][1]. The `Display`
impl always emits the Zulu (Z) timezone, whereas the parser is slightly
lenient in accepting fixed timezone offsets, as well as both the `T` and
space seperated date and time portions. This leniency in parsing makes
it much much more robust to parsing data of less-than-pristine
formatting, even though the protobuf spec allows no leniency.

Also, since it was easy with the new date time functionality, we've gone
ahead and added a trio of convenience construct functions to `Timestamp`
which makes it straightforward to create `Timestamp`s out of calendar
date time information.

This functionality will be used in follow-up PRs to add Serde
[de]serialization support to the types in `prost_types`, which is itself
a stepping stone towards full JSON support.

This PR was extracted from #558, and is based on code which myself and
@konradjniemiec have written, and which has been running in production
at @sisudata for a few years. The to-string portion of this code has
also been incorporated into [`tracing`][2], so it should be pretty well
battle hardened.

[1]: https://developers.google.com/protocol-buffers/docs/proto3#json
[2]: https://github.com/tokio-rs/tracing/blob/master/tracing-subscriber/src/fmt/time/datetime.rs

### DIFF
--- a/prost-types/src/datetime.rs
+++ b/prost-types/src/datetime.rs
@@ -1,0 +1,778 @@
+//! A date/time type which exists primarily to convert [`Timestamp`]s into an RFC 3339 formatted
+//! string.
+
+use core::fmt;
+
+/// A point in time, represented as a date and time in the UTC timezone.
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub(crate) struct DateTime {
+    /// The year.
+    pub(crate) year: i64,
+    /// The month of the year, from 1 to 12, inclusive.
+    pub(crate) month: u8,
+    /// The day of the month, from 1 to 31, inclusive.
+    pub(crate) day: u8,
+    /// The hour of the day, from 0 to 23, inclusive.
+    pub(crate) hour: u8,
+    /// The minute of the hour, from 0 to 59, inclusive.
+    pub(crate) minute: u8,
+    /// The second of the minute, from 0 to 59, inclusive.
+    pub(crate) second: u8,
+    /// The nanoseconds, from 0 to 999_999_999, inclusive.
+    pub(crate) nanos: u32,
+}
+
+impl DateTime {
+    /// The minimum representable [`Timestamp`] as a `DateTime`.
+    pub(crate) const MIN: DateTime = DateTime {
+        year: -292_277_022_657,
+        month: 1,
+        day: 27,
+        hour: 8,
+        minute: 29,
+        second: 52,
+        nanos: 0,
+    };
+
+    /// The maximum representable [`Timestamp`] as a `DateTime`.
+    pub(crate) const MAX: DateTime = DateTime {
+        year: 292_277_026_596,
+        month: 12,
+        day: 4,
+        hour: 15,
+        minute: 30,
+        second: 7,
+        nanos: 999_999_999,
+    };
+
+    /// Returns `true` if the `DateTime` is a valid calendar date.
+    pub(crate) fn is_valid(&self) -> bool {
+        self >= &DateTime::MIN
+            && self <= &DateTime::MAX
+            && self.month > 0
+            && self.month <= 12
+            && self.day > 0
+            && self.day <= days_in_month(self.year, self.month)
+            && self.hour < 24
+            && self.minute < 60
+            && self.second < 60
+            && self.nanos < 1_000_000_000
+    }
+
+    /// Returns a `Display`-able type which formats only the time portion of the datetime, e.g. `12:34:56.123456`.
+    pub(crate) fn time(self) -> impl fmt::Display {
+        struct Time {
+            inner: DateTime,
+        }
+
+        impl fmt::Display for Time {
+            fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+                // Format subseconds to either nothing, millis, micros, or nanos.
+                let nanos = self.inner.nanos;
+                let subsec = if nanos == 0 {
+                    String::new()
+                } else if nanos % 1_000_000 == 0 {
+                    format!(".{:03}", nanos / 1_000_000)
+                } else if nanos % 1_000 == 0 {
+                    format!(".{:06}", nanos / 1_000)
+                } else {
+                    format!(".{:09}", nanos)
+                };
+
+                write!(
+                    f,
+                    "{:02}:{:02}:{:02}{}",
+                    self.inner.hour, self.inner.minute, self.inner.second, subsec,
+                )
+            }
+        }
+
+        Time { inner: self }
+    }
+}
+
+impl fmt::Display for DateTime {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        // Pad years to at least 4 digits.
+        let year = if self.year > 9999 {
+            format!("+{}", self.year)
+        } else if self.year < 0 {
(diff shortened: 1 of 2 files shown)


prost-types/src/datetime.rs
```diff
@@ -0,0 +1,778 @@
...
+            day,
+            ..DateTime::default()
+        };
+
+        ensure!(date_time.is_valid());
+
+        return Some(crate::Timestamp::from(date_time));
+    }
+
+    // Accept either 'T' or ' ' as delimeter between date and time.
```
danburkert, line 512:
```suggestion
    // Accept either 'T' or ' ' as delimiter between date and time.
```

danburkert:
CC @neoeinstein

prost-types/src/lib.rs
```diff
@@ -257,6 +294,33 @@ impl TryFrom<Timestamp> for std::time::SystemTime {
     }
 }
 
+#[derive(Debug)]
+#[non_exhaustive]
+pub struct TimestampParseError;
```
danburkert, line 299: document

prost-types/src/datetime.rs
```diff
@@ -0,0 +1,756 @@
...
+            write!(f, ".{:03}Z", nanos / 1_000_000)
+        } else if nanos % 1_000 == 0 {
+            write!(f, ".{:06}Z", nanos / 1_000)
+        } else {
+            write!(f, ".{:09}Z", nanos)
+        }
+    }
+}
+
+impl From<crate::Timestamp> for DateTime {
```
danburkert, line 94: `use crate::Timestamp`

prost-types/src/datetime.rs
```diff
@@ -0,0 +1,756 @@
...
+    /// musl's [`__secs_to_tm`][1] converted to Rust via [c2rust][2] and then cleaned up by hand.
+    ///
+    /// All existing `strftime`-like APIs in Rust are unable to handle the full range of timestamps
+    /// representable by `Timestamp`, including `strftime` itself, since tm.tm_year is an int.
+    ///
+    /// [1]: http://git.musl-libc.org/cgit/musl/tree/src/time/__secs_to_tm.c
+    /// [2]: https://c2rust.com/
+    fn from(timestamp: crate::Timestamp) -> DateTime {
```
danburkert, line 102: normalize

prost-types/src/datetime.rs
```diff
@@ -0,0 +1,756 @@
...
+        is_leap,
+    )
+}
+
+/// Parses a timestamp in RFC 3339 format from `b`.
+pub(crate) fn parse_timestamp(s: &str) -> Option<crate::Timestamp> {
+    // Check that the string is ASCII, since subsequent parsing steps use byte-level indexing.
+    ensure!(s.is_ascii());
+
+    let (year, month, day, b) = parse_date(s)?;
```
danburkert, line 474: s/b/s here and below

prost-types/src/datetime.rs
```diff
@@ -0,0 +1,756 @@
...
+    /// All existing `strftime`-like APIs in Rust are unable to handle the full range of timestamps
+    /// representable by `Timestamp`, including `strftime` itself, since tm.tm_year is an int.
+    ///
+    /// [1]: http://git.musl-libc.org/cgit/musl/tree/src/time/__secs_to_tm.c
+    /// [2]: https://c2rust.com/
+    fn from(timestamp: crate::Timestamp) -> DateTime {
+        let t = timestamp.seconds;
+        let nanos = timestamp.nanos;
+
+        // 2000-03-01 (mod 400 year, immediately after feb29
```
LucioFranco, line 106: Missing `)`?

prost-types/src/datetime.rs
```diff
@@ -0,0 +1,756 @@
...
+    }
+    leaps += 97 * cycles + 24 * centuries - i64::from(is_leap);
+
+    (
+        i128::from((year - 100) * 31_536_000) + i128::from(leaps * 86400 + 946_684_800 + 86400),
+        is_leap,
+    )
+}
+
+/// Parses a timestamp in RFC 3339 format from `b`.
```
LucioFranco, line 469:
```suggestion
/// Parses a timestamp in RFC 3339 format from `s`.
```

prost-types/src/lib.rs
```diff
@@ -165,6 +169,39 @@ impl Timestamp {
...
+        let date_time = datetime::DateTime {
+            year,
+            month,
+            day,
+            hour,
+            minute,
+            second,
+            nanos,
+        };
+        assert!(date_time.is_valid(), "invalid date time: {}", date_time);
```
LucioFranco, line 202: I not a huge fan of these potential panics, should we make them an error?

LucioFranco (commented):
LGTM mostly one question and also gonna really trust proptest on this one

prost-types/src/datetime.rs
danburkert: This typo exists in the original C version: https://git.musl-libc.org/cgit/musl/tree/src/time/__secs_to_tm.c#n4 .  I've never fixed it because I'm not exactly sure where the closing paren should go.

LucioFranco: lol

prost-types/src/lib.rs
danburkert: Yeah I'm in favor of making this an error. This PR already introduces one new error type, `TimestampParseError`, and there is already an existing `TimestampOutOfSystemRangeError`, and this would be an additional.  WDYT about folding all of these cases into an enum of 3 variants, instead of separate types?  That would be a breaking change, but I imagine there aren't many users of `TimestampOutOfSystemRangeError` so it'd hopefully not be too painful of an upgrade for folks.

danburkert: (for a bit more context, subsequent PRs are likely going to require a new version bump anyway because I'm planning on adding default trait impls for generated types, which is technically a breaking change as well)

prost-types/src/lib.rs
danburkert: Done.

prost-types/src/datetime.rs
danburkert: Done.

prost-types/src/datetime.rs
danburkert: Done.

prost-types/src/datetime.rs
danburkert: Done.

prost-types/src/datetime.rs
danburkert: Done.

prost-types/src/lib.rs
danburkert: I've gone ahead and pushed this as the latest commit

LucioFranco: Yeah I am happy to make that breaking change.

LucioFranco:
@danburkert is this ready for final review?

danburkert:
Yes, this is ready for final review.   Thanks @LucioFranco !

LucioFranco (approved):

Issue #353, opened 2020-08-02
Include LICENSE file in prost-derive crate


nullr0ute:
The LICENSE file isn't shipped in the prost-derive crate, it is in the prost crate.

nullr0ute:
Nor is it shipped in prost-build or rust-prost-types crates

danburkert:
Would accept a PR!  If the repo has symlinks would `cargo publish` handle that correctly, or does it have to be actual file copies?

caspermeijn:
This has been fixed in the latest release

PR #619, opened 2022-04-06
Symlink LICENSE into crates (#353)

tjdetwiler:
Symlink the top-level LICENSE into each published crate.


tjdetwiler:
Hi Lucio, is there a rough estimate for when the next crates.io release may happen that would include this? If there isn't one planned for the near future, would it be possible to cut a patch release containing this change?

PR #620, opened 2022-04-07
Remove cmake

Jake-Shadle:
We noticed due to https://github.com/hyperium/tonic/issues/965 that `cmake` was now a dependency for tonic via prost. We've had cmake banned for quite a while since it is terrible, so I started working on a PR to make prost-build use cc to build protoc instead. However, upon investigation I quickly realized that building and shelling out to protoc is _massive_ overkill for prost's usage, which essentially just boils down to getting a list of the input files and all of their imports. Protoc by contrast includes support for generating multiple different languages (none of which prost uses) and plugin support (again, something that prost doesn't use), so I instead just reimplemented the small part of the command line tool that prost actually depended on into a single C function that is now linked with prost-build, and when compile_protos is called, just invokes that C function to get the output.

### Pros

- No cmake
- Only compiles the minimum amount of C++ code that is actually needed to write the file descriptor set, so the build is significantly faster than compiling the protoc binary
- No shelling out to a separate program and using the filesystem to read/write the output

### Cons

- This does the _one_ thing that "vanilla" prost-build does, so usage of additional protoc arguments that people may be depending on won't do anything. I consider this fine since in those more advances use cases, you can just use an installed protoc compiler.
- Added C++ code. I felt bad while doing it.
- Unsafe, of course, to talk with the C code.

### Other options

- There already exists a Rust native way to read generate FileDescriptorSets at https://github.com/stepancheg/rust-protobuf/tree/master/protobuf-parse, it states that it might not be fully feature complete compared to protoc, but I think it would be fairly trivial to fix any issues there and instead use pure Rust code, allowing prost to completely remove the protobuf code altogether.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -1,3 +1,9 @@
+[advisories]
+ignore = [
+    # serde_cbor is unmaintained, but brought in by criterion
+    "RUSTSEC-2021-0127",
+]
+
 [licenses]
 allow = ["Apache-2.0", "MIT", "BSD-3-Clause"]
 
--- a/deny.toml
+++ b/deny.toml
@@ -1,0 +1,11 @@
+[licenses]
+allow = ["Apache-2.0", "MIT", "BSD-3-Clause"]
+
+[bans]
+deny = [{ name = "cmake" }]
+skip = [
+    # old csv via old criterion
+    { name = "itoa", version = "=0.4.8" },
+    # Proptest includes 2 version :(
+    { name = "quick-error", version = "=1.2.3" },
+]
--- a/.github/workflows/continuous-integration-workflow.yaml
+++ b/.github/workflows/continuous-integration-workflow.yaml
@@ -138,3 +138,11 @@ jobs:
       - name: cargo check
         run: cd test-vendored && cargo check
 
+  deny:
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2
+        with:
+          submodules: recursive
+      - uses: EmbarkStudios/cargo-deny-action@v1
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "prost-build/third-party/protobuf"]
 	path = prost-build/third-party/protobuf
-	url = git@github.com:protocolbuffers/protobuf
+	url = https://github.com/protocolbuffers/protobuf
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "prost"
-version = "0.10.1"
+version = "0.10.3"
 authors = [
     "Dan Burkert <dan@danburkert.com>",
     "Tokio Contributors <team@tokio.rs>",
--- a/prost-build/Cargo.toml
+++ b/prost-build/Cargo.toml
@@ -30,9 +30,9 @@ lazy_static = "1.4.0"
 regex = { version = "1.5.5", default-features = false, features = ["std", "unicode-bool"] }
 
 [build-dependencies]
-which = { version = "4", default-features = false }
+cc = { version = "1.0", features = ["parallel"] }
 cfg-if = "1"
-cmake = "0.1"
+which = { version = "4", default-features = false }
 
 [dev-dependencies]
 env_logger = { version = "0.8", default-features = false }
--- a/prost-build/build.rs
+++ b/prost-build/build.rs
@@ -73,15 +73,62 @@ fn vendored() -> bool {
     }
 }
 
-/// Compile `protoc` via `cmake`.
+/// Compile `protoc`.
 fn compile() -> Option<PathBuf> {
-    let protobuf_src = bundle_path().join("protobuf").join("cmake");
+    let protobuf_src = bundle_path().join("protobuf/src/google/protobuf");
 
     println!("cargo:rerun-if-changed={}", protobuf_src.display());
 
-    let dst = cmake::Config::new(protobuf_src).build();
+    // compile our protoc wrapper lib
+    {
+        let mut build = cc::Build::new();
+        build.cpp(true);
+        build.flag("-Wno-unused-parameter");
 
-    Some(dst.join("bin").join("protoc"))
+        build.includes([bundle_path().join("protobuf/src")]);
+
+        build.files(
+            [
+                "any.cc",
+                "any_lite.cc",
+                "arena.cc",
+                "descriptor.cc",
+                "dynamic_message.cc",
+                "extension_set.cc",
+                "map_field.cc",
+                "message.cc",
+                "message_lite.cc",
+                "generated_message_reflection.cc",
+                "generated_message_util.cc",
+                "parse_context.cc",
+                "repeated_field.cc",
+                "repeated_ptr_field.cc",
+                "text_format.cc",
+                "unknown_field_set.cc",
+                "wire_format.cc",
+                "wire_format_lite.cc",
+                "io/coded_stream.cc",
+                "io/strtod.cc",
+                "io/tokenizer.cc",
+                "io/zero_copy_stream.cc",
+                "io/zero_copy_stream_impl_lite.cc",
+                "stubs/common.cc",
+                "stubs/stringpiece.cc",
+                "stubs/stringprintf.cc",
+                "stubs/structurally_valid.cc",
+                "stubs/strutil.cc",
+            ]
+            .iter()
+            .map(|fname| protobuf_src.join(fname)),
+        );
+
+        // This is our little wrapper that only does the 1 thing prost-build
+        // actually needs from the the bloated protoc binary
+        build.file("src/libprotoc.cpp");
+        build.compile("protoc");
+    }
+
+    Some(PathBuf::from("linked"))
 }
 
 /// Try to find a `protoc` through a few methods.
--- a/prost-build/build.rs
+++ b/prost-build/build.rs
@@ -144,7 +144,11 @@ fn compile(stub: bool) -> Option<PathBuf> {
             );
 
             if env::var("CARGO_CFG_TARGET_FAMILY").as_deref() == Ok("windows") {
-                build.file(protobuf_src.join("io/io_win32.cc"));
+                build.files(
+                    ["io/io_win32.cc", "stubs/status.cc"]
+                        .iter()
+                        .map(|fname| protobuf_src.join(fname)),
+                );
             }
         }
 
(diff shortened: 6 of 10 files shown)


LucioFranco:
I am curious if this supports https://github.com/tokio-rs/console/blob/main/console-api/build.rs#L17 this flag?

Jake-Shadle:
That flag is meaningless https://github.com/protocolbuffers/protobuf/blob/55177c3574ce2718acf3d91cfcfbe9d035a6836d/src/google/protobuf/compiler/command_line_interface.cc#L1854-L1855

LucioFranco:
I must be allergic to c++ code thanks for getting that lmaooo

deny.toml
```diff
@@ -0,0 +1,11 @@
+[licenses]
```
neoeinstein, line 1: Is the intent here to impose Embark Studio’s deny configuration onto Prost?

Jake-Shadle: This isn't "our" deny configuration, this is just a way to ensure that the dependency this PR is removing doesn't return in the future.

olix0r:
It doesn't necessarily have to be done as a part of this PR, but I wonder if we could use a feature flag to control whether any of these dependencies are incurred... We'd prefer to only ever use `PROTOC_NO_VENDOR`, so there's no need for us to incur the `cc` (or `cmake`) dependency. It would be fine for the feature to be enabled by default. And we'd have to follow suit and re-expose the feature from `tonic-build`.

LucioFranco:
@olix0r I wonder if we should provide a subset of `prost-build` that contains no vendoring? `prost-build-cc` and `prost-build-lite` or something like that that gives you clear control. Though I wonder how this would affect down stream users of libraries.

olix0r:
@LucioFranco yeah, I'm not sure what the most usable approach is. The decision to use vs build a `protoc` is really up to the _application_ that is building a library that includes protobuf. While, the library itself needs to depend on `prost-build`. The library shouldn't be opinionated on whether protoc is used or not, which is why I thought the feature-flag approach might be best.

ricnewton:
Is there any more progress on this PR?  So for our code now since the prebuilt protoc has been removed our build time is now dominated by prost-build, especially on low spec machines like the lower end github codespaces machines.

For a build of one of our crates which is mainly dependent on prost-build (and tonic-build), on the lowest spec codespace machine (2 cores, admittedly pretty much worst case), the time it takes to build using normal prost-build was around 11 mins 40s, and with this branch it was around 2 mins 10s.

If the changes here are too much, we also tried just disabling building the protobuf tests: https://github.com/ricnewton/prost/commit/d423ca122b0f1b763a38a996e801fcff25f71364, this built in around 4 mins 40s so significantly better than master but not as good as this branch, I can raise a PR if that is acceptable and this branch isn't, but this branch is faster to build.  Any reason we would need to build the protobuf tests?

Jake-Shadle:
I'm unable to repro the CI failure locally, but I've submitted a change that should fix it.

Jake-Shadle:
Is there some way to mark this branch/me so that CI can run when I do pushes? Due to the time difference I basically have to wait like 8 hours to see if the changes I make pass CI since apparently my local environment is different enough that I don't catch all of the same things that CI does, so the turnaround time is making this PR last waaaaaaay longer than it should.

Jake-Shadle:
This is ready for review by the way.

Jake-Shadle:
We were holding off our upgrade of prost/tonic/etc until this landed but since it's taking a while to get merged/released I'm adding a patch for our code, but that's when I noticed that the protobuf submodule is using an SSH url, which makes fetching via cargo untenable, so I've switched the url to a plain HTTPS so that this repo is easier to patch.

Jake-Shadle:
It also seems that prost `0.10.2` and `0.10.3` were released without bumping the actual version in the Cargo.toml file.

LucioFranco:
@Jake-Shadle those patch releases are on the `v0.10` branch, master contains some breaking changes.

Jasperav:
@Jake-Shadle is this already merged or? My CI is failing:

running: "cmake" "/home/runner/.cargo/git/checkouts/prost-0f318a28dfb453a6/2c9d9c1/prost-build/third-party/protobuf/cmake" "-DCMAKE_INSTALL_PREFIX=/home/runner/work/xx/xx/server/target/debug/build/prost-build-744ea9bdfe2d1c67/out" "-DCMAKE_C_FLAGS= -ffunction-sections -fdata-sections -fPIC -m64" "-DCMAKE_C_COMPILER=/usr/bin/cc" "-DCMAKE_CXX_FLAGS= -ffunction-sections -fdata-sections -fPIC -m64" "-DCMAKE_CXX_COMPILER=/usr/bin/c++" "-DCMAKE_ASM_FLAGS= -ffunction-sections -fdata-sections -fPIC -m64" "-DCMAKE_ASM_COMPILER=/usr/bin/cc" "-DCMAKE_BUILD_TYPE=Debug"

  --- stderr
  CMake Error: The source directory "/home/runner/.cargo/git/checkouts/prost-0f318a28dfb453a6/2c9d9c1/prost-build/third-party/protobuf/cmake" does not exist.

Jake-Shadle:
Nope, don't think so

LucioFranco:
@ricnewton still looking to merge this, I just have not had enough time to full review this yet.

ricnewton:
Great, thanks.  If there are any issues or anything or it will take a long time I can submit the one line change that at least disables running the protocol buffer tests if that helps, that saves loads of time as well.

benesch:
Just dropping by from @MaterializeInc to say that this is unfortunately a nonstarter for us. Appreciate the work you've done here, @Jake-Shadle, and I'm sorry it was our supply chain concerns that kicked off this whole mess, but if prost-build decides to essentially reimplement the protobuf build system we'll have to switch back to a fork that uses a stock vendored protoc. (I'm not a prost maintainer, just an opinionated community member, so please take this as just a data point.)

To provide some more context: I've been burned repeatedly by projects that reimplement upstream build systems. Build systems, especially C++ build systems, tend to accrete important platform-specific knowledge, built up slowly over years of bug reports because there are never any tests. All that knowledge gets lost when you fork the build system. CockroachDB maintained custom build systems for its C++ dependencies for a while (e.g., [RocksDB](https://github.com/cockroachdb/rocksdb)), and it was a persistent source of bugs due to compiler flags not being set correctly [[0](https://github.com/cockroachdb/rocksdb/commit/66bdf29381e608badae6d2afb3ca2584ab23e2a5)][[1](https://github.com/cockroachdb/rocksdb/commit/87c3949e360f5ce178e8074cfb90795c8c229be0)].

There is also a permanent maintainability to forking a build system. Someone will need to update the list of source files here for every protobuf release. Someone will need to look through upstream's autotools/CMake build system changes with every release and ensure that any feature detection that results in new compiler flags or defines gets carried through here. Downstream maintainers are generally not well positioned to do that sort of work, since they are relatively unfamiliar with the upstream codebase.

> We've had cmake banned for quite a while since it is terrible...

I think this is an unfair characterization! CMake, like most tools, has its pros and cons. I've found CMake to be by far the most robust of the C/C++ build systems.

> I think it would be fairly trivial to fix any issues there and instead use pure Rust code, allowing prost to completely remove the protobuf code altogether.

Last I looked into this it is far from trivial, I'm afraid. prost relies on the `--include_source_info` flag to protoc, which is not supported by protobuf-parse. The source info is quite detailed and seemed to require a substantial overhaul of protobuf-parse in order to put the right bookkeeping into place.

----

> If the changes here are too much, we also tried just disabling building the protobuf tests: https://github.com/ricnewton/prost/commit/d423ca122b0f1b763a38a996e801fcff25f71364, this built in around 4 mins 40s so significantly better than master but not as good as this branch, I can raise a PR if that is acceptable and this branch isn't, but this branch is faster to build. Any reason we would need to build the protobuf tests?

This seems like a slam dunk to me, FWIW.

----

Is there a reason that you can't `apt install protobuf-compiler` or download the precompiled release from upstream in your CI? That'd be even faster than the approach in this PR. Coupled with #648 you'd even be able to remove the `cmake` crate from your transitive dependency graph.

In any case, my preference would be to figure out how to modularize prost-build! Then the code here could be in a separate crate/feature for the folks who want it.

Jake-Shadle:
I get where you're coming from, but in this case protobuf is a fairly small project, especially relative to the likes of rocksdb, and prost-build only uses a small fraction of it. In addition, it's not doing really doing anything complicated nor architecture specific, which is the typical reason for special flags and the like, but I could if course be wrong thinking that this would be trivial to maintain since I have never used protoc directly so have no familiarity with their release cadence or the like.

> I think this is an unfair characterization! CMake, like most tools, has its pros and cons. I've found CMake to be by far the most robust of the C/C++ build systems.

Agree to disagree I guess, CMake may the best in C/C++ land but that's really not saying much. It only brought problems for us, easier to purge it completely so that the Rust ecosystem doesn't have to rely on it.

> Last I looked into this it is far from trivial, I'm afraid. prost relies on the --include_source_info flag to protoc, which is not supported by protobuf-parse. The source info is quite detailed and seemed to require a substantial overhaul of protobuf-parse in order to put the right bookkeeping into place.

That's unfortunate, IMO using pure Rust would be the ideal solution since it would mean getting rid of the C++ code altogether and making prost-build and the crates that depend on it much smaller and faster to compile. Bad assumption on my part.

> Is there a reason that you can't apt install protobuf-compiler or download the precompiled release from upstream in your CI? That'd be even faster than the approach in this PR. Coupled with https://github.com/tokio-rs/prost/pull/648 you'd even be able to remove the cmake crate from your transitive dependency graph.

Requiring an additional installation dependency to build a single crate dependency seems pretty terrible IMO. If we had other cmake dependencies it might be acceptable, but we purged them years ago and they are never coming back, requiring cmake for CI and users who build our code would be a major step back.

This PR still supports `PROTOC_NO_VENDOR` I don't understand why this solution can't coexist with those like yourself who want to use the stock protoc.

Jake-Shadle:
Hmm, so should I just close this then? This PR has been open coming up on 2 months now and it feels like it's getting less likely to merge given this recent feedback.

I'd of course rather _not_ maintain our own fork and patch it in to resolve something that also seems to be problematic for other people who use prost, but this code is fairly isolated from the bulk of what prost-build does so should be easy to keep in sync with new releases, barring big changes such as the original one to vendor sources instead of using the bundled protoc.

LucioFranco:
hi! I wrote a proposal that removes cmake and can also support this work. Let me know what you think in there! Thanks for putting this together jake <3

PR #625, opened 2022-04-14
impl Duration::{Display, FromStr} with protobuf format

danburkert:
This PR implements to and from string methods for
`prost_types::Duration` following the duration format defined by the
[Protobuf JSON mapping spec][1].

This is a followup to #615 and re-uses a lot of the same parser helpers.
Additionally I did the same error type unification with a new
`DurationError` type. This is a breaking change for `prost-types` since
public APIs are changed slightly.

[1]: https://developers.google.com/protocol-buffers/docs/proto3#json


LucioFranco (approved):
LGTM there are a few clippy warnings we should probably clean up before merging

PR #628, opened 2022-04-15
prost-types: Implement Eq for Any and Duration

theduke:
Manually implement the Eq trait for the Any and Duration well-known
types.

These types only contain fields that are themselfes Eq.


theduke:
It looks like `protobuf.rs` is auto-generated, but I couldn't find a script or similar that does this, so I just added the impls in `lib.rs`

LucioFranco:
@theduke the code is auto generated in the `bootstrap` test under `tests`.

prost-types/src/lib.rs
```diff
@@ -331,6 +331,17 @@ impl fmt::Display for Timestamp {
...
+// Implement stable Eq for Duration.
+// Duration only contains Eq fields.
+// Done manually since protobuf.rs is generated.
+impl Eq for Duration {}
+
+// Implement stable Eq for Any.
+// Any only contains Eq fields.
+// Done manually since protobuf.rs is generated.
+#[cfg(feature = "std")]
+impl Eq for Any {}
```
LucioFranco, line 343: I feel like this may actually be wrong. How do you define equality for something like `Any` since it erases types.

theduke: Any is just a container for a type descriptor and a raw payload.

It's perfectly reasonable to compare those.

As it stands I need to map the type into my own `Any` to include it in other types that need to be `Eq`

danburkert:
Another way to do this is by injecting the derive impls into the generated code via https://github.com/tokio-rs/prost/blob/master/tests/src/bootstrap.rs#L23 and [`type_attribute()`](https://docs.rs/prost-build/latest/prost_build/struct.Config.html#method.type_attribute).

PR #637, opened 2022-04-28
Implement Message for uuid::Uuid

### LINKED ISSUE #568: Example of docs for `extern_path` doesn't compile
I have followed the example for adding the 'uuid' crate as a dependency in here: https://docs.rs/prost-build/0.9.0/prost_build/struct.Config.html#usage

Everything works, as I can see the generated code is now:

```rust
pub struct Order {
    #[prost(message, optional, tag = "1")]
    pub id: ::core::option::Option<::uuid::Uuid>,
..
}
```

however, it doesn't compile because `the trait prost::Message is not implemented for Uuid`

I find the docs confusing as I understood from this [issue](https://github.com/tokio-rs/prost/issues/537) that Uuid as an external crate is not supported anyway. This is me hoping that I'm missing something from the docs

eexsty:
# Pull Request Information

Title is self-explanatory.

* Closes #568

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "prost-build/third-party/protobuf"]
 	path = prost-build/third-party/protobuf
-	url = git@github.com:protocolbuffers/protobuf
+	url = https://github.com/protocolbuffers/protobuf.git
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,13 +41,15 @@ exclude = [
 bench = false
 
 [features]
-default = ["prost-derive", "std"]
+default = ["prost-derive", "std", "with-uuid"]
 no-recursion-limit = []
 std = []
+with-uuid = ["uuid"]
 
 [dependencies]
 bytes = { version = "1", default-features = false }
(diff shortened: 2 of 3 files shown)


neoeinstein:
We generally have not elected to take optional implementations like this into Prost. This is in part because the implementation you propose here, while one way to do it, is not necessarily the one cannonical form that we should bless. (The Protobuf specification does not describe a canonical way to encode/decode a UUID.) Some users out there are likely encoding UUIDs as strings, while others are encoding them, as you have, by using bytes.

In this case, this implementation is likely better left to a newtype wrapper that does what you are looking for, rather than being an integral part of either `uuid` or `prost`.

LucioFranco:
Right the correct solution is for either `uuid` to provide the `Message` impl  (like it does for serde) or for you to do a newtype warpper and implement it yourself.

adriangb:
> In this case, this implementation is likely better left to a newtype wrapper that does what you are looking for, rather than being an integral part of either `uuid` or `prost`.

It would be nice to point to a complete example of this pattern.

PR #605, opened 2022-03-11
build: Skip empty protobuf files

wangfenjin:
close #228

PR #639, opened 2022-05-04
Revert "build: Skip empty protobuf files (#605)"

LucioFranco:
This reverts commit c577d3f.

Fixes https://github.com/hyperium/tonic/issues/964

### DIFF
--- a/prost-build/src/lib.rs
+++ b/prost-build/src/lib.rs
@@ -880,10 +880,6 @@ impl Config {
 
         let modules = self.generate(requests)?;
         for (module, content) in &modules {
-            if content.is_empty() {
-                continue;
-            }
-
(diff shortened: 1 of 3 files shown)


tustvold:
I'm a little confused by this, reading https://github.com/hyperium/tonic/issues/964 it seems like the issue is that if someone is specifying `include_file("google.protobuf.rs")` and this file is stripped out because it is empty it doesn't work. Is this actually a problem? The ticket says that removing the `include_file` eliminates the issue, so isn't the fix to just not add redundant include_file calls in build.rs?

I mainly ask because I'm adding workarounds for the empty files, and it would be nice to not need to https://github.com/apache/arrow-rs/pull/2005

PR #644, opened 2022-05-07
Min-versions fixes and CI

g2p:
Ensure things build when resolving dependencies to minimal versions.
Add a CI check as well.


LucioFranco (approved):
Sorry for the delay, thanks for doing this!

PR #655, opened 2022-05-24
Disable building protobuf tests when building vendored protoc

ricnewton:
Pending the outcome of https://github.com/tokio-rs/prost/pull/620, can we at least disable running the protocol buffer tests when building protoc, this speeds up the build quite considerably.


LucioFranco:
Thanks! will try to get this out in a release for prost

Issue #653, opened 2022-05-21
prost-build occasionally fails to compile protobuf C++


benesch:
In v0.10, we're seeing `prost-build` occasionally fail to compile protobuf from source. We've now seen this in two projects with entirely different build systems (MaterializeInc/materialize and our internal cloud codebase), which makes me think it's not a bug in our build system. The only commonality is in that in both cases we're compiling inside of Docker. Here's two representative examples:

* https://gist.github.com/antifuchs/bfb4b8e1dc673ce7a09deb5b7c43bfcd
* https://buildkite.com/materialize/tests/builds/34688#9076b67a-eae2-41f6-b144-8483d97f5c15

I have a few separate thoughts here!

 * Is CMake even supported for building protobuf on Unix machines? The [docs](https://github.com/protocolbuffers/protobuf/blob/main/cmake/README.md) seem to suggest that the CMake build system exists to support compilation on Windows. 
 * In our [protobuf-src](https://github.com/MaterializeInc/rust-protobuf-native/tree/master/protobuf-src) crate, we use the autotools build system instead, because it seems to be the officially supported way to compile protobuf on Unix. That was stable in CI for us for several months, when we'd forked prost to use [protobuf-src](https://github.com/MaterializeInc/rust-protobuf-native/tree/master/protobuf-src). 
 * If you don't want to switch to the autotools system, is there some way to turn off compilation of the unit tests? That's what seems to fail most readily. 
 * It'd be great to eventually have prost-build use protobuf-src rather than managing its own protoc compilation. That way projects that use protoc outside of prost only need to compile protobuf from source once. We'd be happy to transfer ownership of protobuf-src to Tokio if that's a supply chain concern.

LucioFranco:
Right, I will say I am not a c++ build tool expert and it could totally be that cmake has some issues on non-windows builds. Though I don't know why 😆 

> In our [protobuf-src](https://github.com/MaterializeInc/rust-protobuf-native/tree/master/protobuf-src?rgh-link-date=2022-05-21T15%3A50%3A52Z) crate, we use the autotools build system instead, because it seems to be the officially supported way to compile protobuf on Unix. That was stable in CI for us for several months, when we'd forked prost to use [protobuf-src](https://github.com/MaterializeInc/rust-protobuf-native/tree/master/protobuf-src?rgh-link-date=2022-05-21T15%3A50%3A52Z).

[...]

Issue #650, opened 2022-05-14
CI fails due to cmake failure


Jasperav:
I recently updated my fork of this and now my CI fails with this error:

```
  running: "cmake" "/home/runner/.cargo/git/checkouts/prost-0f318a28dfb453a6/2c9d9c1/prost-build/third-party/protobuf/cmake" "-DCMAKE_INSTALL_PREFIX=/home/runner/work/xx/xx/server/target/debug/build/prost-build-744ea9bdfe2d1c67/out" "-DCMAKE_C_FLAGS= -ffunction-sections -fdata-sections -fPIC -m64" "-DCMAKE_C_COMPILER=/usr/bin/cc" "-DCMAKE_CXX_FLAGS= -ffunction-sections -fdata-sections -fPIC -m64" "-DCMAKE_CXX_COMPILER=/usr/bin/c++" "-DCMAKE_ASM_FLAGS= -ffunction-sections -fdata-sections -fPIC -m64" "-DCMAKE_ASM_COMPILER=/usr/bin/cc" "-DCMAKE_BUILD_TYPE=Debug"

  --- stderr
  CMake Error: The source directory "/home/runner/.cargo/git/checkouts/prost-0f318a28dfb453a6/2c9d9c1/prost-build/third-party/protobuf/cmake" does not exist.
```

How can I fix this?

The CI: https://github.com/Jasperav/prost/runs/6422877406?check_suite_focus=true

LucioFranco:
@Jasperav can you post the version of `prost-build` you are using?

LucioFranco:
The fork is missing the git submodule of the protobuf repo. https://github.com/Jasperav/prost/tree/master/prost-build/third-party check this folder on the master branch of this repo.

Jasperav:
@LucioFranco Thanks for the reply, I got it working now!

Issue #646, opened 2022-05-09
[prost-build] avoid cmake dependency


nstinus:
Today, prost-build tries to find protoc and if it fails, builds it from the vendored sources.
That adds the cmake dependency even though it may not be needed.

Could the default behavior be changed to:
- look for protoc in PROTOC or PATH, then fail.
- if vendored feature is enable, use this one only.

There is the PROTOC_NO_VENDOR env var but the cmake dep is still required.
On top of removing a dependency in some cases, it would also make clearer what is happening.

LucioFranco:
#657 fixes this

Issue #645, opened 2022-05-08
prost-build fails to find cmake on macOS (M1)


codemaster138:
I'm trying to use a package that in turn uses `prost-build` (`v0.10.3`, current latest release), but cargo refuses to build it (see full error message at the end). I'm on a 2020 M1 MacBook Pro 13" running macOS Monterey (12.0.1)

The error message asks me to make sure `cmake` is installed, which, _it is_ (`cmake --version` outputs `cmake version 3.23.1`). The only other reference to this problem I was able to find was [this question](https://stackoverflow.com/questions/72061535/getting-this-error-message-while-installing-solana-error-failed-to-run-custom) on SO, which has no answers...

```
build failed
error: failed to run custom build command for `prost-build v0.10.3`
Caused by:
  process didn't exit successfully: `/Users/test/Desktop/projects/scratchpad/supercap/server/target/rls/debug/build/prost-build-7df80aee7bef1b3a/build-script-build` (exit status: 101)
  --- stdout
  cargo:rerun-if-changed=/Users/test/.cargo/registry/src/github.com-1ecc6299db9ec823/prost-build-0.10.3/third-party/protobuf/cmake
  CMAKE_TOOLCHAIN_FILE_aarch64-apple-darwin = None
  CMAKE_TOOLCHAIN_FILE_aarch64_apple_darwin = None
  HOST_CMAKE_TOOLCHAIN_FILE = None
  CMAKE_TOOLCHAIN_FILE = None
  CMAKE_GENERATOR_aarch64-apple-darwin = None
  CMAKE_GENERATOR_aarch64_apple_darwin = None
  HOST_CMAKE_GENERATOR = None
  CMAKE_GENERATOR = None
  CMAKE_PREFIX_PATH_aarch64-apple-darwin = None
  CMAKE_PREFIX_PATH_aarch64_apple_darwin = None
  HOST_CMAKE_PREFIX_PATH = None
  CMAKE_PREFIX_PATH = None
  CMAKE_aarch64-apple-darwin = None
  CMAKE_aarch64_apple_darwin = None
  HOST_CMAKE = None
  CMAKE = None
[...]

PR #657, opened 2022-05-24
Remove build.rs

LucioFranco:
This PR is a proposal for updating (again!) how `prost` and more specifically `prost-build` gains access to `protoc`.

## Problem

Prost does not implement a pure rust parser for protobuf file definitions. Instead, it uses `protoc` to parse the protobuf definitions. Version `0.8` and before originally bundled a copy of protoc for each platform to be used when the environment variables are no present and `protoc` doesn't exist in the `PATH`.  This (#575 and #562) provides an avenue for supply chain attacks and was removed in version `0.9` and replaced by bundling the `protobuf` source code and using `cmake` to compile it. While `cmake` _works_ (it is the recommended way to build for windows) it does have some issues (#653, #646, #650, #626,  #645).

## Solution

This PR introduces even tighter requirements for sourcing `protoc`. This change forces users to either have `protoc` in their PATH or to supply it via `PROTOC` environment variable. With this change as well the new recommendation for libraries is to commit their generated protobuf.

To supplement this change, prost will take ownership of [`protobuf-src` ][proto-src] and potentially add `protobuf-src-cc` via the work in  https://github.com/tokio-rs/prost/pull/620.

[proto-src]: htps://github.com/MaterializeInc/rust-protobuf-native/tree/master/protobuf-src

In addition, a blog post will be written on the tokio.rs blog to explain how users should use prost-build with these changes.


LucioFranco:
For CI to pass there are a few upstream fixes that need to happen but locally everything works on my mac.

cc @neoeinstein @benesch @jake-shadle @danburkert I would love some feedback from yall if you get the chance.

benesch:
Thanks so much for working on this, @LucioFranco! This seems like a great solution for application crates. The big question I have is how this will interact with other library crates. Let's assume tonic-build exposes the same two methods to configure protoc. What do downstream library crates like opentelemetry-otlp and console-api do? They have to pick _some_ values for protoc/protoc_include. Is your thinking that they'll all just check the generated protobuf code into source control? (IIRC you recently switched console-api to do that?)

Jake-Shadle:
Yah I have the same concern mentioned by @benesch, for applications this would be fine but seems like it will a bit of a nightmare with transitive dependencies since this method seems like it would need to leak down via eg feature flags otherwise you will run into cases where some libraries will compile from source and some will require protoc be in a know location or `PATH`. If it could become standard to include the generated sources in library crates that would avoid this situation and the need to run protobuf generation at build time at all, but that seems like a big change, and could I imagine be problematic for some scenarios that I'm not thinking of? Either way this seems like this will be another big ecosystem change...?

LucioFranco:
>  Let's assume tonic-build exposes the same two methods to configure protoc. What do downstream library crates like opentelemetry-otlp and console-api do? They have to pick some values for protoc/protoc_include. Is your thinking that they'll all just check the generated protobuf code into source control?

Yes, I think overall this is the right solution moving forward (also helps IDEs etc). That said, I think it still is a challenge, for example, how do you ensure that the proto and generated code are updated in lock step? I guess one solution, would be to feature flag the generation in a test (similar to what prost does for prost-types) and running that on CI only where you can install protoc very easily.

> for applications this would be fine but seems like it will a bit of a nightmare with transitive dependencies since this method seems like it would need to leak down via eg feature flags otherwise you will run into cases where some libraries will compile from source and some will require protoc be in a know location or PATH.

Yeah, this is my concern as well. Though I think libs should include their own way to fetch protoc and should/could provide ways for users to provide their own by exposing methods. The one thing I think going for us in this area is that I don't think there are a large amount of libs that use tonic directly and/or with gRPC users can generally include the protobuf themselves. That said, I think the real solution moving forward is to check in the generated protobuf. 

I wonder, if with this change writing a blog post as well might alleviate some of the pain introduced by explaining the right way to write libraries and applications.

prost-build/src/lib.rs
```diff
@@ -1289,6 +1345,7 @@ mod tests {
         let _ = env_logger::try_init();
         Config::new()
             .service_generator(Box::new(ServiceTraitGenerator))
+            .out_dir(std::env::temp_dir())
```
neoeinstein, line 1348: Would we want to use something like `tempdir`? It can ensure we don't have test case collisions (if we add more that _would_ collide later) and then cleans up the generated temp directory on drop. Not vital by any means.

neoeinstein (approved):
> That said, I think the real solution moving forward is to check in the generated protobuf.

I tend to agree with this. I've moved to check in generated code in our own repositories too, albeit using `buf` to drive a prost-based generator. This is better than making a difficult-to-verify binary executable part of the distribution or spending several times longer building that executable from source just to do the transformation. Thus, I think the default should be to expect a user BYOprotoc or for the library to pre-generate.

prost-build/src/lib.rs
LucioFranco: Yeah, we can fix this in another PR but prob a good idea.

tests/src/well_known_types.proto
```diff
@@ -1,8 +1,17 @@
...
+import "google/protobuf/any.proto";
+import "google/protobuf/field_mask.proto";
+import "google/protobuf/api.proto";
+import "google/protobuf/descriptor.proto";
 import "google/protobuf/wrappers.proto";
+import "google/protobuf/source_context.proto";
+import "google/protobuf/empty.proto";
+import "google/protobuf/struct.proto";
+import "google/protobuf/compiler/plugin.proto";
+import "google/protobuf/duration.proto";
```
LucioFranco, line 14: So I learned that `protoc` bundles all of these includes so you don't actually need to have a protoc_include env var set with these nor do we need to bundle them into the `.crate` for `prost-build`.

neoeinstein (approved):

notmgsk:
Whatever happened to that blog post? :)

LucioFranco:
> Whatever happened to that blog post? :)

Sorry, I ran out of time unfortunately :(

prost-build/src/lib.rs
Kh0088, line 1: ‏مرحبا أنا جاهز مرحبا

PR #659, opened 2022-05-27
Add enum variant to string mapping

vriesk:


vriesk:
@LucioFranco friendly ping :)

LucioFranco:
Sorry for the delay on this was out on vacation and still suffering from some jetlag. I took a cursory review but would like to spend some more time looking at this. Just wanted to let you know what the status was.

As for one thing, since this adds a method on the generated codegen types. I am leaning that we should just add this not as an opt-in but as a on by default without the config. This would also mean that `prost-types` would need to be updated and I would like to see that since it will show case some of the codegen.

vriesk:
Ack, I can remove the config guard. Do you have some specific method to regenerate the `prost-types` or is it just output of `prost-build` run?

vriesk:
> 

Done.

vriesk:
@LucioFranco I have just noticed that `clippy` whines about the `to_str_name()` name:

```
warning: methods with the following characteristics: (`to_*` and `self` type is `Copy`) usually take `self` by value
```

Maybe let's change it to `as_str_name()` then before this gets released? WDYT?

LucioFranco:
@vriesk good idea, lets do that.

PR #668, opened 2022-06-19
fix case on type attribute in prost-build documentation

zkrzyzanowski:
This fixes a small error in the `prost-build` documentation where an example for a `serde` attribute used `kabob-case` instad of `snake_case`. Correct syntax confirmed in their docs [here](https://serde.rs/container-attrs.html#rename_all)


LucioFranco (approved):
Thanks! The CI failures are related to a MSRV change in our deps so once https://github.com/tokio-rs/prost/pull/669 gets merged you can rebase and we can re-run CI and merge.

PR #678, opened 2022-07-05
chore: fix new clippy lint in generated code

thibault-martinez:
With rust version 1.62.0, the clippy lint [derive_partial_eq_without_eq](https://rust-lang.github.io/rust-clippy/master/index.html#derive_partial_eq_without_eq) was introduced. It makes the code generated by `prost-build` to fail clippy checks as it derives `PartialEq` but not `Eq`. This PR adds the `Eq` derive.
A patch release including this would be greatly appreciated as CI checks are failing for a lot of projects 🙏🏻

### DIFF
--- a/prost-build/src/code_generator.rs
+++ b/prost-build/src/code_generator.rs
@@ -184,7 +184,7 @@ impl<'a> CodeGenerator<'a> {
         self.append_type_attributes(&fq_message_name);
         self.push_indent();
         self.buf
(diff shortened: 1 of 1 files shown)


thibault-martinez:
Hmm yeah I forgot about floats 🤦🏻‍♂️
I guess we'll have to ignore the lint, sorry for the noise.

LucioFranco:
Clippy can be too aggressive sometimes (and wrong at times too!)

PR #686, opened 2022-07-21
Permit overflow with wrapping.

### LINKED ISSUE #682: Malformed `Timestamp` leads to Debug Assert Crash  'attempt to negate with overflow' in `impl TryFrom<Timestamp> for std::time::SystemTime`
Hi. Thank you for a very useful library.

I believe this is a bug: as of 25ab1cc77cba01568d7a72b01f7725f13b8cc095, if you instantiate the proptest `check_timestamp_roundtrip_via_system_time` in `prost-types/lib.rs` with out of range-values, you can get `SystemTime::try_from(timestamp.clone())` to crash with arithemetic overflow checks. Even if the input is invalid, I believe TryFrom should return an error value instead of crashing on the spot. Especially b/c this Timestamp struct is public including the internal fields.

Replication (tested on x86 64 bit ubuntu 20.04 and M1 Mac ): Paste the following into `prost-types/lib.rs` at[ this location](https://github.com/tokio-rs/prost/blob/25ab1cc77cba01568d7a72b01f7725f13b8cc095/prost-types/src/lib.rs#L404)
```rust
    #[test]
    fn check_timestamp_roundtrip_via_system_time_oveflow() {
        use std::time::SystemTime;
        let seconds : i64 = -9223372036854775808;
        let nanos : i32 = 999997508;

        let mut timestamp = Timestamp { seconds, nanos };
        timestamp.normalize();
        if let Ok(system_time) = SystemTime::try_from(timestamp.clone()) {
            assert_eq!(Timestamp::from(system_time), timestamp);
        }
    }
```
`cargo test --package prost-types --lib -- tests::check_timestamp_roundtrip_via_system_time_oveflow --exact`  yields
```
thread 'tests::check_timestamp_roundtrip_via_system_time_oveflow' panicked at 'attempt to negate with overflow', prost-types/src/lib.rs:372:56
```

YoshikiTakashima:
Fixes #682 by allowing i64 wrapping during negation. 

This allows `seconds = i64::MIN` to pass with debug checks enabled.

### DIFF
--- a/prost-types/src/lib.rs
+++ b/prost-types/src/lib.rs
@@ -369,7 +369,11 @@ impl TryFrom<Timestamp> for std::time::SystemTime {
             std::time::UNIX_EPOCH.checked_add(time::Duration::from_secs(timestamp.seconds as u64))
         } else {
             std::time::UNIX_EPOCH.checked_sub(time::Duration::from_secs(
-                (-std::num::Wrapping(timestamp.seconds)).0 as u64,
+                timestamp
+                    .seconds
+                    .checked_neg()
+                    .ok_or(TimestampError::OutOfSystemRange(timestamp.clone()))?
+                    as u64,
             ))
         };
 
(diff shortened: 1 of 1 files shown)


LucioFranco (commented):
I wonder if we should actually return an error rather than wrapping. What do you think?

YoshikiTakashima:
So with `seconds = i64::min`, the function already returns `TimestampError::OutOfSystemRange(orig_timestamp)`. 

I also made one that makes this error return more explicit, although it's longer.

LucioFranco (approved):
LGTM! Thanks!

PR #688, opened 2022-07-24
Move edition to 2021

### LINKED ISSUE #553: Fix clippy warnings introduced with 2021 edition

### LINKED ISSUE #595: Fix warnings with 1.51
There were a few warnings introduced with rust 1.51 we should fix these.

chetaldrich:
Looks like since I've last looked at the repo the update to
[MSRV 1.56 happened](https://github.com/tokio-rs/prost/pull/669) and we don't have any clippy errors on the 2021
edition. Therefore, I _think_ we can just up the edition now. 
If there's a good reason not to do so let me know and I can close this. 

I've also set the MSRV explicitly in the Cargo.toml so that users that happen to have an older rust version set (like I did when I was testing 1.51 at some point) get an error telling them they're on the wrong version instead of abstruse compiler errors. 

Closes #553, also closes #595 since 1.51 will no longer be supported.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,7 @@ readme = "README.md"
 description = "A Protocol Buffers implementation for the Rust Language."
 keywords = ["protobuf", "serialization"]
 categories = ["encoding"]
-edition = "2018"
+edition = "2021"
 
 [workspace]
 members = [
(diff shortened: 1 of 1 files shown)


LucioFranco (approved):
Thanks!

chetaldrich:
no problem, thanks for taking a look!

PR #694, opened 2022-08-02
Don't add extra space before RustDoc comments

### LINKED ISSUE #693: prost-build v0.11.0 adds an extra space before RustDoc comment text
It looks like something in how `prost-build` generates RustDoc comments has changed slightly between v0.10 and v0.11. There's now a single extra space between the `///` and the rest of the comment, which wasn't there previously.

This isn't a big deal or anything, but I noticed it increases the size of the diff for a prost upgrade significantly. See e.g. https://github.com/linkerd/linkerd2-proxy-api/pull/137/files#diff-e376e65ab87f7c7679472651d00eba584b6c134cd5ccf8a3b68f09e699944894

sd2k:
In v0.11.0 the `sanitize_line` function was updated to add a space to
the beginning of non-empty RustDoc comments, but often those comments
already contain a space. This commit updates the check to ensure the
extra space is only added if it doesn't already exist.

Fixes #693.

### DIFF
--- a/prost-build/src/ast.rs
+++ b/prost-build/src/ast.rs
@@ -138,6 +138,27 @@ impl Comments {
         }
     }
 
+    /// Checks whether a RustDoc line should be indented.
+    ///
+    /// Lines should be indented if:
+    /// - they are non-empty, AND
+    ///   - they don't already start with a space
+    ///     OR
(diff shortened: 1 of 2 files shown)


LucioFranco (approved):
LGTM thanks!

PR #695, opened 2022-08-02
fixed wrongly escape square brackets in code block

### LINKED ISSUE #683: square brackets are wrongly escaped in code blocks
```protobuf
// ```rust
// let a = vec![1];
// ```
message Test{}
```

Will become
```rust
/// ```rust
/// let a = vec!\[1\];
/// ```
#[derive(prost_helpers::AnyPB)]
pub struct Test {
}
```

trasua:
fixed #683 

I changed the function `get_lines` (`default` feature) to use `pulldown_cmark` to fix the bug and for better escaping square brackets outside of code blocks since the old `RULE_BRACKETS` regex didn't work correctly.

For example, the `RULE_BRACKETS` regex will change the URL `[google](https://google.com)` to `\[google\](https://google.com)` thus the URL won't be rendered as link in `cargo doc`.


prost-build/Cargo.toml
```diff
@@ -15,7 +15,7 @@ edition = "2021"
 rust-version = "1.56"
 
 [features]
-default = []
+default = ["pulldown-cmark", "pulldown-cmark-to-cmark"]
```
LucioFranco, line 18: Can you explain more why this should be enabled by default? I originally wanted this to be an opt-in feature so that users are not forced to pull this crate in.

trasua: oh because I need to use `pulldown-cmark` and `pulldown-cmark-to-cmark` in `get_lines()` function with `default` feature. This means both `default` and `cleanup-markdown` features depend on `pulldown-cmark` and `pulldown-cmark-to-cmark` crates now. Is that ok?

LucioFranco: Would it make sense to just include this under `cleanup-markdown`?

trasua: You mean to just include `pulldown-cmark` and `pulldown-cmark-to-cmark` under `cleanup-markdown`?

LucioFranco: Right, leave the futures as is and add this clean up as part of `cleanup-markdown`?

trasua: This _incorrectly escape square brackets in code blocks_ bug only appears in `default` feature.
So if you want me to add this fix as part of `cleanup-markdown` feature then there's no need for this PR, because `cleanup-markdown` has already worked correctly.

LucioFranco: I am bit confused if that feature fixes it then why is there more code here. Maybe I am missing something and didn't look hard enough. 

I could be convinced to either 1) make this feature default 2) implement the specific parts of this feature internally to prost without the xtra dependency.

prost-build/src/ast.rs
```diff
@@ -329,27 +345,34 @@ mod tests {
             },
             TestCases {
                 name: "invalid_start_bracket",
-                input: "foo [= baz".to_string(),
-                expected: "/// foo [= baz\n".to_string(),
+                input: "foo \\[= baz".to_string(),
+                expected: "/// foo \\[= baz\n".to_string(),
```
caspermeijn, line 349: I think this should be a new test case and leave the original in place

prost-build/src/ast.rs
```diff
@@ -161,15 +180,12 @@ impl Comments {
 
     /// Sanitizes the line for rustdoc by performing the following operations:
     ///     - escape urls as <http://foo.com>
-    ///     - escape `[` & `]`
     fn sanitize_line(line: &str) -> String {
         lazy_static! {
             static ref RULE_URL: Regex = Regex::new(r"https?://[^\s)]+").unwrap();
-            static ref RULE_BRACKETS: Regex = Regex::new(r"(\[)(\S+)(])").unwrap();
```
caspermeijn, line 168: Why was this removed? Is it replaced by `pulldown-cmark`?

caspermeijn (changes requested):
I don't understand what problem you are trying to fix.
- Can you give an example to reproduce the problem?
-  Does the problem only appear when `cleanup-markdown` is disabled?

PR #701, opened 2022-08-12
Add support for enum and oneof field deprecated.

sonder-joker:
I notice it's not support enum and oneof field deprecated.


prost-build/src/code_generator.rs
```diff
@@ -670,6 +679,10 @@ impl<'a> CodeGenerator<'a> {
         self.depth += 1;
 
         for variant in variant_mappings.iter() {
+            if variant.deprecated {
+                self.push_indent();
+                self.buf.push_str("#[allow(deprecated)]\n");
```
caspermeijn, line 684: Why is the `allow` statement now needed and not is previous version?

prost-derive/src/lib.rs
```diff
@@ -304,6 +305,7 @@ fn try_enumeration(input: TokenStream) -> Result<TokenStream, Error> {
             }
         }
 
+        #[allow(deprecated)]
```
caspermeijn, line 308: Will this hide deprecation warnings in non deprecated fields?

tests/src/deprecated_field.rs
```diff
@@ -11,7 +11,12 @@ fn test_warns_when_using_fields_with_deprecated_field() {
     let message = deprecated_field::Test {
         not_outdated: ".ogg".to_string(),
         outdated: ".wav".to_string(),
+        test3: Some(deprecated_field::test::Test3::OneofOutdated(
+            ".mp3".to_string(),
+        )),
     };
+    #[allow(deprecated)]
+    let enum_ = deprecated_field::Test2::Outdated;
```
caspermeijn, line 19: I would like to see an additional test: generate a proto with only deprecated fields and write that to file. This has two purposes:
- It adds a example of all `deprecated` statements to the PR
- It actually tests that the `deprecated` flag is added to the fields.

The repo has examples for a test writing to file and asserting nothing changes: `test_generate_no_empty_outputs`

caspermeijn (changes requested):
I like this idea. We should support deprecated on all fields.

Please rebase the branch to enable CI

prost-build/src/code_generator.rs
sonder-joker: This direct call will raise deprecated warning for every call of `as_str_name`, so we need a `#[allow(deprecated)]` for normal call.
For deprecated enum, direct call `as_str_name` will still raise warning.

prost-derive/src/lib.rs
sonder-joker: fix.

prost-derive/src/lib.rs
```diff
@@ -460,6 +467,7 @@ fn try_oneof(input: TokenStream) -> Result<TokenStream, Error> {
     });
 
     let expanded = quote! {
+        #[allow(deprecated)]
```
caspermeijn, line 470: Why is this not conditional, but the other `allow` statements are?

tests/src/deprecated_field.rs
caspermeijn: Thanks for creating that test

caspermeijn (changes requested):
CI is failing. I believe you can fix that by running this locally and committing the result: `cargo test -p prost-build -p prost-derive -p prost-types --all-targets --no-default-features`

caspermeijn:
Solved via #1316

PR #703, opened 2022-08-12
Remove notes from prost-build about building protoc

jszwedko:
This is no longer supported as of v0.11. See ac57645 for its removal.

Signed-off-by: Jesse Szwedko <jesse@szwedko.me>


LucioFranco (approved):
Good catch! Thanks

PR #706, opened 2022-08-13
Tuple struct support

### LINKED ISSUE #56: Tuple structs with enumerations don't work
It seems tuple structs are broken when they contain an enumeration. I hit it when implementing tests for https://github.com/danburkert/prost/pull/54, but I can reproduce on master:

```rust
/// A special case with a tuple struct
#[test]
fn tuple_struct() {
    #[derive(Clone, PartialEq, Message, Debug)]
    struct NewType(
        #[prost(int32, tag="1")]
        pub i32,
        #[prost(enumeration="BasicEnumeration", tag="5")]
        pub i32,
    );
}
```

The error is:

```
   Compiling conformance v0.0.0 (file:///home/vorner/prog/prost/conformance)
error: expected identifier, found `1`
   --> tests/src/lib.rs:152:32
    |
152 |     #[derive(Clone, PartialEq, Message, Debug)]
```

If I comment the second field out,it compiles. I had no success yet pinpointing where exactly it happens, because cargo expand gives me the same error instead of the (broken) source code. 

As the prost-derive code contains code to handle tuple structs, I guess they are supposed to work.

ldm0:
Fixes #56 
Remove some(4) outdated TODOs by the way.


prost-derive/src/lib.rs
```diff
@@ -153,11 +167,7 @@ fn try_message(input: TokenStream) -> Result<TokenStream, Error> {
 
     let debugs = unsorted_fields.iter().map(|&(ref field_ident, ref field)| {
         let wrapper = field.debug(quote!(self.#field_ident));
-        let call = if is_struct {
-            quote!(builder.field(stringify!(#field_ident), &wrapper))
```
ldm0, line 157: `DebugTuple::field` accepts only one argument.

prost-derive/src/field/scalar.rs
```diff
@@ -267,13 +267,17 @@ impl Field {
...
     /// Returns methods to embed in the message.
-    pub fn methods(&self, ident: &Ident) -> Option<TokenStream> {
+    pub fn methods(&self, ident: &TokenStream) -> Option<TokenStream> {
         let mut ident_str = ident.to_string();
         if ident_str.starts_with("r#") {
             ident_str = ident_str[2..].to_owned();
         }
 
         if let Ty::Enumeration(ref ty) = self.ty {
+            let get = match syn::parse_str::<Ident>(&ident_str) {
```
ldm0, line 277: Add a fallback for tuple struct since "0" cannot be a function name.

LucioFranco (approved):
Sorry for the delay on review but I finally had some time to digest the macro code and this LGTM! Thanks!

Issue #661, opened 2022-05-31
clippy::derive_partial_eq_without_eq violated on nightly


nstinus:
Hi,

Given a simple message, prost violates derive_partial_eq_without_eq on nightly as shown in https://github.com/nstinus/prost-nightly.

Output is:
```
  |
1 | #[derive(Clone, PartialEq, ::prost::Message)]
  |                 ^^^^^^^^^ help: consider deriving `Eq` as well: `PartialEq, Eq`
  |
  = note: `-D clippy::derive-partial-eq-without-eq` implied by `-D clippy::all`
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#derive_partial_eq_without_eq

error: could not compile `prost-nightly` due to previous error
warning: build failed, waiting for other jobs to finish...
error: could not compile `prost-nightly` due to previous error
```

LucioFranco:
Since this is just a clippy lint and not an error from rustc and in addition, to support this working will require additional logic to prost-build. I don't think its worth adding this support now and the correct solution will be to ignore the warning.

nstinus:
In my case, since `clippy::all` is routinely enabled by default, that means disabling the lint wherever this crate is used. Not the end of the world, but it would be much better if this lib could avoid triggering it altogether. I don't feel like it is particularly exotic. Now, it may be difficult to fix but that is a different problem.

Can you confirm that your recommendation is for client code to do something like that?
```
#[allow(clippy::derive_partial_eq_without_eq)]

pub mod pb {
    tonic::include_proto!("dummy");
}
```

LucioFranco:
I think for things like codegen is quite normal to add allow's. Tonic does this for some built in rust warnings https://github.com/hyperium/tonic/blob/master/tonic-build/src/server.rs#L62


I would actually apply the allow just to the mod like so. 

```rust
pub mod pb {
    #![allow(clippy::derive_partial_eq_without_eq)]

    tonic::include_proto!("dummy");
}
```

jialez0:
Hi all, I tried to add `#![allow(clippy::derive_partial_eq_without_eq)]` to my code in this way:
```
pub mod pb {
    #![allow(clippy::derive_partial_eq_without_eq)]
    tonic::include_proto!("dummy");
}
```

It can pass the clippy lint check before `rustc 1.63.0 (4b91a6ea7 2022-08-08)`, but will still failed of `rustc 1.63.0 (4b91a6ea7 2022-08-08)`...

[...]

PR #710, opened 2022-08-23
Fix: clippy lints 1.63

apfitzge:
# Problem
We recently migrated a project to rust 1.63, which introduced some new clippy lints. Some are popping up in the code generated by prost: https://rust-lang.github.io/rust-clippy/master/index.html#derive_partial_eq_without_eq

# Summary of Changes
* Added Eq derive to the generated code for prost messages.
* Fixed internal prost 1.63 clippy lint: borrow_deref_ref

### DIFF
--- a/README.md
+++ b/README.md
@@ -323,7 +323,7 @@ message AddressBook {
 and the generated Rust code (`tutorial.rs`):
 
 ```rust,ignore
-#[derive(Clone, PartialEq, ::prost::Message)]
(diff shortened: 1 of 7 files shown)


LucioFranco:
https://github.com/tokio-rs/prost/issues/661 for adding the clippy lint (you can easily add this to your own project) and if we want equal we will need to follow what C++ does by providing an equality check that is done using the FDS

PR #711, opened 2022-08-24
Suggest how to download protoc in error message

adamchalmers:


LucioFranco (approved):
Seems like its just failing MSRV check but otherwise LGTM!! Thanks!

adamchalmers:
Oh OK, I always use latest rust so I'm not used to thinking about MSRV. I changed the format so it doesn't use "{named_arg_syntax}", it should pass now...

adamchalmers:
OK yep, that fixed it. Mergable now :)

LucioFranco (approved):

LucioFranco:
Thanks!

Issue #724, opened 2022-09-15
Protos that import "google/protobuf/any.proto" generate an extra empty file 


trevoranderson:
Example proto file:
```
syntax = "proto3";
import "google/protobuf/any.proto";

package blob;

// A config object containing a single config and its value
message TypedBlob {
  string key = 1;
  oneof val {
    bool bool_val = 3;
    int64 int_val = 4;
    string  string_val = 5;
    google.protobuf.Any proto_val = 6;
  }
}
```
Expected output: a `blob.rs` file using `::prost_types::Any`.
Actual output: that + an additional empty file called `google.protobuf.rs`

I'm generating the protos and checking them into my source tree, not running as part of `build.rs`. It's not a problem correctness-wise, but slightly annoying to have the clutter of empty files. Is there a workaround for this? I experimented with `extern_path` but that did not seem to matter.

LucioFranco:
Yeah, its a known issue, https://github.com/tokio-rs/prost/issues/228 tried to fix it but I had to revert the fix PR since it broke many other layers for some reason. If you have the time feel free to take a crack at a it.

trevoranderson:
@LucioFranco Made an attempt. I'm not overly familiar with the project so definitely open to feedback from someone who knows more, but it seemed to fix the issues I was having as well as preventing the issue of the previous fix.

Issue #228, opened 2019-10-11
prost-build generates empty files


Timmmm:
Here is my `build.rs`:

```
fn main() {
	prost_build::compile_protos(
		&["src/plugin.proto"],
		&["src/"],
	).expect("protobuf compilation failed");
}
```

And `plugin.proto` is copy/pasted from [this page](https://developers.google.com/protocol-buffers/docs/reference/cpp/google.protobuf.compiler.plugin.pb).

When I `cargo build` it generates no errors, and it does generate these two files:

* `myproject/target/debug/build/myproject-9d87c2b89447dc16/out/google.protobuf.compiler.rs`
* `myproject/target/debug/build/myproject-9d87c2b89447dc16/out/google.protobuf.rs`

However they are both empty files. Any ideas? I've used prost-build successfully on the same machine in a different project with a proto file I wrote myself (using proto3 syntax).

Timmmm:
I think it is an issue with `.`s in the package name.

Timmmm:
Ok the behaviour seems to be that if the package name contains any `.`'s then it is treated as an empty file.

To get the `plugin.proto` to work I had to also add `descriptor.proto` (both copied from this repo) and then change the package names to the same name (that didn't contain `.`).

danburkert:
`prost-build` has special behavior to recognize when you are attempting to compile any protobuf 'well known types', and skips compiling them. This is because the `prost-types` crate already contains pre-compiled versions of these types.

If you'd like to turn off this special casing, use [`Config::compile_well_known_types`](https://docs.rs/prost-build/0.5.0/prost_build/struct.Config.html#method.compile_well_known_types).

danburkert:
but in case it's not clear, I'd suggest _not_ doing that, and instead using the `prost-types` crate.  It exists so that you don't get a proliferation of unique well known types across a bunch of different invocations to `compile_protos`.

WhoisDavid:
When using `compile_protos` with `prost-types`, shouldn't the empty `google.protobuf.rs` not be created though?

mieubrisse:
Hey @danburkert , wanted to ping on this as I'm getting the same issue as @WhoisDavid  - even when using the `prost-types` crate I get an empty `google.protobuf.rs` file

danburkert:
[...]

PR #726, opened 2022-09-28
Ignore empty codegen modules in outputs

trevoranderson:
References:
Original report: https://github.com/tokio-rs/prost/issues/228
My report: https://github.com/tokio-rs/prost/issues/724
Partial fix in: https://github.com/tokio-rs/prost/pull/605
Issue with fix: https://github.com/hyperium/tonic/issues/964
Reverted in: https://github.com/tokio-rs/prost/pull/639

After digging in, the issue with the first fix is that while it prevented the file from being generated, it did not prevent the file from being referenced in the "include" file because both locations iterate the module hash. My proposed solution is to instead avoid adding the empty module to the list at all. Then the file does not get output and the include file does not have an entry that would otherwise be referencing an empty file.

Also did a small refactor pass on the surrounding loop to destructure the tuple instead of using tuple.0 syntax wherever it is used.

Tested by reproducing the original issue using `tonic-build` with the original behavior working then reintroduced the bug, then fixed it with my changes. Manually verified output files and included them in a new rust project to make sure they worked and verified that the extra file is not present in file system or the include file. Also ran cargo test on prost

### DIFF
--- a/prost-build/src/lib.rs
+++ b/prost-build/src/lib.rs
@@ -127,13 +127,6 @@
 //!
 //! [`protobuf-src`]: https://docs.rs/protobuf-src
 
-mod ast;
-mod code_generator;
-mod extern_paths;
-mod ident;
-mod message_graph;
-mod path;
-
 use std::collections::HashMap;
 use std::default;
 use std::env;
@@ -146,6 +139,7 @@ use std::path::{Path, PathBuf};
 use std::process::Command;
 
 use log::trace;
+
 use prost::Message;
 use prost_types::{FileDescriptorProto, FileDescriptorSet};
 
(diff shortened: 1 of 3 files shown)


LucioFranco:
@trevoranderson thank you for doing this! Would it be possible to add a test of the original behavior that was fixed with this so we can make sure we don't regress in the future?

trevoranderson:
Made an attempt at that too. I used an proto adapted from the hyperium issue report which reproduces the issue without my fix and works with the fix applied.

prost-build/src/lib.rs
```diff
@@ -156,6 +150,13 @@ use crate::ident::to_snake;
 use crate::message_graph::MessageGraph;
 use crate::path::PathMap;
 
+mod ast;
```
LucioFranco, line 153: Why this change?

LucioFranco (approved):
LGTM thanks!

PR #728, opened 2022-10-09
Add snazzy module to example

mvanotti:
Modify the example code for prost-build so that it uses the right module structure.

Otherwise this might cause issues if the proto imports other protos.

-----

However, it would be better if the example used directly the Config::include_file method, to make it easier to discover :). I can also change that if needed.


LucioFranco (approved):

LucioFranco:
Thanks!

PR #729, opened 2022-10-11
Allow generated `Message` derive path to be changed

Zagitta:
I have a use case where I'm building a wrapper around `prost-build` in combination with `prost-reflect` to extract protobuf custom options and do some magic.
However, since this wrapper, let's call it `my-build`, will be used from a large number of different crates it would be ideal to have the `my-core` crate re-export `prost` to simplify upgrades down the line.

In order to facilitate this use case I've added a simple configuration option which would allow me to change the default `#[derive(::prost::Message)]` to `#[derive(::my_crate::prost::Message)]`.

Hopefully, it's a simple enough change that it won't be too controversial 😄 

Please let me know if there's anything you'd like me to change.


LucioFranco:
Yeah, we should cover those as well

LucioFranco:
Here is the follow up PR https://github.com/tokio-rs/prost/pull/740

PR #730, opened 2022-10-12
add `Message` support for heapless::{Vec,String}

ryankurte:
hey there,

this PR adds preliminary support for `heapless::String` and `heapless::Vec` behind the `heapless` feature, for those of us running prost without an allocator ^_^

the logic is basically a clone of the existing `alloc::string::String` and `alloc::vec::Vec` implementations because i couldn't see an easy way of genericising across these.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,12 +43,14 @@ bench = false
 
 [features]
 default = ["prost-derive", "std"]
+heapless = ["dep:heapless"]
 no-recursion-limit = []
 std = []
 
 [dependencies]
 bytes = { version = "1", default-features = false }
 prost-derive = { version = "0.11.0", path = "prost-derive", optional = true }
+heapless = { version = "0.7.16", optional = true }
 
 [dev-dependencies]
 criterion = "0.3"
--- a/src/encoding.rs
+++ b/src/encoding.rs
@@ -802,6 +802,7 @@ pub mod string {
         encode_varint(value.len() as u64, buf);
         buf.put_slice(value.as_bytes());
     }
+
     pub fn merge<B>(
         wire_type: WireType,
         value: &mut String,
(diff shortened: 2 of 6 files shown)


LucioFranco:
Hi! I am a bit unsure about adding support for heapless (and countless other things like this which change the types). I don't know if adding it as a feature is the best route but maybe we want to consider using proto3 options and having some sort of plugin system for prost that would let external crates decorate. That way we don't need to include a lot of other crate specific code but just the abstractions for adding that codegen.

ryankurte:
thanks for the response! i guess there's a wider question of whether y'all are open to / interested in supporting use in `no_std`?

> I am a bit unsure about adding support for heapless (and countless other things like this which change the types).

`heapless` is pretty pivotal in the embedded / no_std / no_alloc ecosystem, up there with `bytes` elsewhere imo. 
i don't think there are any equal equivalents / alternatives, and a naive impl wouldn't necessarily be correct for embedded.

> I don't know if adding it as a feature is the best route but maybe we want to consider using proto3 options and having some sort of plugin system for prost that would let external crates decorate. 

i think reworking the core traits to support external definition of storage types + a plugin system for the builder/generator would be great, with the caveat that all types would need to use these APIs (so `alloc` types could be packaged separately from the [core traits]((https://docs.rs/prost/latest/prost/trait.Message.html#method.encode_to_vec)) ). this would however seem to require a reasonably significant refactoring of both internal and external interfaces...
 
for related context the [nanopb](https://jpa.kapsi.fi/nanopb/docs/reference.html#defining-the-options-in-the-.proto-file) generator uses annotations to clue in the builder about type constraints, something like this would be a great eventual outcome.

LucioFranco:
prost already has no_std support (but I actually don't know what this really means in terms of prost since I didn't write this and it was here before I started to maintain the crate). 

Your reasoning is very valid and yeah it would be a big refactor but I think it could be wroth it. That said, do you think you could get the support you wanted via an external crate like `prost-heapless` that newtypes the heapless types and adds a `MessageExt` trait with the encode_heapless? This may work to just get the ground moving. I think what we may want to do then is provide an option that allows you to rename the path of the types so that you can swap `String` for `prost_heapless::String` and it would just work?

ryankurte:
> prost already has no_std support

ahh, sorry i'm conflating no [`std`](https://github.com/tokio-rs/prost/blob/master/src/lib.rs#L2) and no [alloc](https://github.com/tokio-rs/prost/blob/master/src/lib.rs#L7)... prost can run without `std` but not without `alloc` which is sortof half way there.

> That said, do you think you could get the support you wanted via an external crate like prost-heapless that newtypes the heapless types and adds a MessageExt trait with the encode_heapless?

ideally this could also go into `heapless`, but for now the dep on `alloc` is a dealbreaker so a third party one might be the way.

> I think what we may want to do then is provide an option that allows you to rename the path of the types so that you can swap String for prost_heapless::String and it would just work?

nearly, but because the heapless types are stack allocated / sized they'd also need to pick up an attribute for max length (`String` becomes `String<const N: usize>` etc.)

LucioFranco:
Got it yeah, if you can do it in a separate crate I think that makes sense for now.

ryankurte:
> if you can do it in a separate crate I think that makes sense for now.

related to #782, it seems you actually -can't- implement array types externally at the moment..?

the `#[prost(bytes, tag=1]` attribute seems to manually support `Bytes` and `Vec` types, and expects any wrapper type to implement `BytesAdapter`, which depends on `sealed::BytesAdapter` and is thus not externally implementable.

you can externally implement and use `prost::Message`, but a byte array -isn't- a message so this breaks the encoding.

so it seems it's either land something here or don't support heapless / generics until things are refactored.

LucioFranco:
> so it seems it's either land something here or don't support heapless / generics until things are refactored.

I think its better to wait until things are refactored, I think that would provide a better end state.

PR #738, opened 2022-10-27
fix disable_comments comment

bouzuya:


prost-build/src/lib.rs
```diff
@@ -490,7 +490,7 @@ impl Config {
     ///
     /// ```rust,ignore
```
LucioFranco, line 491: I wonder if we want `no_run` here instead so we catch this on CI?

LucioFranco (changes requested):
LGTM, I just have one request

prost-build/src/lib.rs
bouzuya: Thanks for the review. Fixed to use no_run instead of ignore. 

https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#attributes

LucioFranco (approved):
awesome, thanks!!!!

PR #748, opened 2022-11-08
implement Enum::from_str_name(&str)

timurnkey:
A method `Enum::as_str_name() -> &str` already exists. This provides the opposite functionality.


LucioFranco:
Does this exist in other protobuf implementations?

timurnkey:
I'm not sure. Our use case is that we're allowing end users to make JSON requests using strings instead of ints for enum fields. While we can convert the string to the enum by matching against all the `enum::as_str_name()` variants, it'd be a lot nicer to just have this built in.

timurnkey:
Just realized that this wouldn't help me. The logic would need to live in `prost-derive` instead, so I'll close this ticket.

PR #760, opened 2022-11-21
Support custom recursion limits at runtime

seanlinsley:
We decode protobuf from both trusted and untrusted sources, so the `no-recursion-limit` feature wasn't a good fit for our workload.

As a followup to #186, this PR introduces a `decode_with_context` function that allows users to customize the `DecodeContext`. I chose a generic approach (instead of `decode_with_recursion_limit`) in case there are additional decode settings that might be added in the future.

Thoughts on the best place to document this, and whether any additional tests are required?

Todo:
- [ ] update the inline docs on `DecodeContext`

### DIFF
--- a/src/encoding.rs
+++ b/src/encoding.rs
@@ -199,7 +199,7 @@ pub struct DecodeContext {
     /// customized. The recursion limit can be ignored by building the Prost
     /// crate with the `no-recursion-limit` feature.
     #[cfg(not(feature = "no-recursion-limit"))]
-    recurse_count: u32,
+    pub recurse_count: u32,
 }
 
 #[cfg(not(feature = "no-recursion-limit"))]
--- a/src/message.rs
+++ b/src/message.rs
@@ -116,6 +116,20 @@ pub trait Message: Debug + Send + Sync {
         Self::merge(&mut message, &mut buf).map(|_| message)
     }
 
+    /// Decodes an instance of the message from a buffer.
+    ///
+    /// The entire buffer will be consumed.
+    ///
+    /// This is an alternative to `decode`, which allows users to specify a `DecodeContext`.
+    fn decode_with_context<B>(mut buf: B, ctx: DecodeContext) -> Result<Self, DecodeError>
+    where
+        B: Buf,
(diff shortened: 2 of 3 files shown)


seanlinsley:
CC @nrc @danburkert

seanlinsley:
A build time option isn't a good fit for our use case. We deal with both trusted and untrusted protobuf sources in the same application so it would be preferable if this option could be set at runtime.

That said: could we find a recursion limit that is high enough to work with both without risking a stack overflow? Probably.

Could you say more about how the procedural macro would work? I haven't written one previously.

seanlinsley:
Another option is to add a crate feature that sets the recursion limit to a higher value. For example, 1,000.

LucioFranco:
> Another option is to add a crate feature that sets the recursion limit to a higher value. For example, 1,000.

Yeah, though I am not sure how to make this work in an additive fashion. Maybe adding a method to the Message trait is the only way which is unfortunate.

seanlinsley:
Here's a first pass at moving it to a macro attribute that's configured through prost-build:

https://github.com/tokio-rs/prost/compare/master...pganalyze:prost:recursion-limit-macro

```rs
prost_build.type_attribute("YourStruct", "#[RecursionLimit(1000)]"); // right now
prost_build.recurstion_limit("YourStruct", 1000); // maybe in the future
```

Thoughts on that as an alternative API? Note that branch isn't complete yet.

LucioFranco:
Oh this approach is really smart!!! I didn't think about embedding a const in the impl and using that!!! I like it! I would maybe have the codegen either use the default impl if the recursion limit param is not supplied. On-top of that we could maybe add a specific builder item to prost build's config such that you don't need to use type_attribute but I think that works for now. I like this approach.

seanlinsley:
Unfortunately it looks like an associated const isn't going to work. The best we could do is implement a `Message::recursion_limit()`

Screenshot 2022-12-15 at 3 30 43 PM

LucioFranco:
I think we can close this one in favor of #785

PR #774, opened 2022-12-06
Expose from_str_name() for generated enums

segfault87:
Hi.

This PR enables `prost-build` codegen to include `from_str_name()` for each generated enumerations.

Currently there is no way to create Protobuf enums from names. Since [Protobuf JSON format](https://developers.google.com/protocol-buffers/docs/proto3#json) states enum value can be represented as string value, I think it is necessary when it comes to adding support for JSON serialization/deserialization.


grant0417:
@dependabot squash and merge

PR #780, opened 2022-12-10
chore: Fix clippy lint

tottoto:
Fixes clippy lint.

### DIFF
--- a/conformance/src/main.rs
+++ b/conformance/src/main.rs
@@ -17,7 +17,7 @@ fn main() -> io::Result<()> {
     loop {
         bytes.resize(4, 0);
 
-        if io::stdin().read_exact(&mut *bytes).is_err() {
+        if io::stdin().read_exact(&mut bytes).is_err() {
             // No more test cases.
             return Ok(());
         }
@@ -25,7 +25,7 @@ fn main() -> io::Result<()> {
         let len = bytes.as_slice().get_u32_le() as usize;
 
         bytes.resize(len, 0);
-        io::stdin().read_exact(&mut *bytes)?;
+        io::stdin().read_exact(&mut bytes)?;
 
         let result = match ConformanceRequest::decode(&*bytes) {
             Ok(request) => handle_request(request),
--- a/prost-build/src/lib.rs
+++ b/prost-build/src/lib.rs
@@ -1081,8 +1081,8 @@ impl Config {
 
     #[cfg(feature = "format")]
     fn fmt_modules(&mut self, modules: &mut HashMap<Module, String>) {
-        for (_, buf) in modules {
-            let file = syn::parse_file(&buf).unwrap();
+        for buf in modules.values_mut() {
+            let file = syn::parse_file(buf).unwrap();
             let formatted = prettyplease::unparse(&file);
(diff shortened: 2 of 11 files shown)


tottoto:
There are some unresolved warnings (these codes are at [v0.11.4](https://github.com/tokio-rs/prost/releases/tag/v0.11.4)). Clippy warns them as derive_partial_eq_without_eq. I'm not sure that how they should be handled.

## `WireType` in `src/encodeng.rs`

https://github.com/tokio-rs/prost/blob/d3ba4e8fbc0155d9640fa67c7382a6dd4920b082/src/encoding.rs#L267-L276

## `DurationError` in `prost-types/src/lib.rs`

https://github.com/tokio-rs/prost/blob/d3ba4e8fbc0155d9640fa67c7382a6dd4920b082/prost-types/src/lib.rs#L145-L165

## `TimeSampError` in `prost-types/src/lib.rs`

https://github.com/tokio-rs/prost/blob/d3ba4e8fbc0155d9640fa67c7382a6dd4920b082/prost-types/src/lib.rs#L318-L335

LucioFranco:
@tottoto we should handle them the same way that we handle the other by having an allow. I think that makes the most sense, what do you think?

tottoto:
I think `WireType` should implement Eq since it have to return true when all same variants ​​are compared (according to [check_wire_type](https://github.com/tokio-rs/prost/blob/d3ba4e8fbc0155d9640fa67c7382a6dd4920b082/src/encoding.rs#L344)).

I also considered whether the PartialEq are actually needed or not on error enums. It seem that they are implemented only for tests. In particular the errors enum do not have to be implemented as PartialEq ([like hyper's error](https://docs.rs/hyper/0.14.23/hyper/struct.Error.html)) imho. If prost does not intend to provide the PartialEq functionality to its users, we might be able to remove PartialEq from them (and implement PartialEq and Eq, or implement PartialEq and allow derive_partial_eq_without_eq on them by `#[cfg_attr(tests, ...)]`).

LucioFranco:
Yeah, I actually don't think I mind Eq on enums since that actually kinda makes sense.  Im cautious to add Eq to anytime stamp stuff since generally you shouldn't compare them like that. What do you think?

tottoto:
In my opinion, `WireType` should implement `Eq` because without `Eq` it can not assure that the same wire type check always return true. 

On the other hand, `DurationError` and `TimeStampError` should not do. We should handle error enum by pattern matching so maybe `PartialEq` is not also unnecessary imho. However, since removing `PartialEq` is a big breaking change and discussing it is also a too big in order to fix lint warnings, I'd like to leave them unchanged and add `allow` rules.

tottoto:
These are my proposal: https://github.com/tokio-rs/prost/pull/780/commits/e71d6f1c1b34d3f22ecc9591256476979ded75c0, https://github.com/tokio-rs/prost/pull/780/commits/bf2696ba077ddce2d018fab17b4a6f4c9f5c7516.

LucioFranco (approved):

LucioFranco:
Thanks for doing all these PRs!!! I really appreciate it!

PR #785, opened 2022-12-15
Support custom recursion limits at build time

seanlinsley:
This PR allows users to set a custom recursion limit for specific protobuf structures at build time:
```rs
let mut config = prost_build::Config::new();
config.recursion_limit("my_messages.MyMessageType", 1000);
```


src/message.rs
```diff
@@ -14,6 +14,13 @@ use crate::EncodeError;
 
 /// A Protocol Buffers message.
 pub trait Message: Debug + Send + Sync {
+    /// The recursion limit for decoding protobuf messages.
+    ///
+    /// Defaults to 100. Can be customized in your build.rs or by using the no-recursion-limit crate feature.
+    fn recursion_limit() -> u32
```
LucioFranco, line 20: instead of adding this which is a breaking change, can we just generate adhoc const that are outside the impl like `const MY_MESSAGE_NAME_RECURSION_LIMIT = xxx`. That would allow it to still be obj safe and wouldn't be a breaking change but just an addition to the codegen. We only add this const and use it if the attribute for recursion is there.

LucioFranco (changes requested):

src/message.rs
seanlinsley: If it's outside the impl, how would we know which const to use for any given message at runtime?

I'm not sure I understand how this is a breaking change, but if we can't change the functions in the `Message` trait, how about adding a new `RecursionLimit` trait? Then the prost-derive macro would generate:
```rs
let expanded = quote! {
    impl #impl_generics ::prost::RecursionLimit for #ident #ty_generics #where_clause {
        fn recursion_limit() -> u32 {
            #recursion_limit

Comment (LucioFranco) [reply src/message.rs:20]: Adding a new required method to a trait is a breaking change because anyone that has impl it manually will get a build failure saying they are missing a trait. We could make it have a default impl though which I think could work well. What do you think? The feature flag we have could just be used there to have it either emit the limit or the max?

Comment (seanlinsley) [reply src/message.rs:20]: I'm not sure what you mean by adding a default impl. What do you think about the `RecursionLimit` trait option?

Comment (seanlinsley) [reply src/message.rs:20]: Are there really that many prost users who manually implement the `Message` trait? I would expect there are very few projects who have such specialty needs that they would implement it themselves.

Comment (LucioFranco) [reply src/message.rs:20]: I would like to avoid introducing a new trait tbh. As for manual implementation of message that is a good question but it is something that is public and we can't just release a non breaking change with a breaking change to a trait.

Comment (seanlinsley) [reply src/message.rs:20]: As far as we know, no one manually implements this trait so we can treat it as purely macro-generated. As such, as long as we don't change the signature of `Message` functions that are public (that is, not `#[doc(hidden)]`), I don't think this should be treated as a breaking change.

Comment (seanlinsley) [reply src/message.rs:20]: After asking around, it seems that adding a trait function with a default implementation isn't considered a breaking change in the broader community. And that also makes sense in this context: if someone happened to be manually implementing this trait, it would fall back to the default function impl. I'll update this PR to do that.

https://twitter.com/steveklabnik/status/1613012472889331712

**Comment by seanlinsley** [comment]
Here's a downstream PR using this feature: https://github.com/pganalyze/pg_query.rs/pull/17

Note that to prevent a stack overflow from the recursion, I had to increase the stack size in a separate thread. I wonder if that should be documented here.

File: src/encoding.rs [hunk 2]
```diff
@@ -195,29 +195,25 @@ pub struct DecodeContext {
...
     /// the recursion limit.
     ///
-    /// The recursion limit is defined by `RECURSION_LIMIT` and cannot be
-    /// customized. The recursion limit can be ignored by building the Prost
-    /// crate with the `no-recursion-limit` feature.
+    /// It defaults to 100 and can be changed using `prost_build::recursion_limit`,
+    /// or it can be disabled entirely using the `no-recursion-limit` feature.
     #[cfg(not(feature = "no-recursion-limit"))]
-    recurse_count: u32,
+    pub recurse_count: u32,
```
LucioFranco, line 201: Should this be doc(hidden) so its not actually public?

src/message.rs
LucioFranco: Yeah, after thinking about this for a bit I agree. We can add this, thanks for diving deep on this!

LucioFranco (changes requested):
Overall LGTM! Some small suggestions and we can get this merged. Thanks for the patience I went on vacation in between :)

seanlinsley:
@LucioFranco this should be ready for re-review

prost-derive/src/lib.rs
```diff
@@ -23,6 +23,20 @@ fn try_message(input: TokenStream) -> Result<TokenStream, Error> {
 
     let ident = input.ident;
 
+    let recursion_limit: u32 = if let Some(attr) = input
+        .attrs
+        .iter()
+        .find(|attr| attr.path.is_ident("RecursionLimit"))
```
LucioFranco, line 29: I wonder if we should make this one camel case like serde does https://serde.rs/container-attrs.html

so `#[prost(recursion_limit = 5)]` etc

src/encoding.rs
```diff
@@ -195,29 +195,26 @@ pub struct DecodeContext {
     /// How many times we can recurse in the current decode stack before we hit
     /// the recursion limit.
     ///
-    /// The recursion limit is defined by `RECURSION_LIMIT` and cannot be
-    /// customized. The recursion limit can be ignored by building the Prost
-    /// crate with the `no-recursion-limit` feature.
+    /// It defaults to 100 and can be changed using `prost_build::recursion_limit`,
```
LucioFranco, line 198: So because this is hidden this doc won't actually be readable. So we need to make sure this is documented at the lib level of `prost`.

src/lib.rs
```diff
@@ -23,11 +23,6 @@ use bytes::{Buf, BufMut};
 
 use crate::encoding::{decode_varint, encode_varint, encoded_len_varint};
 
-// See `encoding::DecodeContext` for more info.
-// 100 is the default recursion limit in the C++ implementation.
-#[cfg(not(feature = "no-recursion-limit"))]
-const RECURSION_LIMIT: u32 = 100;
```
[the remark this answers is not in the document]
LucioFranco: I actually kinda like having this as a constant. I wonder if we could just make a `constants` module that contains just this one. And we can then have all the deep dive docs on the recursion implementation there and then we just need to link to there from the lib doc page. I feel like that would make it easier to maintain down the line and centralize it a bit.

LucioFranco (changes requested):
Sorry for the super long delay on the review here and thank you for the ping (I generally don't mind them if you're waiting for a review). I left a few comments that I would like to see changed but they are pretty minor and I think once those are good we are good to merge. Thank you for pushing through on this!

src/lib.rs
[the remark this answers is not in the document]
seanlinsley: I don't have experience with doc wrangling in Rust. If you feel strongly about this, I'll just revert this change and leave it to a possible future PR.

prost-derive/src/lib.rs
seanlinsley: I'm having trouble getting a bespoke attribute parser working. Would it be okay to pull in darling as a dependency? https://crates.io/crates/darling

Another option is to skip proper attribute parsing and only handle the single attribute we have for now.

seanlinsley: Two years later and I finally got around to this. The code now uses `#[prost(recursion_limit = 5)]`

src/lib.rs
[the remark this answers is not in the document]
seanlinsley: 🤷 I'm not sure if a `constants` module adds much. If you still feel a constant is necessary, I'd lean towards simply adding back the constant here in `lib.rs`

README.md
```diff
@@ -467,7 +467,7 @@ configured with the required dependencies to compile the whole project.
 - `std`: Enable integration with standard library. Disable this feature for `no_std` support. This feature is enabled by default.
 - `derive`: Enable integration with `prost-derive`. Disable this feature to reduce compile times. This feature is enabled by default.
 - `prost-derive`: Deprecated. Alias for `derive` feature.
-- `no-recursion-limit`: Disable the recursion limit. The recursion limit is 100 and cannot be customized. 
+- `no-recursion-limit`: Disable the recursion limit. The recursion limit is 100, and can be changed with `prost_build::recursion_limit`.
```
caspermeijn, line 470: That should be `prost_build::Config::recursion_limit`, right?

prost-derive/src/field/mod.rs
```diff
@@ -225,7 +225,7 @@ impl fmt::Display for Label {
 }
 
 /// Get the items belonging to the 'prost' list attribute, e.g. `#[prost(foo, bar="baz")]`.
-fn prost_attrs(attrs: Vec<Attribute>) -> Result<Vec<Meta>, Error> {
+pub(super) fn prost_attrs(attrs: Vec<Attribute>) -> Result<Vec<Meta>, Error> {
```
caspermeijn, line 228: Just to verify: this doesn't make the function visible outside this crate, right?

prost-derive/src/lib.rs
```diff
@@ -203,6 +216,10 @@ fn try_message(input: TokenStream) -> Result<TokenStream, Error> {
             fn clear(&mut self) {
                 #(#clear;)*
             }
+
+            fn recursion_limit() -> u32 {
```
caspermeijn, line 220: Is this a breaking change for messages with a field named `recursion_limit`?

I think for enum fields, an accessor is created. However, this function is in a trait, so not directly accessible.

I would like to see a test case for that.

prost/src/message.rs
```diff
@@ -145,12 +145,22 @@ pub trait Message: Send + Sync {
...
         )
     }
 
     /// Clears the message, resetting all fields to their default.
     fn clear(&mut self);
+
+    /// The recursion limit for decoding protobuf messages.
+    ///
+    /// Defaults to 100. Can be customized in your build.rs or by using the no-recursion-limit crate feature.
+    fn recursion_limit() -> u32
```
caspermeijn, line 158: Is this a breaking change? I feel like it is not, because you add a default implementation. However rust rules for breaking change are not always logical. Could you verify this with `cargo-semver-checks`?

caspermeijn (changes requested):
Thank you for coming back to this after two years. I see why this is useful and this is a smart solution.

I am a bit concerned whether this could be a breaking change. I would like to see some prove that it is not.

prost/src/encoding.rs
```diff
@@ -37,29 +37,27 @@ pub struct DecodeContext {
...
     /// the recursion limit.
     ///
-    /// The recursion limit is defined by `RECURSION_LIMIT` and cannot be
-    /// customized. The recursion limit can be ignored by building the Prost
-    /// crate with the `no-recursion-limit` feature.
+    /// It defaults to 100 and can be changed using `prost_build::recursion_limit`,
+    /// or it can be disabled entirely using the `no-recursion-limit` feature.
     #[cfg(not(feature = "no-recursion-limit"))]
-    recurse_count: u32,
+    #[doc(hidden)]
```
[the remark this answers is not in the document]
seanlinsley: @LucioFranco [previously suggested](https://github.com/tokio-rs/prost/pull/785#discussion_r1096512577) that it should be `doc(hidden)` because the field was changed to public. I assume a previous version of this PR needed it to be public but it seems like it's no longer needed. I'll revert this field back to private.

prost-derive/src/lib.rs
seanlinsley: I'm not sure I follow: how could this possibly affect a field named `recursion_limit`? This is implemented on the `Message` trait, not on the struct itself.

caspermeijn: Sorry for another long delay. Well, if a .proto file has a enum field named `recursion_limit`, it will generate a function named `recursion_limit()` on the struct. I would like to see a test that proved that the struct function can still be called.

prost/src/encoding.rs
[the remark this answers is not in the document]
caspermeijn: Yes, private is better. My experience is that `pub` and `doc(hidden)` makes a sort of secret public API. That will be used by someone.

prost/src/message.rs
caspermeijn: Thank you for checking that. The last time I looked `cargo-semver-checks` recommended to not do it in CI.

src/lib.rs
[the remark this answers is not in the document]
caspermeijn: I like it to be a constant as well. It makes it explicit in the code. But I do like it to be in the documentation as well.

prost-derive/src/lib.rs
```diff
@@ -11,23 +11,36 @@ use proc_macro2::{Span, TokenStream};
...
     let ident = input.ident;
 
     syn::custom_keyword!(skip_debug);
     let skip_debug = input
         .attrs
-        .into_iter()
+        .iter()
         .any(|a| a.path().is_ident("prost") && a.parse_args::<skip_debug>().is_ok());
 
+    let mut recursion_limit: u32 = 100;
```
caspermeijn, line 30: I think the default should not be named here. If the attribute is not set, then no code should be generated. Which automatically chooses the default implementation of the trait. This makes sure that `prost-derive` doesn't have to know about the default value,

caspermeijn (changes requested):
Thank you for your contribution and patience. The code needs a rebase with master.

prost-derive/src/lib.rs
seanlinsley: This is tested in `tests/src/recursion_limit_field.rs`

seanlinsley:
@caspermeijn all review comments should be addressed

PR #786, opened 2022-12-18
Allow file descriptor be generated without --include_source_info

dfreese:
The file descriptor sets generated by rules_proto in bazel don't have this by default, so this allows some flexibility for users to reuse results that are already available to them.  It's fairly trivial adjustment so it seemed reasonable to me to allow the flexibility.

### DIFF
--- a/prost-build/src/ast.rs
+++ b/prost-build/src/ast.rs
@@ -134,7 +134,7 @@ pub struct Service {
     /// The package name as it appears in the .proto file.
     pub package: String,
     /// The service comments.
-    pub comments: Comments,
+    pub comments: Option<Comments>,
     /// The service methods.
     pub methods: Vec<Method>,
     /// The service options.
@@ -149,7 +149,7 @@ pub struct Method {
     /// The name of the method as it appears in the .proto file.
     pub proto_name: String,
     /// The method comments.
-    pub comments: Comments,
+    pub comments: Option<Comments>,
     /// The input Rust type.
     pub input_type: String,
     /// The output Rust type.
--- a/prost-build/src/ast.rs
+++ b/prost-build/src/ast.rs
@@ -5,7 +5,7 @@ use pulldown_cmark::{CodeBlockKind, Event, Options, Parser, Tag};
 use regex::Regex;
 
 /// Comments on a Protobuf item.
-#[derive(Debug, Clone)]
+#[derive(Debug, Default, Clone)]
 pub struct Comments {
     /// Leading detached blocks of comments.
     pub leading_detached: Vec<Vec<String>>,
@@ -134,7 +134,7 @@ pub struct Service {
     /// The package name as it appears in the .proto file.
     pub package: String,
     /// The service comments.
-    pub comments: Option<Comments>,
+    pub comments: Comments,
     /// The service methods.
     pub methods: Vec<Method>,
     /// The service options.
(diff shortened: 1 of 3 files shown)


prost-build/src/ast.rs
```diff
@@ -134,7 +134,7 @@ pub struct Service {
     /// The package name as it appears in the .proto file.
     pub package: String,
     /// The service comments.
-    pub comments: Comments,
+    pub comments: Option<Comments>,
```
LucioFranco, line 137: instead of making this breaking change can we instead make it so comments has a `Comments::default()`?

dfreese: Looking at prost-build/src/lib.rs:154, the ast module isn't public, so I assumed this was an implementation detail of `prost-build`.
```rust
mod ast
```
Did I misunderstand that somehow?

dfreese: oh, nevermind.  Missed line 147:
```rust
pub use crate::ast::{Comments, Method, Service};
```

dfreese: Added a derived `Default` to `Comment`.  I went that direction instead of directly defining one since all of the fields of `Comment` are public, so unlikely to change, and the derived default is what I'd expect here.

LucioFranco:
Awesome, I am gonna need some time to review this but I am keen to merge this. One cool thing that this could allow us is to use the `protobuf-parse` crate's pure rust protobuf parser since the blocker was SourceCodeInfo. I have a small branch with it trying to see if we could just replace all usage of protoc with it but it ran into some issues not related to this PR but thought I would mention that since I think its a cool step forward for prost anyways!

LucioFranco (approved):

andrewhickman:
> Awesome, I am gonna need some time to review this but I am keen to merge this. One cool thing that this could allow us is to use the `protobuf-parse` crate's pure rust protobuf parser since the blocker was SourceCodeInfo. I have a small branch with it trying to see if we could just replace all usage of protoc with it but it ran into some issues not related to this PR but thought I would mention that since I think its a cool step forward for prost anyways!

I've been working on another pure-rust replacement for protoc, [protox](https://docs.rs/protox/0.2.0/protox/), which does have full support for SourceCodeInfo. Its not very mature yet, but it should be sufficient for use with prost-build. My main concern with using it directly in prost-build would be that it uses prost-types internally, which might make publishing new versions difficult

LucioFranco:
> I've been working on another pure-rust replacement for protoc, [protox](https://docs.rs/protox/0.2.0/protox/), which does have full support for SourceCodeInfo. Its not very mature yet, but it should be sufficient for use with prost-build. 

Oh wow, this is really really awesome! If you're interested, would love to add this to prost or something like that. I think this could be a huge valuable change to the protobuf ecosystem in rust. 

> My main concern with using it directly in prost-build would be that it uses prost-types internally, which might make publishing new versions difficult

I don't think this matters since prost-build depends on prost-types?

andrewhickman:
> I don't think this matters since prost-build depends on prost-types?

I was thinking the process would have to be publish prost-types -> publish protox -> publish prost-build. Its certainly doable, and I'd be happy to share ownership of protox or merge it into the prost repo to make releases easier

LucioFranco:
@andrewhickman yeah, I am open to what ever you are interested in. I don't have much time to take on the additional maintenance but if you're interested in contributing it, I can add you as a maintainer. I would love some help on prost anyways if you're interested.

PR #796, opened 2023-01-12
Added try_normalize to Timestamp

oliverbrowneprima:
Hi!

I'm using `Timestamp` in a context where the overflow handling in `normalize` isn't really ideal (I'm transforming a `Timestamp` to a `chrono::DateTime<Utc>`, which needs unsigned nanos, and I am better off dropping the timestamp than recording faked values), and I implemented this in my project and though it might be handy to upstream. Let me know what you think!

Thanks,
Olly

### DIFF
--- a/prost-types/src/lib.rs
+++ b/prost-types/src/lib.rs
@@ -233,6 +233,25 @@ impl Timestamp {
         //               "invalid timestamp: {:?}", self);
     }
 
+    /// Normalizes the timestamp to a canonical format, returning the original value if it cannot be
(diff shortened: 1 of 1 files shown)


oliverbrowneprima:
Was going to write a test for the condition, but it's non-trivial to construct an error we should actually retry, and I'm reasonably happy without it (since most of the logic is library code anyway)

LucioFranco (approved):
LGTM thanks! Sorry for the delay

Issue #795, opened 2023-01-12
Feature request: Optionally leave out Debug impl


audunska:
We are using tonic, prost and tracing in a service handling customer data. All of our endpoint handler functions look like for example
```rust
#[instrument(skip(self))]
async fn handle_search_files(
    &self,
    request: SearchFilesRequest,
) -> Result<Response<SearchFilesStream>, Status> {
    ...
}
```
where the `SearchFilesRequest` is a prost message. This pushes the request body to the tracing provider using its `Debug` implementation. However, we would like to limit the amount of customer data pushed to the tracing provider, while still seeing the structure of the request as well as internal ids.

Possible ways of doing this is to manually go through all the endpoints and only push allowed fields to tracing, or to wrap all the request objects in a custom wrapper type with a restricted Debug impl. However, this seems quite brittle and boilerplate-heavy.

One solution that would be really neat is to derive our own Debug impl for the messages, by using a prost_build config with `.type_attribute(".", "#[derive(CustomDebug)]")`. This custom debug deriver could then consult an internal allowlist for which fields to include in the debug output, and which fields to leave out. However, this is not possible if `prost_build` already provides its own `Debug`.

So, my question is: Would it be possible to optionally leave out the Debug impl for messages and oneofs?

LucioFranco:
Yeah, I think we can just do this https://docs.rs/prost-build/latest/prost_build/struct.Config.html#method.disable_comments but for debug impl? Should be pretty easy to add and isn't a breaking change.

audunska:
> Yeah, I think we can just do this https://docs.rs/prost-build/latest/prost_build/struct.Config.html#method.disable_comments but for debug impl? Should be pretty easy to add and isn't a breaking change.

You would think that, but the Debug impl is generated in `prost-derive` as part of the `#[derive(Message)]` and friends, and that doesn't have access to the config. I guess we could add a struct attribute to control whether `Debug` is generated.

LucioFranco:
Ah you're right, maybe we can have a attribute option that sets no debug and then that can be enabled via prost-build.

audunska:
> Ah you're right, maybe we can have a attribute option that sets no debug and then that can be enabled via prost-build.

Yes, I am investigating this option.

AlJohri:
Related: #334 and #367

caspermeijn:
[...]

PR #797, opened 2023-01-13
Make Debug impl optional for types

audunska:
See #795 for motivation. If this looks like an ok approach, I'll go ahead and add more documentation.

Where is a good place to put some tests for this behavior?


LucioFranco (commented):
This seems like a fine approach, I would like to see a lot of tests added around the different combinations of debug impls and to also make sure the root level debug isn't a problem.

AlJohri:
Friendly bump on this @audunska-- would you be able to add some tests? This would be a really useful feature.

audunska:
I didn't end up prioritizing it, but I can do it now :)

audunska:
Ok, I've added some tests testing both the Message trait directly, and testing code generation. Anything else?

LucioFranco (approved):
Thanks for adding the tests!

PR #798, opened 2023-01-13
Add Kani to CI.

YoshikiTakashima:
This PR adds Kani to the CI as previously discussed with @LucioFranco.

- This PR adds Kani verification to only the `tests::check_timestamp_roundtrip_via_system_time` proptest harness. We focused on this harness because it was involved in bug #682
- Remaining harnesses within `prost-types` are blocked by [Kani issue 274](https://github.com/model-checking/kani/issues/274). We are working to implement the required features to resolve 274 and we will submit a second PR to enable these harnesses once Kani implements the feature.
- Harnesses outside of `prost-types` will be difficult to support within GitHub Actions due to Kani's large memory consumption when running these harnesses.

PR #799, opened 2023-01-18
Added Kani documentation.

YoshikiTakashima:
As requested in #798, this PR adds documentation on how to locally run Kani and PropProof. 

Since the documents are a bit long, I am adding `KANI.md` as a separate file in the same style as `FUZZING.md` rather than at the bottom of `README.md`.


LucioFranco:
Thanks!

PR #801, opened 2023-01-19
chore(types): Remove including generated code

tottoto:
Removes including generated codes from prost-types. This wil improve our development experience allowing rust-analyzer to jump to the actual generated codes instead of the include sentence.


neoeinstein:
FYI, this change was a breaking change due to the visibility of the `compiler` module going from public to private. See: #823

PR #803, opened 2023-02-02
Fix issue with negative nanos in Duration::try_from(), and add tests

danielsn:
If `seconds` are `0`, but `nanos` are negative, `Duration::try_from` incorrectly treats the duration as positive.  The negative `nanos` are unconditionally cast to `u32`, leading to an invalid positive `Duration`.  

1. Fixes the check.
2. Adds a regression test for the reproducer.
3. Adds a proptest to detect any cases where this might occur.

This issue was found using [Kani Rust Verifier](https://github.com/model-checking/kani)

### DIFF
--- a/prost-types/src/lib.rs
+++ b/prost-types/src/lib.rs
@@ -104,7 +104,7 @@ impl TryFrom<Duration> for time::Duration {
     /// Converts a `Duration` to a `std::time::Duration`, failing if the duration is negative.
     fn try_from(mut duration: Duration) -> Result<time::Duration, DurationError> {
         duration.normalize();
-        if duration.seconds >= 0 {
+        if duration.seconds >= 0 && duration.nanos >= 0 {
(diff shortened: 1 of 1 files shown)


prost-types/src/lib.rs
```diff
@@ -454,6 +454,51 @@ mod tests {
...
+        let nanos: u32 = 1;
+        let std_duration = std::time::Duration::new(seconds, nanos);
+
+        let neg_prost_duration = Duration {
+            seconds: 0,
+            nanos: -1,
+        };
+
+        assert!(matches!(
+           dbg!(time::Duration::try_from(neg_prost_duration)),
```
LucioFranco, line 499: Do we still need this dbg!?

danielsn: Removed

LucioFranco (approved):

PR #811, opened 2023-02-05
Remove usage of the flag experimental_allow_proto3_optional

silvergasp:
Optional fields in the proto3 syntax are now stabilised. This flag has been removed in newer version of protoc and now causes issues building tests.

### DIFF
--- a/tests/src/build.rs
+++ b/tests/src/build.rs
@@ -91,7 +91,6 @@ fn main() {
         .unwrap();
 
     prost_build::Config::new()
-        .protoc_arg("--experimental_allow_proto3_optional")
         .compile_protos(&[src.join("proto3_presence.proto")], includes)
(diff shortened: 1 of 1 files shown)


caspermeijn (commented):
Is this still relevant?

Can you provide `protoc` version numbers which are affected?

caspermeijn:
`--experimental_allow_proto3_optional` is required for supporting  `protoc` version 3.12. That version is still in use in Ubuntu 22.04. 

Modern versions of  `protoc` [ignore](https://github.com/protocolbuffers/protobuf/blob/c12c73212d40d212957db4893c4c2bf2a0e012b9/src/google/protobuf/compiler/command_line_interface.cc#L2235) the argument for backwards compatibility.

So I think we should leave this in place

PR #828, opened 2023-03-13
Add existing roundtrip test to Kani CI and avoid recursive submoduling

adpaco-aws:
* Adds the `check_duration_roundtrip_nanos` test to the harnesses run with Kani in CI
 * Avoid recursive submodule initialization for `propproof` because it isn't needed (see https://github.com/model-checking/kani-github-action/pull/24)
 * Upgrade Kani's GitHub Action to use version `0.23`

Note: This PR doesn't change the test, it was added in #803

### DIFF
--- a/.github/workflows/continuous-integration-workflow.yaml
+++ b/.github/workflows/continuous-integration-workflow.yaml
@@ -104,7 +104,7 @@ jobs:
       - name: checkout
         uses: actions/checkout@v3
       - name: Verify with Kani
-        uses: model-checking/kani-github-action@v0.23
+        uses: model-checking/kani-github-action@0.23
         with:
           enable-propproof: true
           args: |
--- a/.github/workflows/continuous-integration-workflow.yaml
+++ b/.github/workflows/continuous-integration-workflow.yaml
@@ -104,7 +104,7 @@ jobs:
       - name: checkout
         uses: actions/checkout@v3
       - name: Verify with Kani
-        uses: model-checking/kani-github-action@v0.19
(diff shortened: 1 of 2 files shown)


.github/workflows/continuous-integration-workflow.yaml
```diff
@@ -103,15 +103,14 @@ jobs:
...
       - name: Verify with Kani
-        uses: model-checking/kani-github-action@v0.19
+        uses: model-checking/kani-github-action@0.23
         with:
           enable-propproof: true
           args: |
             --tests -p prost-types --default-unwind 3 \
-            --harness "tests::check_timestamp_roundtrip_via_system_time"
+            --harness "tests::check_timestamp_roundtrip_via_system_time" \
+            --harness "tests::check_duration_roundtrip_nanos"
```
LucioFranco, line 113: It would be nice if we could include kani tests via some sort of wildcard instead of listing each test.

LucioFranco (approved):
LGTM thanks!

.github/workflows/continuous-integration-workflow.yaml
adpaco-aws: Actually, this can already be done since we upgraded to `0.23`. The `--harness <harness>` argument will run any harnesses for which `<harness>` is a sub-string will be run (i.e., it works like `cargo test --test <test>`). 

Unfortunately, `--harness tests::check` matches other tests which take much longer to verify. Because of that, I'd suggest leaving full names for now.

PR #839, opened 2023-04-10
Fix build in directory not named `prost`

3541:
This library will fail to build with the following error when checked out into a directory not named exactly `prost`:

```
error: failed to load manifest for workspace member `/home/alex/src/prost-rs/tests/single-include`

Caused by:
  failed to load manifest for dependency `prost`

Caused by:
  failed to read `/home/alex/src/prost/Cargo.toml`

Caused by:
  No such file or directory (os error 2)
```

This is because the `single-include` test depends on `prost` with the path `../../../prost`. This patch fixes the issue by correcting the path to `../..`.

### DIFF
--- a/tests/single-include/Cargo.toml
+++ b/tests/single-include/Cargo.toml
@@ -7,7 +7,7 @@ publish = false
 license = "MIT"
 
 [dependencies]
-prost = { path = "../../../prost" }
+prost = { path = "../.." }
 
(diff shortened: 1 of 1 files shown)


LucioFranco (approved):
haha didn't realize this was there, that is actually kinda funny. Nice catch!

PR #845, opened 2023-04-16
Fix invalid `fq_message_name` when appending nested message in empty package

### LINKED ISSUE #843: Invalid generated code when using a nested message with a map field
Using a protobuf file with these contents:

```
syntax = "proto3";
message M {
  message SubMessage {
    map<string, string> item = 1;
  }
  SubMessage reply = 2;
}
```

I'm getting invalid code:

```
pub struct M {
    #[prost(message, optional, tag = "2")]
    pub reply: ::core::option::Option<m::SubMessage>,
}
/// Nested message and enum types in `M`.
pub mod m {
    pub struct SubMessage {
        #[prost(message, repeated, tag = "1")]
        pub item: ::prost::alloc::vec::Vec<sub_message::ItemEntry>,
    }
}
```

the error: 
```
error[E0433]: failed to resolve: use of undeclared crate or module `sub_message`
  --> D:\code\sni_things\target\debug\build\sni_things-e5c2e47c00096f71\out/_.rs:13:44
   |
13 |         pub item: ::prost::alloc::vec::Vec<sub_message::ItemEntry>,
   |                                            ^^^^^^^^^^^ use of undeclared crate or module `sub_message`
```

This works if I change the `map` in the proto definition into a scalar, or if I move `SubMessage` out of `M` (so it's no longer nested). (of course in real life I don't control the protobuf file, so this solution doesn't work very well).

ldm0:
On appending nested message, if package is empty, wrong `fq_message_name` is generated(e.g. `..M.SubMessage` in this case), which leads to codegen failure. Making `CodeGenerator::package` vec and removing the string hacks solve this issue.

fixes #843

### DIFF
--- a/prost-build/src/code_generator.rs
+++ b/prost-build/src/code_generator.rs
@@ -2,6 +2,7 @@ use std::ascii;
 use std::borrow::Cow;
 use std::collections::{HashMap, HashSet};
 use std::iter;
+use std::ops::Deref;
 
 use itertools::{Either, Itertools};
 use log::debug;
@@ -28,7 +29,7 @@ enum Syntax {
 
 pub struct CodeGenerator<'a> {
     config: &'a mut Config,
-    package: String,
+    package: Vec<String>,
     source_info: Option<SourceCodeInfo>,
     syntax: Syntax,
     message_graph: &'a MessageGraph,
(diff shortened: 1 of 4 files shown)


caspermeijn:
This has been fixed by https://github.com/tokio-rs/prost/pull/981

Issue #836, opened 2023-04-05
clippy warn


liyongjing:
run cargo clippy -- -W clippy::pedantic 

```  
    |
    | /// Creates an enum from field names used in the ProtoBuf definition.
    |                                                                                   ^^^^^^^^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#doc_markdown
help: try
    |
    |   /// Creates an enum from field names used in the `ProtoBuf` definition.
    |                                                                                     ~~~~~~~~~~

warning: item in documentation is missing backticks
```

```
    |
    |     pub fn as_str_name(&self) -> &'static str {
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: add the attribute: `#[must_use] pub fn as_str_name(&self) -> &'static str`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#must_use_candidate
```

LucioFranco:
We don't follow clippy and only suppress compiler warnings. So in this case I would recommend annotating your code with the correct allow.

LucioFranco:
Sorry I didn't see that first error I am not sure why that is happening but I believe Prost just copies the docs directly from the proto file so if it doesn't have backticks there then that is why.

LucioFranco:
Oh right! yes, you're 100% right, we can add that. Feel free to open a PR! Thanks for calling this out!

PR #847, opened 2023-04-18
Fixing issues with clippy and rustdoc

ianmichell:
Pull request to fix the rustic issues that come up as errors when running clippy -Dclippy::pedantic

Referenced in #836


ianmichell:
That's ok , it's your project! I do think clippy is very useful, I ran clippy over the whole project and found loads that could make things neater. I've bypassed clippy in my project for Prost generated code now anyway.

It's a great project. I've just started using it in my distributed log project for raft 😎

PR #849, opened 2023-04-23
generate assoc. consts in enums representing protobuf enum aliases, resolves #792

### LINKED ISSUE #792: Unable to compile protos with aliased default values
Given this valid protobuf:

```proto
syntax = "proto2";

package test;

enum Foo {
  option allow_alias = true;

  BAR = 1;
  ZAA = 2;

  DEFAULT = 1;
}

message MyMessage {
  optional Foo f = 1 [ default = DEFAULT ];
}
```

prost generates this error
```
❯ cargo build
   Compiling prost-bug v0.1.0 (/Users/jason/projects/prost-bug)
error[E0599]: no variant or associated item named `Default` found for enum `Foo` in the current scope
 --> /Users/jason/projects/prost-bug/target/debug/build/prost-bug-6fb2c18e0f681f70/out/test.rs:2:28
  |
2 | #[derive(Clone, PartialEq, ::prost::Message)]
  |                            ^^^^^^^^^^^^^^^^ variant or associated item not found in `Foo`
...
9 | pub enum Foo {
  | ------------ variant or associated item `Default` not found for this enum
  |
  = note: this error originates in the derive macro `::prost::Message` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0599`.
error: could not compile `prost-bug` due to previous error
```

The issue seems to be that prost ignores `DEFAULT` when generating an enum for `Foo` since rust enums do not allow aliases like C++. However, it still references `Default` when generating generating `MyMessage`.

I'm not sure what the best fix would be here, but since this is explicitly allowed by the protobuf spec it feels like it should compile with prost. Since the `f` field just becomes an `i32` anyways, maybe  prost can generate an integer constant for the aliased values?

I put a simple reproduction in a small project here: https://github.com/jteplitz/prost-alias-default-bug

ousado:
This PR adds associated constants to enum `impl`s representing enum aliases in protocol buffer enums as shown below. The aliases can be referenced by prost when used as default values, causing compilation to fail, see #792.

```rust
#[derive(PartialEq, Eq)]
pub enum X {
    A,
    B,
}

impl X {
    pub const A1: X = X::A;
}
```

### DIFF
--- a/prost-build/src/code_generator.rs
+++ b/prost-build/src/code_generator.rs
@@ -662,7 +662,12 @@ impl<'a> CodeGenerator<'a> {
 
         self.depth += 1;
         self.path.push(2);
+        let mut aliases = vec![];
         for variant in variant_mappings.iter() {
+            if variant.alias_of.is_some(){
+                aliases.push(variant);
+                continue;
+            }
             self.path.push(variant.path_idx as i32);
 
             self.append_doc(&fq_proto_enum_name, Some(variant.proto_name));
@@ -688,7 +693,21 @@ impl<'a> CodeGenerator<'a> {
         self.buf.push_str(" {\n");
         self.depth += 1;
         self.path.push(2);
-
+        self.push_indent();
+        self.buf.push_str(
+            "/// Aliases.\n",
+        );
+        for variant in &aliases {
+            self.push_indent();
+            self.buf.push_str("#[allow(non_upper_case_globals)]");
+            self.push_indent();
+            self.buf.push_str(&format!("pub const {}: {} = {}::{};\n",
+                                       variant.generated_variant_name,
+                                       enum_name,
+                                       enum_name,
+                                       variant.alias_of.as_ref().unwrap())
+                              );
+        }
         self.push_indent();
         self.buf.push_str(
             "/// String value of the enum field names used in the ProtoBuf definition.\n",
@@ -713,6 +732,7 @@ impl<'a> CodeGenerator<'a> {
         self.depth += 1;
 
         for variant in variant_mappings.iter() {
+            if variant.alias_of.is_some() {continue;}
             self.push_indent();
             self.buf.push_str(&enum_name);
             self.buf.push_str("::");
@@ -1136,6 +1156,7 @@ struct EnumVariantMapping<'a> {
     proto_name: &'a str,
     proto_number: i32,
     generated_variant_name: String,
+    alias_of: Option<String>,
 }
 
 fn build_enum_value_mappings<'a>(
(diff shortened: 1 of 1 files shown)


prost-build/src/code_generator.rs
```diff
@@ -1145,15 +1169,20 @@ fn build_enum_value_mappings<'a>(
...
     for (idx, value) in enum_values.iter().enumerate() {
-        // Skip duplicate enum values. Protobuf allows this when the
+        // Remember name for duplicate enum values. Protobuf allows this when the
         // 'allow_alias' option is set.
+        let mut alias_of = None;
         if !numbers.insert(value.number()) {
-            continue;
+            for m in &mappings {
+                if m.proto_number == value.number() {
+                    alias_of = Some(m.generated_variant_name.clone());
```
caspermeijn, line 1181: So the first value is the original and any values with the same integer value are aliases. Do I understand correct?

prost-build/src/code_generator.rs
```diff
@@ -1145,15 +1169,20 @@ fn build_enum_value_mappings<'a>(
...
 
     for (idx, value) in enum_values.iter().enumerate() {
-        // Skip duplicate enum values. Protobuf allows this when the
+        // Remember name for duplicate enum values. Protobuf allows this when the
         // 'allow_alias' option is set.
+        let mut alias_of = None;
         if !numbers.insert(value.number()) {
-            continue;
+            for m in &mappings {
+                if m.proto_number == value.number() {
```
caspermeijn, line 1180: `genereated_names` can be used to do this lookup in one step, right?

prost-build/src/code_generator.rs
```diff
@@ -1162,15 +1191,22 @@ fn build_enum_value_mappings<'a>(
 
         if let Some(old_v) = generated_names.insert(generated_variant_name.to_owned(), value.name())
         {
-            panic!("Generated enum variant names overlap: `{}` variant name to be used both by `{}` and `{}` ProtoBuf enum values",
-                generated_variant_name, old_v, value.name());
+            // if alias ends up being a duplicate, we don't need it, and can skip it.
+            // TODO: check if enum values are actually the same
```
caspermeijn, line 1195: I think this TODO needs to be solved

caspermeijn (commented):
This changes makes sense to me. 

- I think the PR title should be something like: `Add support for enum aliases`.
- You need to add some tests for the new use cases. 
- Does prost need to check for `allow_alias` option? If `protoc` throw an error when `allow_alias` is not set, then we don't have to.

prost-build/src/code_generator.rs
ousado: Yes, in this rust representation that's the case, I don't think protobuf itself makes any distinction there, it simply allows the same value to be repeated.

prost-build/src/code_generator.rs
ousado: I don't see how, that would require a lookup by value instead of name, no?

prost-build/src/code_generator.rs
```diff
@@ -688,7 +693,22 @@ impl<'a> CodeGenerator<'a> {
         self.buf.push_str(" {\n");
         self.depth += 1;
         self.path.push(2);
-
+        if aliases.len() > 0 {
+            self.push_indent();
+            self.buf.push_str("/// Aliases.\n");
```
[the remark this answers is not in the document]
ousado: right

prost-build/src/code_generator.rs
caspermeijn: You are right. The HashMap is the other way around. Please disregard my comment.

prost-build/src/code_generator.rs
caspermeijn: I feel like duplicates should always be disallowed. Even if they are an alias, then overlap is not disirable.

Issue #850, opened 2023-04-23
Linking to items in doc comment is not possible because brackets are always escaped


pgimalac:
Since https://github.com/tokio-rs/prost/pull/510, writing a doc comment linking to an item is not possible because brackets are escaped, meaning that the following proto definition
```proto
// A message, represented by a [`Data`].
message Msg {
    Data content = 1;
}
```
produces the following rust code
```rust
/// A message, represented with by a \[`Data`\].
#[allow(clippy::derive_partial_eq_without_eq)]
#[derive(Clone, PartialEq, ::prost::Message)]
pub struct Msg {
    #[prost(message, optional, tag = "1")]
    pub content: ::core::option::Option<Data>,
}
```

A trick which works is to add a space between the brackets because [the regex used](https://github.com/tokio-rs/prost/blob/2037251cc50e25247972d5c1131905de3064016c/prost-build/src/ast.rs#L115) expects non-space characters between the brackets, but then the generated text has an extra space too.

I understand that not all brackets are for doc links (like with https://github.com/tokio-rs/prost/issues/495) but it would be nice to at least have a way to write such doc links properly.

I think that one of the following solutions would be reasonable:
- not escaping brackets which contain backticks:
```proto
// A message, represented by a [`Data`].
```
- not escaping brackets which are followed by parenthesis
```proto
// A message, represented by a [Data](Data).
```

PR #851, opened 2023-04-23
prost-build: do not escape brackets followed by parenthesis in comments

pgimalac:
A proposal to fix https://github.com/tokio-rs/prost/issues/850.

Allows linking to items in doc comments by not escaping brackets followed by parenthesis.

- Before (brackets always escaped):
`// Some comment with a [Link]` -> `/// Some comment with a \[Link\]`
`// Some comment with a [Link](Link)` -> `/// Some comment with a \[Link\](Link)`
- After (brackets not escaped if followed by parenthesis):
`// Some comment with a [Link]` -> `/// Some comment with a \[Link\]`
`// Some comment with a [Link](Link)` -> `/// Some comment with a [Link](Link)`

EDIT: I also implemented the case with brackets following brackets `[Link][Link]` since it appears in some of the tests.
EDIT: I also implemented not escaping brackets which are already escaped, it happens when using the `cleanup-markdown` feature, or if the user manually escapes brackets.

### DIFF
--- a/prost-build/Cargo.toml
+++ b/prost-build/Cargo.toml
@@ -31,7 +31,7 @@ prost = { version = "0.11.8", path = "..", default-features = false }
 prost-types = { version = "0.11.8", path = "../prost-types", default-features = false }
 tempfile = "3"
 lazy_static = "1.4.0"
-regex = { version = "1.5.5", default-features = false, features = ["std", "unicode-bool"] }
+regex = { version = "1.8.1", default-features = false, features = ["std", "unicode-bool"] }
 which = "4"
 
 prettyplease = { version = "0.2", optional = true }
--- a/prost-build/src/ast.rs
+++ b/prost-build/src/ast.rs
@@ -108,16 +108,16 @@ impl Comments {
 
     /// Sanitizes the line for rustdoc by performing the following operations:
     ///     - escape urls as <http://foo.com>
-    ///     - escape `[` & `]` if not followed by a parenthesis or bracket
+    ///     - escape `[` & `]` if not already escaped and not followed by a parenthesis or bracket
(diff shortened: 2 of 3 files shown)


pgimalac:
I'm seeing some weird behaviors on the `bootstrap` test, it looks like the input is escaped twice sometimes (there are times where the brackets in the input of `sanitize_line` is already escaped)... I'll look into it when I have more times.

pgimalac:
Some lines are escaped both by `ast::get_lines` (which reads some file using `pulldown_cmark_to_cmark`) and then by `sanitize_input`, resulting in the double escaping I mentioned before.

Note that it's already the case, for example [here](https://github.com/tokio-rs/prost/blob/2037251cc50e25247972d5c1131905de3064016c/prost-types/src/protobuf.rs#L1259) where the regex only matched the outer brackets.

It seems like the `cleanup-markdown` feature (which is enabled for the tests) already escapes brackets.

I'll update the regex to avoid double escaping, since anyway if the user escapes the brackets then we don't want to do it again.

pgimalac:
The diff of `prost-types/src/protobuf.rs` is really nice:
- removes double escaping (eg. line 1259)
- escapes brackets which weren't escaped before (eg. line 1987)
- do not escape brackets which are a link (eg. line 2288)

baxterjo:
Hoping to bump this thread. Would be nice to have intra-doc linking in rustdocs.

LucioFranco:
@pgimalac if you want to get the conflicts resolved we can get this merged.

pgimalac:
@LucioFranco I fixed the conflicts but there is a build error (unrelated to this PR) in the CI

Issue #812, opened 2023-02-06
[Question] why not TryFrom<i32> instead of `from_i32` in derive(prost::Enumeration)?


Tudyx:
I've noticed that in derive `prost::Enumeration` we don't implement the std trait [TryFrom](https://doc.rust-lang.org/stable/std/convert/trait.TryFrom.html) but instead we use a custom function `from_i32`. Is there a particular reason for this (for instance MSRV) ?

I'm currently working on a macro that generates the boiler plate conversion functions from `Prost` generated struct that might contain unwanted `Option` because proto3 which doesn't support required fields.

For that it would be nice for me if I could have a generic implementation for types that have this function `from_i32`, that's why it would be nice for me to have it on a trait. 
Its currently working for all other protobuf type (nested or not) and I might have found a workaround for this, I was just wondering why the things are done this way.

Leulz:
> That code was written before TryFrom was stable :) I think we can add it now though.

Hey, @LucioFranco, I'm interested in implementing TryFrom<i32> for the enums that derive prost::Enumeration. Can you assign this issue to me, please?

Also, should the change make `from_i32` deprecated and change whatever uses from_i32 (like [this](https://github.com/tokio-rs/prost/blob/4e14adf4be890d031d05da7521e3c1a1984007fa/prost-derive/src/field/map.rs#L282)) to use try_from? Or should it just add the option to use TryFrom but keep from_i32?

messense:
I'm getting a `conflicting implementations of trait TryFrom<i32> for type Type` error when trying to upgrade to tonic 0.10.0 for [`etcd-client`](https://github.com/etcdv3/etcd-client), I think it might related to #853?

```diff
diff --git a/Cargo.toml b/Cargo.toml
index 1b4737d..b4dbcb8 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,8 +19,8 @@ tls-roots = ["tls", "tonic/tls-roots"]
 pub-response-field = ["visible"]

 [dependencies]
-tonic = "0.9.2"
-prost = "0.11.9"
+tonic = "0.10.0"
+prost = "0.12.0"
 tokio = "1.28.0"
 tokio-stream = "0.1.14"
 tower-service = "0.3.2"
@@ -35,7 +35,7 @@ hyper-openssl = { version = "0.9", optional = true }
 tokio = { version = "1.28.0", features = ["full"] }

 [build-dependencies]
[...]

PR #853, opened 2023-04-25
feat: add TryFrom<i32> implementation to Enumeration

Leulz:
Related to #812.

This allows us to more easily have generic functions that deal with conversions between i32 and prost Enums.

I took the liberty of deprecating `from_i32` and updating the README.md to show how to use the TryFrom<i32> instead of `from_i32`. I also changed code that used `from_i32` so that it uses `try_from`. If this is not how things are usually done in this repo, please tell me so I can undo those changes and make it so the only change is the addition of the TryFrom<i32> implementation and the new test.

Also, one point that I am not sure about is the Error type of the TryFrom<i32> implementation. I used DecodeError, but if that is not suitable and I should create a new Error struct/enum, please tell me so I can fix it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,21 +163,22 @@ The `#[derive(::prost::Enumeration)]` annotation added to the generated
 ```rust,ignore
 impl PhoneType {
     pub fn is_valid(value: i32) -> bool { ... }
+    #[deprecated]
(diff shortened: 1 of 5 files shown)


LucioFranco (approved):
LGTM! Sorry for the delay, thanks for the contribution.

Issue #299, opened 2020-04-05
Serializing a Message to Any


KaiserKarel:
I'm unable to find any docs on how to create an Any from a Message. Do I need to serialize the message myself and set the typeURL?

fridge-dev:
+1, I need to deserialize an `Any` to a known type. In other languages, I see there are pack/unpack methods. I couldn't find such methods on the `Any` type (https://docs.rs/prost-types/0.6.1/src/prost_types/protobuf.rs.html#964-998 )

fdeantoni:
Similar to #277.

cetanu:
Subscribed... I know how to go from a type to Any, but not the other way around.

petergarnaes:
The tools for de-serializing an [Any](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#any) type is already there, you can match on the type URL, and decode the value field using [Message::decode](https://docs.rs/prost/0.6.1/prost/trait.Message.html#method.decode).

With a little bit of macro magic you don't have to write a lot of code, I have made an [example repo](https://github.com/petergarnaes/prost_error_details). IMO this is simpler and can easily be adapted to your own needs.

vbkaisetsu:
If there is a mechanism for Message to return type_url, it can support generics. 

```rust
use prost::Message;
use prost_types::Any;

fn to_any<T>(message: &T) -> Any where T: Message {
    Any {
        type_url: T::type_url().to_string(),
        value: message.encode_to_vec(),
    }
}

fn from_any<T>(message: &Any) -> Result<T, DecodeError> where T: Message + Default {
    if &message.type_url == T::type_url() {
        T::decode(message.value.as_slice())
    } else {
        Err(DecodeError::new("Invalid type_url"))
    }
}
```

colin-grapl:
Adding a `type_url` to Message would definitely help. After that it would be straightforward to add pack/unpack methods for messages.

nyanchor:
> Create a `datafusion-proto` crate for datafusion protobuf serialization apache/arrow-datafusion#1887

@vbkaisetsu is there a type_url for `prost::Message`? I don't think there is one if not this will be straightforward

vbkaisetsu:
@nyanchor I previously created https://github.com/tokio-rs/prost/pull/535 before to support seamless seriarization, but it needs further discussion.

LucioFranco:
[...]

PR #858, opened 2023-05-17
feat: `TypeUrl` trait + `Any` encoding support

tarcieri:
As discussed in #299, adds a `TypeUrl` trait which associates a type URL constant with a `Message` type.

This enables adding methods to `Any` for decoding/encoding messages:

- `Any::from_message`: encodes a given `Message`, returning `Any`.
- `Any::to_message`: decodes `Any::value` as the given `Message`, first validating the message type has the expected type URL.

### DIFF
--- a/src/type_url.rs
+++ b/src/type_url.rs
@@ -1,0 +1,19 @@
+//! Support for associating a type URL with a [`Message`].
+
+use crate::Message;
+
+/// Associate a type URL with the given [`Message]`.
+pub trait TypeUrl: Message {
+    /// Type URL for this [`Message`]. They take the form:
+    ///
+    /// ```text
+    /// /<package>.<TypeName>
+    /// ```
+    ///
+    /// For example:
+    ///
+    /// ```text
+    /// /foo.bar.baz.MyTypeName
+    /// ```
+    const TYPE_URL: &'static str;
+}
--- a/prost-types/src/lib.rs
+++ b/prost-types/src/lib.rs
@@ -24,6 +24,11 @@ use core::i64;
 use core::str::FromStr;
 use core::time;
 
+use prost::alloc::borrow::ToOwned;
+use prost::alloc::format;
+use prost::alloc::vec::Vec;
+use prost::{DecodeError, EncodeError, Message, TypeUrl};
+
 pub use protobuf::*;
 
 // The Protobuf `Duration` and `Timestamp` types can't delegate to the standard library equivalents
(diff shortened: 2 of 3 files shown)


src/type_url.rs
```diff
@@ -0,0 +1,19 @@
...
+    /// ```text
+    /// /<package>.<TypeName>
+    /// ```
+    ///
+    /// For example:
+    ///
+    /// ```text
+    /// /foo.bar.baz.MyTypeName
+    /// ```
+    const TYPE_URL: &'static str;
```
tarcieri, line 18: Just went with a `&'static str` for now, rather than trying to use a special type which validates the URL.

In my mind the goal is to get `prost-build` to spit these out eventually, in which case they should be well-formed. But maybe down the road it would be possible to add a URL type with a `const fn` parser.

src/type_url.rs
```diff
@@ -0,0 +1,19 @@
...
+    /// Type URL for this [`Message`]. They take the form:
+    ///
+    /// ```text
+    /// /<package>.<TypeName>
+    /// ```
+    ///
+    /// For example:
+    ///
+    /// ```text
+    /// /foo.bar.baz.MyTypeName
```
bouk, line 16: I don't think the slash is part of the type URL, the type name is `foo.bar.baz.MyTypeName` which then gets encoded into the Any message as `type.googleapis.com/foo.bar.baz.MyTypeName`

tarcieri: Hmm, I guess this particular encoding is an [artifact of the type URLs I'm working with](https://github.com/cosmos/cosmos-sdk/issues/10949) and not the standard encoding.

I'm still unclear though if all type URLs are expected to be prefixed with `https://type.googleapis.com/` or just Google's own well-known types, which is what the example is for.

testinginprod: There is difference between **type url** and **full type name** of a protobuf message.

For instance the **full type name** is not expected to have slashes.

The **Type URL** in a `google.Protobuf.Any`  is expected to [have at least one slash](https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/any.proto#L130).

So if this Type URL mentioned here is supposed to end up in an any then by specification, it should contain at least one slash.

testinginprod: If this **TYPE URL** we are talking about in this PR is supposed to be used to compose the TYPE URL of an any then we shouldn't be talking about *TYPE URL* but we should be talking about **FULLNAME* of a protobuf  message, they're different concepts.

bouk: Yes, the type url in Any should have a slash, but the type name that we attach to the trait should just be the full type name without a slash.

testinginprod: Then this trait should not be called `TypeURL`  it should be called `Fullname`. 

And the way we match a type URL is not the correct one:
- Not compliant: https://github.com/tokio-rs/prost/pull/858/files#diff-fe4eba36fffb306e25d27bf29ddea9cd1634ea4ad86705045c1cb13a3aa27346R59

- Compliant(golang): https://github.com/protocolbuffers/protobuf-go/blob/v1.31.0/types/known/anypb/any.pb.go#L337

tarcieri: The use case is populating the `type_url` field of an `Any` message.

If you're saying the type URL can always be computed from the fullname by prepending a slash, that's fine. That's what the implementation I'm using is expecting.

ewoolsey:
In addition to a TypeUrl trait, It would be pretty nice to get a `Response` trait that contains the associated type of the response. This would make sending a query and deserializing the response super effortless.

tarcieri:
Based on the feedback from this PR, I've opened #896 which provides similar functionality, but uses a `Name` trait with provided `full_name` and `type_url` methods.

Will go ahead and close this PR out.

PR #864, opened 2023-05-24
chore: Update to criterion 0.5

tottoto:
Updates to criterion 0.5.


tottoto:
I overlooked the msrv.

PR #867, opened 2023-06-03
Don't ignore errors in prost_attrs

glandium:

PR #868, opened 2023-06-04
Update prost-derive to syn 2

glandium:
(note this happens to include #867 too)


LucioFranco (approved):
Sorry for the delay, thanks!

Issue #429, opened 2021-01-19
Support for `std::sync::Arc` as indirection


segeljakt:
Hi, I wonder, would be possible to allow prost-types to be wrapped inside `std::sync::Arc`? The reason is that I want my messages to be sendable across both processes and threads. When sent across processes, my messages should be serialized/deserialized with `prost`. When sent across threads, I would like to use reference counting.

```rust
#[derive(prost::Message)]
struct MyMessage {
    #[prost(string, tag = "1")]
    label: Arc<String>,
    #[prost(message, tag = "2")]
    payload: Arc<Payload>,
}

#[derive(prost::Message)]
struct Payload {
    #[prost(i32, repeated, tag = "1")]
    stuff: Vec<i32>,
}
```

wchargin:
This would also be nice in conjunction with Tonic servers. If we had
something analogous to `bytes` or `btree_map`, like:

```rust
Config::new()
    .arc(&[".my_service.FooRequest.my_big_field"])
    .compile_protos(...)
```

…then we could get away with fewer wasted clones in RPC handlers.

I think that this might be nice, but am not strongly lobbying for it,
and wouldn’t be too upset if it were wontfixed.

danburkert:
Sounds like a good addition!  We should probably also support `Box` and `Rc` similarly.

bjchambers:
This could also be helpful (I think) for cases where a large object is received in the request and needs to outlive it. Allowing that object to be an `Arc` would allow it to be taken outside the request/response lifetime without copying.

name1e5s:
🤔Seems it's possible to implement the feature, I'll try this.

LambdaMan2K:
will this be on the road map?

jasongoodwin:
Yeah I can here just to comment/upvote this. 
I'm trying to use a watch with tonic and I need to clone the message that hits each receiver because arc isn't supported. I can probably modify some of the generated code.

AshleySchaeffer:
> I can probably modify some of the generated code.

@jasongoodwin do you happen to have an implementation for this? I'm in a similar position, only the data I'm sending can get quite large.

morrisonlevi:
> Rc cannot be supported since prost::Message requires Send + Sync

Honestly... why _does_ `Message` require `Send + Sync`? I peeked at the associated functions and there didn't seem any reason for it.

[...]

PR #878, opened 2023-06-30
Allow using `Box<[u8]>`/`Box<str>` for `bytes`/`string` fields

poliorcetics:
This reduces the size of objects by one byte each time they're used, which can help quite a lot in some cases.

This is fully optional and the defaults have not changed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -329,12 +329,12 @@ and the generated Rust code (`tutorial.rs`):
 ```rust,ignore
 #[derive(Clone, PartialEq, ::prost::Message)]
 pub struct Person {
-    #[prost(string, tag="1")]
+    #[prost(string = "string", tag="1")]
     pub name: ::prost::alloc::string::String,
     /// Unique ID number for this person.
     #[prost(int32, tag="2")]
     pub id: i32,
-    #[prost(string, tag="3")]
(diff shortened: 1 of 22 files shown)


poliorcetics:
I saw https://github.com/tokio-rs/prost/issues/429 but it's not what I wanted since it asks for two levels of indirection, `Arc<Vec<u8>>` for example

poliorcetics:
I tried to implement this in an external lib but it's inefficient if done naively (conversions to `String/Vec<u8>`) everywhere and since `prost` doesn't expose its helpers to build `Message`s, doing it efficiently means copying a lot of code that is intended for internal use only and is quite brittle when used in another crate.

michaelvanstraten:
Please merge this, would extremely helpful.

poliorcetics:
Rebased to fix conflicts, no actual changes made

caspermeijn:
It seems like https://github.com/tokio-rs/prost/pull/979 is interesting for this PR. It would simplify the encoding needed for `Box<str>` support, am I right?

poliorcetics:
I'm not working with prost anymore and this PR is way out of date, someone interested can retake it if necessary :)

PR #886, opened 2023-07-30
Add optional support for simdutf8

JakubOnderka:
[simdutf8](https://github.com/rusticstuff/simdutf8) is faster variant of `from_utf8` method than from `core`.


caspermeijn:
Sorry, we are not interested in a custom `from_utf8` implementation. If you really need this, I suggest improving `core` standard library.

PR #889, opened 2023-08-01
add Config::preserve_type_name to prevent renaming

ziprandom:
Another approach to prevent type renamings for selected paths. This is not finished, I just put it here to see if there is interest in merging a solution like that. If yes, I can wrap this up. 

* [ ] unit tests 

#207

### DIFF
--- a/prost-build/src/code_generator.rs
+++ b/prost-build/src/code_generator.rs
@@ -190,7 +190,8 @@ impl<'a> CodeGenerator<'a> {
         ));
         self.push_indent();
         self.buf.push_str("pub struct ");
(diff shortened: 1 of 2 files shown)


caspermeijn:
The intention of this project is to generate idiomatic code. The naming convention in Rust for structs is UpperCamelCase. If we ignore that conversion, the rust compiler will generate a warning and that is unwanted.

I believe preserving type names is not needed when the protobuf style guide is followed: https://protobuf.dev/programming-guides/style/

I feel like the `prost` code will become more complicated with this change, and it doesn't align with the description in the  `prost` README. Therefore I am closing the PR.

PR #890, opened 2023-08-01
prost-types: fix tests without default features

ema:
Only include tests using std if the feature is available. When running cargo test with --no-default-features there are a few failures due to the fact that std is missing.


LucioFranco:
I think you need to merge from master

caspermeijn:
@ema It this still relavant? I feel like the CI should find this error if it is still in master, right?

ema:
Closing, this issue affects 0.11.9 but is fixed in master.

PR #891, opened 2023-08-02
Fix `non_snake_case` warning on generated identity functions

aalexandrov:
The `Message` derive macro currently generates `Debug` implementations that contain code that violates the `non_snake_case` linter rule[^1]. This PR changes the macro expansion to add an `#[allow(...)]` attribute for this rule.

**Note**: If there is consensus that the solution is OK, I'll fix any failing tests to make this PR ready to merge.

[^1]: https://doc.rust-lang.org/stable/nightly-rustc/rustc_lint/nonstandard_style/static.NON_SNAKE_CASE.html

### DIFF
--- a/prost-derive/src/field/scalar.rs
+++ b/prost-derive/src/field/scalar.rs
@@ -230,6 +230,7 @@ impl Field {
             }
         } else {
             quote! {
+                #[allow(non_snake_case)]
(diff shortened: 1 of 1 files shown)


LucioFranco (approved):

temeddix:
When will the new version with this change come out? Currently this makes error output a mess.

PR #896, opened 2023-08-14
feat: `Name` trait + `Any` encoding support

tarcieri:
As discussed in #299 and #858, adds a `Name` trait which associates a type name and package constants with a `Message` type. It also provides `full_name` and `type_url` methods.

The `type_url` method is used by newly added methods on the `Any` type which can be used for decoding/encoding messages:

- `Any::from_msg`: encodes a given `Message`, returning `Any`.
- `Any::to_msg`: decodes `Any::value` as the given `Message`, first validating the message type has the expected type URL.

### DIFF
--- a/prost-types/src/lib.rs
+++ b/prost-types/src/lib.rs
@@ -37,6 +37,8 @@ pub use protobuf::*;
 const NANOS_PER_SECOND: i32 = 1_000_000_000;
 const NANOS_MAX: i32 = NANOS_PER_SECOND - 1;
 
+const PACKAGE: &str = "google.protobuf";
+
 impl Any {
     /// Serialize the given message type `M` as [`Any`].
     pub fn from_msg<M>(msg: &M) -> Result<Self, EncodeError>
@@ -75,6 +77,15 @@ impl Any {
         ));
         err.push("unexpected type URL", "type_url");
         Err(err)
+    }
+}
+
+impl Name for Any {
+    const PACKAGE: &'static str = PACKAGE;
+    const NAME: &'static str = "Any";
+
+    fn type_url() -> String {
+        type_url_for::<Self>()
     }
 }
 
@@ -127,6 +138,15 @@ impl Duration {
         // TODO: should this be checked?
         // debug_assert!(self.seconds >= -315_576_000_000 && self.seconds <= 315_576_000_000,
         //               "invalid duration: {:?}", self);
+    }
+}
+
+impl Name for Duration {
+    const PACKAGE: &'static str = PACKAGE;
+    const NAME: &'static str = "Duration";
+
+    fn type_url() -> String {
+        type_url_for::<Self>()
     }
 }
 
@@ -343,6 +363,15 @@ impl Timestamp {
     }
 }
 
+impl Name for Timestamp {
+    const PACKAGE: &'static str = PACKAGE;
+    const NAME: &'static str = "Timestamp";
+
+    fn type_url() -> String {
+        type_url_for::<Self>()
+    }
+}
+
 /// Implements the unstable/naive version of `Eq`: a basic equality check on the internal fields of the `Timestamp`.
 /// This implies that `normalized_ts != non_normalized_ts` even if `normalized_ts == non_normalized_ts.normalized()`.
 #[cfg(feature = "std")]
@@ -501,6 +530,12 @@ impl<'a> TypeUrl<'a> {
 
         Some(Self { full_name })
     }
+}
+
+/// Compute the type URL for the given `google.protobuf` type, using `type.googleapis.com` as the
+/// authority for the URL.
+fn type_url_for<T: Name>() -> String {
+    format!("type.googleapis.com/{}.{}", T::PACKAGE, T::NAME)
 }
 
 #[cfg(test)]
@@ -825,6 +860,22 @@ mod tests {
                 case.0,
             );
         }
+    }
+
+    #[test]
+    fn check_any_serialization() {
+        let message = Timestamp::from(UNIX_EPOCH);
+        let any = Any::from_msg(&message).unwrap();
+        assert_eq!(
+            &any.type_url,
+            "type.googleapis.com/google.protobuf.Timestamp"
+        );
+
+        let message2 = any.to_msg::<Timestamp>().unwrap();
+        assert_eq!(message, message2);
+
+        // Wrong type URL
+        assert!(any.to_msg::<Duration>().is_err());
     }
 
     #[test]
--- a/prost-types/src/lib.rs
+++ b/prost-types/src/lib.rs
@@ -25,6 +25,7 @@ use core::str::FromStr;
 use core::time;
 
 use prost::alloc::format;
+use prost::alloc::string::String;
 use prost::alloc::vec::Vec;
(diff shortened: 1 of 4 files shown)


prost-types/src/lib.rs
```diff
@@ -33,6 +37,39 @@ pub use protobuf::*;
...
+
+    /// Decode the given message type `M` from [`Any`], validating that it has
+    /// the expected type URL.
+    pub fn to_msg<M>(&self) -> Result<M, DecodeError>
+    where
+        M: Default + Name + Sized,
+    {
+        let expected_type_url = M::type_url();
+
+        if self.type_url == expected_type_url {
```
tarcieri, line 60: This check is naive. Apparently we should split at "/" and compare the fullnames afterward to see if they match

tarcieri:
@LucioFranco I'm trying to interop with a Go reference implementation which makes pretty extensive use of `Any`, though that said, I am also trying to figure out which things are standard conventions and which are idioms of my particular implementation, as we observed in #858.

That said you did make me take a second look and I did notice something a bit problematic in the type URL comparison. I'm not sure how other implementations handle the leading domain component of a type URL, i.e. if it mismatches, should the full type URL mismatch. Should type URLs with leading `/` be treated as "relative"?

All that said, I deliberately made `type_url` a provided method so it implements a convention but can be overridden if desired.

We might also consider an associated constant for the domain component, which could be blank.

src/name.rs
```diff
@@ -0,0 +1,28 @@
+//! Support for associating type name information with a [`Message`].
+
+use crate::Message;
+use alloc::{format, string::String};
+
+/// Associate a type name with a [`Message`] type.
+pub trait Name: Message {
```
LucioFranco, line 7: Can this just live in `prost_types`, why does it need to live in the core crate?

LucioFranco:
@tarcieri this looks good besides my one question, I am okay with merging mostly because a lot of this code can be kept outside the core crate.

vitruvvius:
nice !!!

src/name.rs
tarcieri: I was hoping it could eventually be autogenerated by `prost-build`, at least optionally. We're hand annotating them right now and it's getting quite cumbersome.

It seems like `prost-build` currently only refers to types/traits in `prost`. If it were optional and it could refer to types/traits in `prost-types`, I'd be fine with the `Name` trait living there.

prost-types/src/lib.rs
tarcieri: I added more robust type URL matching in [01fde2d](https://github.com/tokio-rs/prost/pull/896/commits/01fde2daebb7e75d5376a560ee50ac900119890c)

tarcieri:
Build failure seems unrelated:

https://github.com/tokio-rs/prost/actions/runs/5981227819/job/16228748609?pr=896#step:7:112

```
error: package `petgraph v0.6.4` cannot be built because it requires rustc 1.64 or newer, while the currently active rustc version is 1.63.0
Error: Process completed with exit code 101.
```

tarcieri:
Regardless of what crate `Name` ends up living in, I think it'd also be good to impl this trait for the `Message` types in `prost-types`.

That would also make it possible to test `Any::{from_msg, to_msg}`.

tarcieri:
Aside from the question of where the `Name` trait should live, this should be ready to merge

src/name.rs
```diff
@@ -0,0 +1,28 @@
...
+
+    /// Full name of this message type containing both the package name and
+    /// type name, e.g. `google.protobuf.TypeName`.
+    fn full_name() -> String {
+        format!("{}.{}", Self::NAME, Self::PACKAGE)
+    }
+
+    /// Type URL for this message, which by default is the full name with a
+    /// leading slash, but may also include a leading domain name, e.g.
+    /// `type.googleapis.com/google.profile.Person`.
```
tarcieri, line 24: Perhaps this could include some of the documentation I added to the internal `TypeUrl` type.

src/name.rs
```diff
@@ -0,0 +1,28 @@
...
+    fn type_url() -> String {
+        format!("/{}", Self::full_name())
```
tarcieri, line 26: This could potentially benefit from an easier way to customize the URL authority (i.e. hostname+port). It was a bit tricky to do this for the "well known" types:

https://github.com/tokio-rs/prost/pull/896/files#diff-fe4eba36fffb306e25d27bf29ddea9cd1634ea4ad86705045c1cb13a3aa27346R536-R540

src/name.rs
LucioFranco: I see your point, do we just want to go ahead and do the work to also add the prost-build parts? The next release will be breaking and I would be ok if we add something like that.

src/name.rs
```diff
@@ -0,0 +1,28 @@
...
+    /// e.g. `TypeName`.
+    const NAME: &'static str;
+
+    /// Package name this message type is contained in. They are domain-like
+    /// and delimited by `.`, e.g. `google.protobuf`.
+    const PACKAGE: &'static str;
+
+    /// Full name of this message type containing both the package name and
+    /// type name, e.g. `google.protobuf.TypeName`.
+    fn full_name() -> String {
```
LucioFranco, line 18: Should these be `&'static str`, is there a case where these are not statically defined with the generated proto?

LucioFranco:
@tarcieri sorry for dragging this a long, would you be interested in adding the prost-build part too then we can keep the trait in prost. I merged the msrv fix in another PR so would be good to back that out of this one. We could also merge this part now (once the msrv is backed out and this is rebased) and then we can add the prost-build stuff in a follow up.

src/name.rs
tarcieri: If `prost-build` filled them in they potentially could be all computed at build-time.

Otherwise it's not really possible to construct a `&'static str` as a concatenation of `Self::PACKAGE` + `'.'` + `Self::NAME` AFAIK.

tarcieri:
I can attempt to add the `prost-build` part, although I'm not sure how difficult that will be. Hopefully it has all of the needed information a priori

LucioFranco:
@tarcieri ok cool, let me know if I can help (feel free to ping on discord). I think I want to wait on this for the next (breaking) release which will basically happen after we get this merged.

LucioFranco:
I'd like to get things merged and released today so I will go ahead and get this merged and figure out what else we need today and then we can add more stuff in point releases since they shouldn't be breaking changes.

LucioFranco (approved):

sillykelvin:
Sorry to disturb, is there any progress on the `prost-build` part? Currently we are writing `impl prost::Name for Foo { ... }` manually, however it's awful and error-prone, it would be great if we can do this with `prost-build`.

LucioFranco:
@tarcieri were you able to follow up on this? If not if anyone else is interested in picking this up?

tarcieri:
Not yet, and I'm on vacation. Would be great if someone else could pick it up.

LucioFranco:
@tarcieri enjoy! Sounds good!

Well if someone wants to pick up this work I opened https://github.com/tokio-rs/prost/issues/926 and we can continue discussing there.

Missjones2 (approved):

Missjones2 (approved):

PR #900, opened 2023-08-27
fix protoc install link

calebcartwright:
The upstream change https://github.com/protocolbuffers/protobuf/commit/3c3dfe0efa859563e1fdbe51badd0ebf74760ff9 broke the anchor link

This corrects the link


calebcartwright:
CI failure seems unrelated:

```terminal
error: package `indexmap v2.0.0` cannot be built because it requires rustc 1.64 or newer, while the currently active rustc version is 1.63.0
```

Issue #399, opened 2020-12-02
Make it easy to write forward-compatible code: struct builders?


wchargin:
[In protobuf, adding a field to a message is considered a backward
compatible change.][1] It’s compatible at the wire format level and also
at the source level for many sets of bindings, including the standard
bindings for Python, Go, C++, and Java.

[1]: https://google.aip.dev/180#adding-components

Prost makes it easy to directly initialize structs of message type:

```rust
// message Person { string name = 1; string email = 2; }
let person = proto::Person {
    name: "A. U. Thor".into(),
    email: "author@example.com".into(),
};
```

But then if a new type is added to the message, the Rust code fails to
compile, because it’s no longer initializing all the fields.

One way to achieve compatibility is to spread in `..Default::default()`:

```rust
let person = proto::Person {
    name: "A. U. Thor".into(),
    email: "author@example.com".into(),
    ..Default::default() // more fields may be added later
};
```

This works, but (a) you have to remember to do it every time, and (b) it
can get a bit spammy when there are multiple layers of nesting:

```rust
            ..Default::default()
        },
        ..Default::default()
    },
    ..Default::default()
}
```

It would be nice to have a way to guarantee better compatibility. One
approach would be some kind of optional builder API:

```rust
let person = proto::Person::builder()
    .name("A. U. Thor".into())
    .email("author@example.com".into())
    .build();
```

I don’t know what exactly the right ergonomics are here, especially
around nested fields. (I half-wonder if the builder should take a
closure, but I think that that might get a bit hairy.)

A nice side-effect of such a builder pattern would be that the methods
could take `impl Into<FieldType>`, so you could pass `&str` literals
rather than having to write `.to_string()`/`.into()` everywhere.

An option to set `#[non_exhaustive]` on structs doesn’t sound quite
right to me, since (a) that doesn’t do anything inside the same crate
and (b) then you can’t use functional record updates (`..default()`).

It’s fine with me if the struct fields remain public; we can have a
style recommendation that says “always use builders when building
messages of a type that you don’t control”, or something. Though being
able to control that via a codegen option would be nice, too.

This is an important consideration for using Prost in shared or large
codebases, especially monorepos. It’s common to have a shared protobuf
definition with autogenerated bindings that many teams depend on.
Enabling the proto owners to make forward-compatible changes without
breaking other teams’ builds would be a requirement to use Prost.

danburkert:
[...]

PR #901, opened 2023-08-28
Extend prost-build to generate message builders

mzabaluev:
First crack at #399


mzabaluev:
Notes on the present implementation:

* Each message gets an associated `builder` function to start off the builder.
* The builder type for a message is generated in the message-named module alongside the nested types in the protobuf.
  It is named "{message_name}Builder" to reduce possibility for name conflicts.
* The final method of the builder is named `build`, but I intend to make it a configurable default.
* The field setter methods of the builder use `From` and `FromIterator` conversions on the argument.
  - For primitive types, this allows passing any values where the conversion is lossless as per the convention of `From`.
  - For boxed fields, `impl Into<Box<Foo>>` provides the convenience of being able to pass `Foo` or `Box<Foo>`.
  - For repeated fields, the `FromIterator` bound allows population from any iterable object. For example, a `repeated string` field can be initialized from an array of string literals.
  - For message fields, the conversion can be useful if the application provides a Rust-idiomatic "domain type" with a `From` conversion to the generated message struct.
* The generated code assumes that the Rust type of every field in the message struct has a `Default` impl. As far as I'm aware, this is true for anything generated from proto3.

caspermeijn (commented):
I think this PR is difficult to review. Could you do these things to make it easier to review?

- Can you explain why this APi is better then a generic builder? For example: https://crates.io/crates/derive_builder
- Add documentation about how to use this builder.
- Reorganize the commits so that they change a single thing with a description of what and why you do that. If that commit generally improves the codebase, consider opening a separate PR for just that commit. This allows me to review each commit individually.

- I am not sure about this one: Is it beneficial to split the build code generator into a separate source file?

mzabaluev:
Thanks for the feedback @caspermeijn! I have rebased the branch and squashed the changes into a few topical commits.

> Can you explain why this APi is better then a generic builder? For example: https://crates.io/crates/derive_builder

I have added an explanation in the description comment.

> Add documentation about how to use this builder.

I mean to do it, but first I'd like to settle on a more flexible way to configure the builders.
What do you think of a way to selectively generate the builders for matching paths, and be able to configure the name to satisfy author's stylistic preferences and avoid name conflicts?

```rust
prost_buid::Config::new()
    // .mypackage.Foo gets foo::Builder and the associated Foo::builder() function
    .builders(".mypackage", "Builder")
    // ...
```

> I am not sure about this one: Is it beneficial to split the build code generator into a separate source file?

I was hesitant to make a separate module for this, as hardly any functionality used in `code_generator.rs` had been broken out at the time, and there is some code reuse and similarities with other generator functions. But I can reorganize if you think it's better.

Also, I feel that the builders should be a core feature, because as discussed in #399, this together with `#[non_exhaustive]` brings prost-build into compliance with Protobuf conventions on evolving message definitions, where adding a field should not result in breaking code that uses the generated bindings.

mzabaluev:
@caspermeijn
> Reorganize the commits so that they change a single thing with a description of what and why you do that. If that commit generally improves the codebase, consider opening a separate PR for just that commit.

I could try to split off a PR with the refactoring changes in `code_generator.rs` that I needed to reuse data in the builder code generation, if that makes reviewing easier for you.

mzabaluev:
> I could try to split off a PR with the refactoring changes in `code_generator.rs` that I needed to reuse data in the builder code generation, if that makes reviewing easier for you.

Done in #1011 and rebased on top.

jregistr:
What else does this PR need to be merged?

prost-build/src/code_generator.rs
```diff
@@ -676,6 +696,233 @@ impl<'a> CodeGenerator<'a> {
...
+        self.buf.push_str(") -> Self {\n");
+        self.depth += 1;
+
+        self.push_indent();
+        self.buf.push_str("self.inner.");
+        self.buf.push_str(&rust_name);
+        self.buf.push_str(" = ");
+        if repeated {
+            self.buf
+                .push_str("value.into_iter().map(Into::into).collect();\n");
```
caspermeijn, line 883: A repeated type is a `Vec`, but this setter doesn't allow the direct setting of a `Vec`. It will reallocate. I think the setter should just use `Vec` as its input type.

prost-build/src/code_generator.rs
```diff
@@ -676,6 +696,233 @@ impl<'a> CodeGenerator<'a> {
...
+        } else if optional {
+            self.buf.push_str("Some(value.into());\n")
```
caspermeijn, line 885: The setter for optional fields should accept a `Option`. For "normal" fields I think it is good to forgo the `Option`.

prost-build/src/config.rs
```diff
@@ -608,6 +609,41 @@ impl Config {
...
+    /// Generate the [builder pattern] API for matched message types.
+    ///
+    /// [builder pattern]: https://rust-lang.github.io/api-guidelines/type-safety.html#c-builder
+    ///
+    /// # Arguments
+    ///
+    /// **`path`** - a path matching any number of types. It works the same way as in
+    /// [`btree_map`](#method.btree_map), just with the field name omitted.
+    ///
+    /// **`builder_name`** - A name for the builder type. The struct with this name
```
caspermeijn, line 621: The path can match multiple messages. I think the `builder_name` should be a suffix so that it doesn't generate multiple builders with the same name.

tests-2015/Cargo.toml
```diff
@@ -11,6 +11,7 @@ edition = "2015"
 build = "../tests/src/build.rs"
 
 [lib]
+name = "tests"
```
caspermeijn, line 14: Why is this needed? It is not related to builders.

tests/src/builders.rs
```diff
@@ -0,0 +1,139 @@
+//!#[doc(hidden)]
+
+/// # Doc tests
```
caspermeijn, line 3: Why do you use hidden doc tests?

tests/src/lib.rs
```diff
@@ -24,6 +24,7 @@ cfg_if! {
     }
 }
 
+pub mod builders;
```
caspermeijn, line 27: I would like to see an additional test: generate a builder for TestAllTypesProto3 and write that to file. This has two purposes:
- It adds a example of all posible setters to the PR
- It forces changes to builder API to be visible in future PRs.

The repo has examples for a test writing to file and asserting nothing changes: `test_generate_no_empty_outputs`

caspermeijn (changes requested):

prost-build/src/code_generator.rs
mzabaluev: I consider the setter a convenience method. The field is public, so there is always an option of setting it directly rather than going through `FromIterator`.

When specialization is stabilized in future Rust, the setter can be specialized for `Vec` without breaking the API.

prost-build/src/code_generator.rs
mzabaluev: I applied the same logic as for repeated fieds: the setter is complementary and allows shorthand code, while there is the option of assigning the field directly.

mzabaluev: Thinking of the most common use case here, where the application has a valid value to set. It should be rare to need to explicitly set a field to `None`, that is not addressed by `Default` impls on the struct and the `Option` field type.

prost-build/src/config.rs
mzabaluev: The builder is generated inside the submodule named after each message, so a collision could only occur with the message's nested types and enums. This should be rare enough that it can be solved by a specific matcher for messages where it occurs.

Stylistically, I prefer `my_message::Builder` to `my_message::MyMessageBuilder` or `self::MyMessageBuilder`.

tests/src/builders.rs
mzabaluev: These are `compile_fail` test cases and other tests of generator behavior and corner cases that are of no interest to a normal library user.

Since the tests crate is `publish = false`, this should not matter much though.

tests-2015/Cargo.toml
mzabaluev: This is needed so that the doctests I have added to the `tests` crate also work in these shim crates.

prost-build/src/code_generator.rs
mzabaluev: On the other hand, maybe the case of an explicitly optional enum field (as opposed to a "normal" field which must have a default for the uninitialized case) is special enough that the developers meant it to be used this way. I will change to it to `Option`.

tests-2015/Cargo.toml
```diff
@@ -11,6 +11,7 @@ edition = "2015"
...
 doctest = false
```
mzabaluev, line 15: There is `doctest = false` in both crates, but doctests are compiled by `cargo test` even without the `--doc` option.

prost-build/src/code_generator.rs
mzabaluev: An `impl Into<Vec<T>>` argument looks like a reasonable compromise: apart from `Vec<T>` with zero copying,  it accepts `&[T]`, `&[T; N]`, and `[T; N]`, so it should cover the most common cases without extra conversions. To set from an iterator, the caller would have to call `.collect::<Vec<_>>()`

mzabaluev: Too much hidden allocation can be bad indeed, let's not go overboard with ergonomics here.

Perhaps a different method, named e.g. `{field_name}_from_iter` could be additionally generated for each repeated field.

mzabaluev: I have resolved this with a helper type to set repeated fields in builder methods: `builder::RepeatedField`.
The only case where it currently provides convenience over just setting a `Vec` is repeated enum fields, but I think it would be good to have a uniform and more future-proof API here.

PR #909, opened 2023-09-08
chore: Use taiki-e/install-action to setup cargo-machete

tottoto:
`taiki-e/install-action` supports `cargo-machete` at [2.16.0](https://github.com/taiki-e/install-action/blob/v2.16.0/CHANGELOG.md#2160---2023-08-19). This action downloads the binary release.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,10 +20,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: dtolnay/rust-toolchain@stable
-      - name: install cargo-machete
-        uses: baptiste0928/cargo-install@v2
-        with:
-          crate: cargo-machete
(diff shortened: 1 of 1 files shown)


caspermeijn (commented):
Is it also possible to replace this cargo install? https://github.com/tokio-rs/prost/blob/a8e8c9db3a1ff5ce51a02f93d6f285901dff7391/.github/workflows/ci.yml#L111-L114

tottoto:
What pros and cons do you think to do that?

caspermeijn:
> What pros and cons do you think to do that?

I was thinking you are replacing `baptiste0928/cargo-install` with another generic cargo installer. But I now understand `taiki-e/install-action` installs specific tools and `cargo-no-std-check` is not part of that list.

caspermeijn (approved):
LGTM

caspermeijn:

PR #912, opened 2023-09-10
prost-type: impl Hash for Duration

QuentinPerez:
👋 

Our proxy, which checks if we are not making the same RPC with the same payload at the same time, requires the implementation of a hash.


QuentinPerez:
@LucioFranco do you think that this patch can be merged ?

LucioFranco (approved):

Issue #905, opened 2023-09-04
Type error regression when using enum called Result


thedjinn:
I am working on a project that uses Prost (via Tonic) to compile a protobuf file for use with gRPC. The file compiles fine with prost 0.11.9, but when switching to prost 0.12.0 the generated Rust file contains type errors.

The minimal protobuf file required to reproduce the problem is this one:

```proto
syntax = "proto3";

package fail;

enum Result {
  HELLO = 0;
}

message FailMessage {
  Result result = 1;
}
```

When attempting to compile the generated rust file the following errors are returned:

```
error[E0107]: enum takes 0 generic arguments but 2 generic arguments were supplied
 --> /private/tmp/tonicfail/target/debug/build/tonicfail-ca0c3cb604ffea00/out/fail.rs:2:28
  |
2 | #[derive(Clone, PartialEq, ::prost::Message)]
  |                            ^^^^^^^^^^^^^^^^- help: remove these generics
  |                            |
  |                            expected 0 generic arguments
  |
note: enum defined here, with 0 generic parameters
 --> /private/tmp/tonicfail/target/debug/build/tonicfail-ca0c3cb604ffea00/out/fail.rs:9:10
  |
9 | pub enum Result {
  |          ^^^^^^
  = note: this error originates in the derive macro `::prost::Message` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0308]: mismatched types
   --> /private/tmp/tonicfail/target/debug/build/tonicfail-ca0c3cb604ffea00/out/fail.rs:2:28
    |
2   | #[derive(Clone, PartialEq, ::prost::Message)]
    |                            ^^^^^^^^^^^^^^^^
    |                            |
    |                            expected `Result`, found `Result<_, _>`
    |                            expected due to this
    |
    = note: `std::result::Result<_, _>` and `proto::Result` have similar names, but are actually distinct types
note: `std::result::Result<_, _>` is defined in crate `core`
   --> /Users/djinn/.rustup/toolchains/stable-x86_64-apple-darwin/lib/rustlib/src/rust/library/core/src/result.rs:502:1
    |
502 | pub enum Result<T, E> {
    | ^^^^^^^^^^^^^^^^^^^^^
note: `proto::Result` is defined in the current crate
   --> /private/tmp/tonicfail/target/debug/build/tonicfail-ca0c3cb604ffea00/out/fail.rs:9:1
    |
9   | pub enum Result {
    | ^^^^^^^^^^^^^^^
[...]

PR #917, opened 2023-09-14
Fixes 905, fix: Use full path of Result type + unit test + Option unti test

anstylian:
Same as PR #916 plus one commit that contains a unit test that covers the case  where `Option` was used instead of `::core::option::Option`.

PR #916  description: 
this pull request contains 3 commits:
1. [add cmake and ninja in the flake](https://github.com/tokio-rs/prost/commit/7b15ffd0d1774f78244e3a35068f657f09723331): Was needed to run cargo test from the test directory
2. [unit-test added to replicate issue](https://github.com/tokio-rs/prost/commit/30c800d03f75e376293a49e1271760d8985ff418): Replicates the error described in issue #905 
3. [fix: Use full path of Result type](https://github.com/tokio-rs/prost/commit/16e7e35865a16cfa8dc877c1986d0bffda0944eb): Fixes issue
https://github.com/tokio-rs/prost/issues/905

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -14,7 +14,7 @@
       {
         devShells.default = pkgs.mkShell {
           packages = with pkgs; [ cargo rustc ];
-          buildInputs = with pkgs; [ pkg-config protobuf curl ];
+          buildInputs = with pkgs; [ pkg-config protobuf curl cmake ninja ];
         };
       });
 }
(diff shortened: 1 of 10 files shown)


LucioFranco (approved):
LGTM! Thanks!

PR #919, opened 2023-09-16
chore: Add dep: prefix to feature dependencies

tottoto:
Adds `dep:` prefix to feature dependencies.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,6 +43,7 @@ bench = false
 
 [features]
 default = ["prost-derive", "std"]
+prost-derive = ["dep:prost-derive"]
 no-recursion-limit = []
 std = []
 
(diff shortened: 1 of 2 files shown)


caspermeijn (approved):
Please rebase to latest master to fix CI

tottoto:
Rebased.

caspermeijn (approved):
Looks good to me

Cargo.toml
```diff
@@ -43,6 +43,7 @@ bench = false
 
 [features]
 default = ["prost-derive", "std"]
+prost-derive = ["dep:prost-derive"]
```
LucioFranco, line 46: Can we call this feature `derive`?

caspermeijn: That will break existing users of the feature, right?

LucioFranco: You're right...this is fine for now then...I didn't know cargo exposed this implicitly outside the crate itself.

LucioFranco (approved):

PR #923, opened 2023-09-18
fix: generate `Name::full_name` properly

NBonaparte:
`full_name()` seems to generate the full name in the wrong order (given the examples in the doc comments, `TypeName.google.protobuf` instead of `google.protobuf.TypeName`).


tarcieri:
Would it be possible to get a point release with this?

PR #931, opened 2023-10-11
feat: Generate implementation for `Name` trait

### LINKED ISSUE #926: Follow up, implement `Name` trait implementation via `prost-build`
Follow up work on https://github.com/tokio-rs/prost/pull/896 to generate the prost-build portion of this implementation. We should generate the impl based on the filedescriptorset values.

tinrab:
Adds an option to implement the `Name` trait (added in #896) for message types.

Domains names for type URLs can be set with `type_name_domains`.
It's possible that this doesn't have to be a `PathMap`, but a single value for the entire config. Or something completely different.

Resolves #926.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,7 +33,7 @@ jobs:
       matrix:
         toolchain:
           - stable
-          - "1.64"
+          - "1.65"
         os:
           - ubuntu-latest
           - macos-latest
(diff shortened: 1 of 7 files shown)


getong:
I test your code, it works. Thanks for your time.

getong:
@tinrab 
I think you can update ci.yaml version from 1.64 to 1.65, which might test ok.
You can see my test here : https://github.com/getong/prost/pull/2

getong:
@LucioFranco 
The test is ok now, would you take a look at this pr and push a new version to crates.io?

LucioFranco (approved):
Nice! Thanks!

tinrab:
@LucioFranco @getong Turns out, this doesn't work. Declaring `PACKAGE` on module-level causes name collision.

PR #944, opened 2023-11-17
Fix code generation for Name trait

tinrab:
Fixes a bug in #931.

In code generator, `package` field was used to track nesting of messages and didn't represented the protobuf's "package" name. Nested names are now contained in `type_path` and `package` refers to file's package.


BKDaugherty:
Closes #946 .

Thanks @tinrab for both the quick fix and adding the feature to `prost-build` in the first place!

LucioFranco (approved):
Thanks!

PR #950, opened 2023-11-24
feat: derive Copy trait for messages where possible

caspermeijn:
Rust primitive types can be copied by simply copying the bits. Rust structs can also have this property by deriving the Copy trait.

Automatically derive Copy for:
- messages that only have fields with primitive types
- the Rust enum for one-of fields
- messages whose field type are messages that also implement Copy

Generated code for Protobuf enums already derives Copy.

### DIFF
--- a/prost-build/src/code_generator.rs
+++ b/prost-build/src/code_generator.rs
@@ -190,7 +190,12 @@ impl<'a> CodeGenerator<'a> {
         self.buf
             .push_str("#[allow(clippy::derive_partial_eq_without_eq)]\n");
         self.buf.push_str(&format!(
-            "#[derive(Clone, PartialEq, {}::Message)]\n",
+            "#[derive(Clone, {}PartialEq, {}::Message)]\n",
+            if self.message_graph.can_message_derive_copy(&fq_message_name) {
+                "Copy, "
+            } else {
+                ""
+            },
             prost_path(self.config)
         ));
         self.append_skip_debug(&fq_message_name);
@@ -595,8 +600,14 @@ impl<'a> CodeGenerator<'a> {
         self.push_indent();
         self.buf
             .push_str("#[allow(clippy::derive_partial_eq_without_eq)]\n");
+
+        let can_oneof_derive_copy = fields.iter().map(|(field, _idx)| field).all(|field| {
+            self.message_graph
+                .can_field_derive_copy(fq_message_name, field)
+        });
         self.buf.push_str(&format!(
-            "#[derive(Clone, PartialEq, {}::Oneof)]\n",
+            "#[derive(Clone, {}PartialEq, {}::Oneof)]\n",
+            if can_oneof_derive_copy { "Copy, " } else { "" },
             prost_path(self.config)
         ));
         self.append_skip_debug(fq_message_name);
--- a/prost-build/src/fixtures/field_attributes/_expected_field_attributes.rs
+++ b/prost-build/src/fixtures/field_attributes/_expected_field_attributes.rs
@@ -23,12 +23,12 @@ pub struct Foo {
     pub foo: ::prost::alloc::string::String,
 }
 #[allow(clippy::derive_partial_eq_without_eq)]
-#[derive(Clone, PartialEq, ::prost::Message)]
+#[derive(Clone, Copy, PartialEq, ::prost::Message)]
 pub struct Bar {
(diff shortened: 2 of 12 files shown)


gibbz00:
Would it be possible to make this configurable? I personally like to avoid automatic Copy derives to avoid implicit cloning.

caspermeijn:
> Would it be possible to make this configurable? I personally like to avoid automatic Copy derives to avoid implicit cloning.

What do you mean with implicit cloning? The trait `Copy` means that the bytes can simply be copied to create a second object.

The docs says: When should my type be Copy? Generally speaking, if your type can implement Copy, it should.

https://doc.rust-lang.org/std/marker/trait.Copy.html

gibbz00:
This example is perhaps trivial, but my experience is that these things can become really tricky to pin down in large projects.

```rust
fn main() {
    let copyable = 1;
    read(&copyable);
}

fn read(x: &u32) {
    update(*x); // implicit clone
    dbg!(x); // 1
}

fn update(mut x: u32) {
    x += 1;
    dbg!(x); // 2
}
```

caspermeijn:
>     update(*x); // implicit clone

I would not call that an implicit clone, as you explicitly deref `x`. With a deref, you choose to copy the value from behind the reference to a local variable. So with `*` you choose to copy, am I right?

https://micouy.github.io/rust-dereferencing/

gibbz00:
Well things get a bit tricky because derefmut and deref both use the same symbol `*`.

Sure, my example was maybe not the best. I'm also just playing a bit of devil's advocate. 

There's still the basic case of implicit cloning mentioned in [the link you posted above](https://doc.rust-lang.org/std/marker/trait.Copy.html#whats-the-difference-between-copy-and-clone):

```rust
#[derive(Copy, Clone)]
struct GiantStruct(u32);

fn main() {
    let copyable0 = GiantStruct(1);
    let copyable1 = copyable0;
}
```

caspermeijn:
> There's still the basic case of implicit cloning mentioned in [the link you posted above](https://doc.rust-lang.org/std/marker/trait.Copy.html#whats-the-difference-between-copy-and-clone):

Well, that is how the language is supposed to work, right? If you want two variables pointing to the same instance, then you want to take a reference: `let copyable1 = &copyable0;`. 

Sorry, but I don't understand in what situation you don't want an `impl Copy` for your types. From my perspective, it is always useful.

gibbz00:
You're probably right. I usually want the compiler to tell me that I must copy things if that's what I'm trying to do.

The Rust API Guidelines book also recommends implementing `Copy` for public types: https://rust-lang.github.io/api-guidelines/interoperability.html?highlight=copy#types-eagerly-implement-common-traits-c-common-traits

LucioFranco (approved):
Really nice work! I like this a lot. Do you think we should release this as part of a breaking change or not?

caspermeijn:
> Really nice work! I like this a lot. Do you think we should release this as part of a breaking change or not?

I agree this is a breaking change.

ShaneMurphy2:
I know this is closed, but I think this is absolutely not a change that should be automatically implemented. The trait Copy is _not_ an auto trait for a reason. It carries a semantic choice about whether your type has move semantics or copy semantics. This choice is opt-in at the language level by design. Rust is a language about explicitness, and implicitly making a choice during codegen that inserts implicit behavior is antithetical to that principle. There is a [great article](https://smallcultfollowing.com/babysteps/blog/2024/06/21/claim-auto-and-otherwise/) about Copy/Clone that outlines how in today's Rust, Copy is really two ideas in one. One is what you mentioned that it is "a bitwise copy", the other is "this type has value semantics". Those are (and should in the language) separate concepts.

LucioFranco:
@ShaneMurphy2 would be helpful if you could call out specific cases where prost is generating a copy impl for a type that isn't trivially copyable.

ShaneMurphy2:
> @ShaneMurphy2 would be helpful if you could call out specific cases where prost is generating a copy impl for a type that isn't trivially copyable.

That isn't the issue. The type is trivially copyable. Being trivially copyable and "having value semantics" are independent concepts. If I want move semantics for my type, because instances of it are not interchangeable (one definition of value semantics), then I should be able to have that for my type. It is a semantic property ascribed to the type separate from the mechanical property of "you can make a clone by doing a bitwise memcpy".

PR #951, opened 2023-11-25
better checking of tag duplicates, avoid discarding invalid variant errs

mumbleskates:
Currently it seems that the "invalid oneof variant" Errs are always discarded because they are being passed directly into `flat_map` from a closure, rather than actually propagating the error.


prost-derive/src/lib.rs
```diff
@@ -414,23 +412,20 @@ fn try_oneof(input: TokenStream) -> Result<TokenStream, Error> {
...
+    }
+    if let Some((duplicate_tag, _)) = fields
         .iter()
-        .flat_map(|&(ref variant_ident, ref field)| -> Result<u32, Error> {
-            if field.tags().len() > 1 {
-                bail!(
-                    "invalid oneof variant {}::{}: oneof variants may only have a single tag",
-                    ident,
-                    variant_ident
-                );
```
mumbleskates, line 425: the `Err` this returns is formerly always discarded via flattening the `Result`.

prost-derive/src/lib.rs
```diff
@@ -88,19 +87,17 @@ fn try_message(input: TokenStream) -> Result<TokenStream, Error> {
...
-        .collect::<Vec<_>>();
-    let num_tags = tags.len();
-    tags.sort_unstable();
-    tags.dedup();
-    if tags.len() != num_tags {
-        bail!("message {} has fields with duplicate tags", ident);
-    }
+        .flat_map(|(_, field)| field.tags())
+        .sorted_unstable()
+        .tuple_windows()
```
mumbleskates, line 97: it would kinda be preferable to write this as `.duplicates().next()` using `Itertools`, but that method unfortunately requires `use_std`.

prost-derive/src/lib.rs
```diff
@@ -88,19 +87,17 @@ fn try_message(input: TokenStream) -> Result<TokenStream, Error> {
...
+        .find(|(a, b)| a == b) {
```
mumbleskates, line 98: it would kinda be preferable to write this as `.duplicates().next()` using `Itertools`, but that method unfortunately requires `use_std`.

prost-derive/src/lib.rs
```diff
@@ -88,19 +87,18 @@ fn try_message(input: TokenStream) -> Result<TokenStream, Error> {
...
-    let num_tags = tags.len();
-    tags.sort_unstable();
-    tags.dedup();
-    if tags.len() != num_tags {
-        bail!("message {} has fields with duplicate tags", ident);
-    }
+        .flat_map(|(_, field)| field.tags())
+        .sorted_unstable()
+        .tuple_windows()
+        .find(|(a, b)| a == b)
```
mumbleskates, line 98: it would kinda be preferable to write this as `.duplicates().next()` using `Itertools`, but that method unfortunately requires `use_std`.

prost-derive/src/lib.rs
mumbleskates: i believe the intention here was originally to write this as `.map(..).collect::<Result<Vec<_>, _>>()?;`, but the nuance was lost

prost-derive/src/lib.rs
```diff
@@ -7,8 +7,7 @@ extern crate proc_macro;
 
 use anyhow::{bail, Error};
 use itertools::Itertools;
-use proc_macro::TokenStream;
-use proc_macro2::Span;
+use proc_macro2::{Span, TokenStream};
```
caspermeijn, line 10: Why is `proc_macro2::TokenStream` better?

prost-derive/src/lib.rs
caspermeijn: As I understand prost_derive will run on the host, so it is not subject to `no_std` rules. The code does also `use std::fmt;`, so `std` must be available. So I think it is safe to `use_std` for Itertools. Do you agree?

prost-derive/src/lib.rs
```diff
@@ -91,16 +90,19 @@ fn try_message(input: TokenStream) -> Result<TokenStream, Error> {
...
-    tags.dedup();
-    if tags.len() != num_tags {
-        bail!("message {} has fields with duplicate tags", ident);
-    }
+        .sorted_unstable()
+        .tuple_windows()
+        .find(|(a, b)| a == b)
+    {
+        bail!(
+            "message {} has multiple fields with tag {}",
```
caspermeijn, line 101: So this basically improves the error message. Do I understand correctly?

prost-derive/src/lib.rs
```diff
@@ -412,23 +414,21 @@ fn try_oneof(input: TokenStream) -> Result<TokenStream, Error> {
         }
     }
 
-    let mut tags = fields
+    if fields.iter().any(|(_, field)| field.tags().len() > 1) {
+        panic!("variant with multiple tags"); // Not clear if this is possible, but good to be safe
```
caspermeijn, line 418: This error message is a regression, right? The old message was: "invalid oneof variant {}::{}: oneof variants may only have a single tag"

prost-derive/src/lib.rs
```diff
@@ -519,10 +519,52 @@ fn try_oneof(input: TokenStream) -> Result<TokenStream, Error> {
...
+                B(bool),
+                #[prost(bool, tag = "1")]
+                C(bool),
+            }
+        ));
+        assert!(output.is_err());
+        assert_eq!(
+            output.unwrap_err().to_string(),
+            "invalid oneof Invalid: multiple variants have tag 1"
+        );
```
caspermeijn, line 568: The test for `variant with multiple tags` is missing.

caspermeijn (changes requested):
This fixes the error reporting for duplicate tags, improves the error message and adds tests. I like it.

prost-derive/src/lib.rs
mumbleskates: `proc_macro` can't be used in tests; this seems to be the main usecase for `proc_macro2`. we translate at the boundary of the `proc_macro` api itself rather than letting its untestable types penetrate all the way into our functionality.

`quote!` unconditionally produces `proc_macro2::TokenStream` as its output; the only thing that really changes with this is that the boundary at which we translate from and back into the un-testable `proc_macro::TokenStream` type changes from *just inside* the function (from `syn::parse(input)` at the top, and via `expanded.into()` at the bottom) to just *outside* the function (where the trivial `.into()` calls happen inside the function marked with `#[proc_macro_derive]`).

prost-derive/src/lib.rs
mumbleskates: yeah from what i've seen since i published this that is totally possible. i'll update it

prost-derive/src/lib.rs
mumbleskates: correct :+1:  and i feel the expression of the inadmissible condition is a little more straightforward. "deduplicated list is the same size as the original list" is perfectly valid, but it feels unnecessarily roundabout when we can find duplicates directly.

prost-derive/src/lib.rs
mumbleskates: i can re-expand this, sure

prost-derive/src/lib.rs
mumbleskates: as far as i've been able to tell, that error case is unreachable by any input. any input that tries to put multiple "tag" attrs on a variant errors earlier, during the parsing of the attributes on the field.

i've added a test to demonstrate this here.

prost-derive/src/lib.rs
```diff
@@ -412,23 +414,28 @@ fn try_oneof(input: TokenStream) -> Result<TokenStream, Error> {
         }
     }
 
-    let mut tags = fields
+    for (variant_ident, field) in &fields {
+        // Not clear if this condition is reachable since multiple "tag" attributes are already
+        // rejected, but good to be safe
```
caspermeijn, line 419: I don't like to leave this comment here. We should either:
- make sure this can't happen by validating the input, then this check can be replaced by an assertion, or
- verify that multiple tags are possible, but not allowed for oneof, then this code is good, and the comment can be removed.

It seems like the origin of the Vec is here. Can you trigger that code in a test? https://github.com/tokio-rs/prost/blob/993a6d132187a95c624bab527440e280e2f58af8/prost-derive/src/field/oneof.rs#L36

caspermeijn (changes requested):

prost-derive/Cargo.toml
```diff
@@ -20,7 +20,7 @@ proc_macro = true
 
 [dependencies]
 anyhow = "1.0.1"
-itertools = { version = ">=0.10, <=0.12", default-features = false, features = ["use_alloc"] }
-proc-macro2 = "1"
+itertools = ">=0.10.1, <=0.12"
+proc-macro2 = ">=1.0.60, <2"
```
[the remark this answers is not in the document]
mumbleskates: ah that's right, good catch.

prost-derive/src/lib.rs
mumbleskates: nope, because that's the origin for the `tags = "1, 2, 3"` attribute that appears on `oneof` members *where they appear in a message*. This case we are testing is for the *variant* fields *within* a oneof enum, which is populated here:
https://github.com/tokio-rs/prost/blob/02bc13232e5a3d9aa041e8657456ea8590992863/prost-derive/src/field/scalar.rs#L34

...and elsewhere, the same as normal scalar/map/group/message fields are parsed. dispatch occurs from up above in [field/mod.rs](https://github.com/tokio-rs/prost/blob/04be6049ea676dde5dfac8a21da2559f4f380ed2/prost-derive/src/field/mod.rs#L66-L73), and the difference between `new()` and `new_oneof()` in these cases is at most some extra validation. The only kind of field we can parse that is allowed to have multiple tags is a oneof, which is parsed via `field::oneof::Field::new()`; there is no corresponding `::new_oneof()` function, and we do not have a branch for it in the upper dispatch. Naturally this is because oneof fields cannot themselves contain other oneofs.

Since all this is kind of obnoxious and long-winded to prove, and i don't hate the assertion, i'm happy with including it in whatever way we feel fits best. I'd reduced it from the full error message we see here to a simpler `panic!` originally; we can do that again. perhaps it's semantically clearer to express that as an `assert!`: that way it can stand in the code to provide context on the operating conditions of the surrounding code, even if we're already confident it's unreachable.

```rust
assert!(fields.iter().all(|field| field.tags().len() == 1));
```
...perhaps with a brief appropriate comment about how oneofs can't directly nest.

prost-derive/Cargo.toml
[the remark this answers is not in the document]
mumbleskates: to note i don't think worrying about the age of dependencies in the proc macro crate is as important here. because this crate has no library it cannot be depended upon except by its observed behavior, so dependency versions shouldn't be able to conflict with any other binaries.

prost-derive/src/lib.rs
mumbleskates: well i spelled it wrong in that comment but i got it right in the end.

mumbleskates:
incidentally, most (all? probably, all) of my contributions to `Prost` were discovered/developed while forking and [almost completely rewriting](https://github.com/mumbleskates/bilrost) it. Downstream of here i have changed the nature of the encoding slightly (not dramatically), but i have also achieved huge strides in usability:

* fully trait-based dispatch for encoding and decoding, with derive attributes simple enough to use on existing user types and support for generics and lifetimes on message types
* ignored fields
* useful functionality on the object-safe traits
* oneofs that may bear their own empty variant rather than needing to nest within `Option`
* oneofs in messages must have known matching tags, checked at compile time
* greatly increased type support, including nesting values in ways that protobuf schemas cannot even express
* novel linear-time encoding via writing the output back-to-front, which can be dramatically faster without caching extra data inside the structs
* guaranteed field ordering (protobuf as an encoding has no real use for this)
* distinguished decoding of canonical data (protobuf can't do this as easily)
* all of this and it's actually [faster](https://github.com/djkoloski/rust_serialization_benchmark)

if any of these sound useful to upstream it could be a good project for myself or someone to trace back over my steps in achieving all these. some of these can probably be upstreamed in a day or two, others could take longer.

prost-derive/Cargo.toml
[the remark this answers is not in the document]
caspermeijn: I mention the age due to our MSRV policy, I would like to apply that policy also to the dependencies.

prost-derive/src/lib.rs
caspermeijn: Yes, that makes sense

prost-derive/src/lib.rs
caspermeijn: Thanks

caspermeijn (approved):
Thanks for your contribution

caspermeijn:
> if any of these sound useful to upstream it could be a good project for myself or someone to trace back over my steps in achieving all these. some of these can probably be upstreamed in a day or two, others could take longer.

Improvements are welcome. However, review capacity is low, so bringing in a reviewer would be appreciated.

Keep in mind that prost is focused on an unopinionated implementation of protobuf.

PR #957, opened 2023-12-13
ci: Add Dependabot

sjackman:
Enable [Dependabot](https://docs.github.com/en/code-security/dependabot) to open PRs to update dependencies.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,8 @@
+version: 2
+updates:
+  - package-ecosystem: "cargo"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    commit-message:
+      prefix: "build(deps)"


caspermeijn:
What kind of changes will this propose? 

Because prost is a library and we want to support a wide range of dependency version numbers, I don't want to always update the version numbers in Cargo.toml. 

Can dependabot add a version range like in https://github.com/tokio-rs/prost/pull/1013?

sjackman:
> Can dependabot add a version range like in https://github.com/tokio-rs/prost/pull/1013?

Yes, it will. Dependabot follows the existing form of the depenency. When the dependency is a range, like
```toml
multimap = { version = ">=0.8, <=0.10", default-features = false }
```
it will suggest increasing the upper range like so:
```toml
multimap = { version = ">=0.8, <=0.11", default-features = false }
```

caspermeijn (approved):
I have looked into it. I think the [Dependabot version updates](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/about-dependabot-version-updates) are useful for this project. The Dependabot alerts and Dependabot security updates seem less useful to me. 

As I understand it, Dependabot version updates will open PRs for new library versions. That will at least notify us that a new version is available and probably a change we want to accept. 

We can choose to ignore the PR and open new PR if the change is not what you want. We probably want to do a version range when possible to improve compatibility.

sjackman:
👏 Excellent! Thank you, Casper!

Issue #960, opened 2023-12-14
prost-build doc link broken on docs.rs


paxsonsa:
On the main page for Prost, the link to prost-build is linking to the readme path in the repo. This works on Crates and GitHub but breaks on doc.rs

Happy to submit a fix but wondering if we prefer for it to link to doc.rs or continue to link to GitHub?

paxsonsa:
This link:
https://github.com/tokio-rs/prost/blob/master/README.md?plain=1#L39

caspermeijn:
Fixed in https://github.com/tokio-rs/prost/pull/967

PR #961, opened 2023-12-15
nit: update broken prost-build documentation link

paxsonsa:
Reference Issue: #960 

This is a PR to update the Prost-build link at that top of the read me to go to the doc.rs page. On the docs.rs for prost, the the link in question is broken as its using markdown link relative to the repo. This works on GitHub and cargo.io but breaks on doc.rs which is where we probably want it to be the most reliable.

Thanks.


caspermeijn:
Fixed in https://github.com/tokio-rs/prost/pull/967

PR #875, opened 2023-06-22
build(deps): update to itertools 0.11

sjackman:
Depend on itertools >=0.10, <0.12

PR #948, opened 2023-11-22
Allow itertools 0.12

kayabaNerve:
0.12 didn't change the syntax of any APIs in-use.

Additionally, the tests do pass with 0.12.

PR #969, opened 2023-12-30
build(deps): update to itertools 0.12

yotamofek:
First, thanks for all your work on this crate!

I'm trying to dedup some deps in my project and currently prost is the only dep I have that is still on `itertools 0.11`.
I saw in #875 a mention of the MSRV needing to be bumped, but it seems this new update doesn't affect that.


yotamofek:
So it seems the culprit that caused the MSRV to accidentally be bumped is [`which-rs`](https://github.com/harryfei/which-rs/pull/81/files).
I'm not sure why bumping `itertools` surfaced or if the CI was just broken due to the newer version of `which` being released.
Anyways, I also missed #962 and #948 trying to solve these issues, sorry about the duplication.
Closing this PR now.

PR #971, opened 2024-01-12
Mark compatibility with itertools 0.12

fornwall:
Allows consumers to use less duplicated crates as they update to itertools 0.12.


fornwall:
Duplicate of https://github.com/tokio-rs/prost/pull/948 and https://github.com/tokio-rs/prost/pull/963

PR #975, opened 2024-01-30
chore: Bump msrv to 1.70

tottoto:
`home` 0.5.9 requires Rust 1.70.

### DIFF
--- a/prost-build/Cargo.toml
+++ b/prost-build/Cargo.toml
@@ -12,7 +12,7 @@ documentation = "https://docs.rs/prost-build"
 readme = "README.md"
 description = "A Protocol Buffers implementation for the Rust Language."
 edition = "2021"
-rust-version = "1.60"
+rust-version = "1.70"
 
 [features]
 default = ["format"]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,7 +33,7 @@ jobs:
       matrix:
         toolchain:
           - stable
-          - "1.65"
+          - "1.70"
         os:
           - ubuntu-latest
           - macos-latest
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ start-to-finish example.
 
 ### MSRV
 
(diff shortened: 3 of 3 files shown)


caspermeijn:
Can you give more context why you think this is needed?

For me, `cargo msrv verify` reports the MSRV is correct.

```
$ cargo msrv verify
Fetching index
Verifying the Minimum Supported Rust Version (MSRV) for toolchain x86_64-unknown-linux-gnu
Using check command cargo check
   Finished Satisfied MSRV check: 1.60.0
```

prost-build/Cargo.toml
```diff
@@ -12,7 +12,7 @@ documentation = "https://docs.rs/prost-build"
 readme = "README.md"
 description = "A Protocol Buffers implementation for the Rust Language."
 edition = "2021"
-rust-version = "1.60"
+rust-version = "1.70"
```
caspermeijn, line 15: There are actually more than one `rust-version` definitions in the code base. Can you update the others as well?

caspermeijn (commented):
After looking further into it, my previous command didn't account for the workspace. The correct command to reproduce this problem is:
```
$ cargo +1.60 check --workspace --all-features
error: package `indexmap v2.2.2` cannot be built because it requires rustc 1.63 or newer, while the currently active rustc version is 1.60.0
```

caspermeijn:
This PR can be closed as the issue is fixed in #982

tottoto:
Are there any reason to bump the msrv of the crates which do not depend on `home`?

caspermeijn:
It feels like there is too little maintainer capacity, and I think a single MSRV for the whole repo is easier to maintain. This PR doesn't actually verify the MSRV of main `prost` crate in CI.

tottoto:
> This PR doesn't actually verify the MSRV of main prost crate in CI.

Yes, and it has not been performed in this project as well so far.

Bumping the `rust-version` field make build fail the build lower than its version, and as `prost` is a widely used crate, it effect is relatively large.

caspermeijn:
Let's continue this discussion in #983

PR #978, opened 2024-02-05
feat: add str encoding helper

morrisonlevi:
This allows users to encode structs that have types like `&str`, `Box<str>`, `Cow<str>`, etc if they are willing to get their hands dirty. This more efficient than creating a parallel type that holds a `String`, copying it over, just for the purpose of encoding it, then throwing it away.

Ideally we'd support types like `Cow<'a, str>` directly but there are plenty of issues and challenges with that. I am willing to get my hands dirty, but I prefer having this helper in prost rather than copying the relevant bits into my own project.

### DIFF
--- a/src/encoding.rs
+++ b/src/encoding.rs
@@ -9,7 +9,7 @@ use alloc::format;
 use alloc::string::String;
 use alloc::vec::Vec;
 use core::cmp::min;
-use core::convert::TryFrom;
+use core::convert::{AsRef, TryFrom};
 use core::mem;
 use core::str;
 use core::u32;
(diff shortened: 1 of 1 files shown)


guilload:
Hello @morrisonlevi,

I'm interested in this feature, too, for `str` and `[u8]` slices. Two questions:
1. Why not rewrite `string::encode` directly as `fn encode<S, B>(tag: u32, value: &impl AsRef<str>, buf: &mut B)`?
2. Would you consider doing the same thing for bytes?

For bytes, all we have to do is:
1. Add an `AsRef<[u8]>` constraint on `BytesAdapter`: `trait BytesAdapter: Default + Sized + AsRef<[u8]> + 'static`
2. Remove `BytesAdapter::append_to`
3. Rewrite `bytes::encode` as:

```rust
pub fn encode<A, B>(tag: u32, value: &A, buf: &mut B)
where
        A: AsRef<[u8]>,
        B: BufMut,
    {
        let bytes = value.as_ref();
        encode_key(tag, WireType::LengthDelimited, buf);
        encode_varint(bytes.len() as u64, buf);
        buf.put_slice(bytes);
    }
```

morrisonlevi:
When I tried changing `encode` directly, tests would fail because of mismatches in expected types in a bunch of test helpers. It was easier to add this helper and keep the tests green.

guilload:
`cargo test` runs fine on my [branch](https://github.com/tokio-rs/prost/compare/master...quickwit-oss:prost:guilload/encode-str-and-bytes-slices?expand=1).

morrisonlevi:
@guilload If your tests are passing, feel free to open a PR and I'll close this one.

guilload:
Ok. I opened #979. CI is failing because of the MSRV issue, but tests are passing locally.

PR #979, opened 2024-02-09
feat: encode string and bytes slices

guilload:
This PR adds the ability to encode `&str` and `&[u8]` using `string::encode` and `bytes::encode`. This is useful for users writing custom `Message` implementations who want to avoid unnecessary allocations.

### DIFF
--- a/src/encoding.rs
+++ b/src/encoding.rs
@@ -794,13 +794,15 @@ macro_rules! length_delimited {
 pub mod string {
     use super::*;
 
-    pub fn encode<B>(tag: u32, value: &String, buf: &mut B)
+    pub fn encode<S, B>(tag: u32, value: &S, buf: &mut B)
     where
+        S: AsRef<str>,
         B: BufMut,
     {
+        let slice = value.as_ref();
         encode_key(tag, WireType::LengthDelimited, buf);
-        encode_varint(value.len() as u64, buf);
-        buf.put_slice(value.as_bytes());
+        encode_varint(slice.len() as u64, buf);
+        buf.put_slice(slice.as_bytes());
     }
     pub fn merge<B>(
         wire_type: WireType,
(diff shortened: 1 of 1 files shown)


LucioFranco:
Can you provide some motivation for this change

guilload:
@morrisonlevi provided some context in his original PR: #978.

My use case is similar. I recently wrote a few custom `Message` implementations for types with some fields for which I can obtain a `&str` or `&[u8]` for free but not a `String` or `Vec<u8>` without allocating.

guilload:
@caspermeijn, is there any chance you can look into this PR?

src/encoding.rs
```diff
@@ -876,21 +878,16 @@ pub mod string {
 pub trait BytesAdapter: sealed::BytesAdapter {}
 
 mod sealed {
-    use super::{Buf, BufMut};
+    use super::Buf;
 
-    pub trait BytesAdapter: Default + Sized + 'static {
+    pub trait BytesAdapter: Default + Sized + AsRef<[u8]> + 'static {
```
caspermeijn, line 883: BytesAdapter is not used anymore for `encode`, but it seems to still be used for `encoded_len`. Will this mismatch cause issues for other people?

src/encoding.rs
```diff
@@ -794,13 +794,15 @@ macro_rules! length_delimited {
 pub mod string {
     use super::*;
 
-    pub fn encode<B>(tag: u32, value: &String, buf: &mut B)
+    pub fn encode<S, B>(tag: u32, value: &S, buf: &mut B)
```
caspermeijn, line 797: I would like to see tests to guarantee `String` can still be used on this function.

And you need to add tests for your new use case, to make sure it is not broken in the future.

src/encoding.rs
```diff
@@ -934,26 +924,20 @@ impl sealed::BytesAdapter for Vec<u8> {
...
-    }
 }
 
 pub mod bytes {
     use super::*;
 
     pub fn encode<A, B>(tag: u32, value: &A, buf: &mut B)
     where
-        A: BytesAdapter,
+        A: AsRef<[u8]>,
```
caspermeijn, line 934: I would like to see tests to guarantee `impl BytesAdapter`, `Bytes` and `Vec<u8>` can still be used on this function.

And you need to add tests for your new use case, to make sure it is not broken in the future.

caspermeijn (commented):
This PR lacks a description. I believe you want to expand the types that are accepted by `string::encode` and `bytes::encode`.

PR #981, opened 2024-02-16
fix: fq_message_name should begin with one dot

caspermeijn:
When package name is empty, but type_path is filled, the fq_message_name should begin with a single dot.

Two duplicate implementations are moved to a separate function.

A test is added that creates a sub message without a package name. The code generator uses the message name to generate a rust module path, which is invalid.

fixes: #843


LucioFranco (approved):
LGTM thanks!

PR #982, opened 2024-02-16
chore: Bump MSRV to 1.70

caspermeijn:
Rust 1.70 is 7 month old at this time. A transient dependency `home` requires at least rust 1.70.

```
prost-build v0.12.3 (/workspaces/prost/prost-build) 
└── which v4.4.2
    └── home v0.5.9
```

PR #983, opened 2024-02-17
chore: Relax MSRV

tottoto:
Relaxes MSRV which are bumped in #982. As the use of `prost-build` can be separated to the different build process, it is too strict to set the same MSRV of `prost-build` to the other crates for users. `cargo-hack`'s `--rust-version` option can be used to perform checking different MSRVs on the project crates.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,7 +40,7 @@ jobs:
       matrix:
         toolchain:
           - stable
-          - "1.70"
+          - "1.70"              # max version of msrv
         os:
           - ubuntu-latest
           - macos-latest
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -61,6 +61,21 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v4
+      - name: install protoc
+        uses: taiki-e/install-action@v2
+        with:
+          tool: protoc@${{ env.PROTOC_VERSION }}
+      - uses: taiki-e/install-action@cargo-hack
+      - uses: Swatinem/rust-cache@v2
+      - name: check with no-default-features
+        run: cargo hack --rust-version --no-private --no-dev-deps check --workspace --no-default-features
```
caspermeijn, line 75: What is the added value of `--no-dev-deps`? I read the docs, but I don't understand what it brings to a MSRV check.

caspermeijn (commented):
I prefer to keep the whole repo at the same MSRV. With the removal of `which` in #962 the MSRV of the published crates can be lowered to rust 1.65.  Would that be enough for your use case?

I suggest to only do a CI build with stable toolchain (so remove the 1.70 toolchain) and to add your proposed MSRV check.

.github/workflows/ci.yml
tottoto: `cargo-hack` removes `dev-dependencies` temporarily when being performed with `--no-dev-deps` option. As dependencies are affected by `dev-dependencies` and they might affect MSRV checking, this option is used to remove the effects.

tottoto:
Since checking MSRV for different crates can be performed easily as this pull request describes, and the crates except for `prost-build` only depend on very few crates, which have high stability, the value imposing unnecessarily restrictions seems small.

.github/workflows/ci.yml
caspermeijn: That makes sense, thanks

caspermeijn (changes requested):
I stick to my position. I want to keep the whole repo at the same MSRV. I believe the MSRV of the crates can be lowered to rust 1.65.

caspermeijn:
I opened a new PR with only the CI improvement: #1057

caspermeijn:
The CI improvement from this PR is merged. I won't accept multiple MSRV versions in one repo. If you want to lower the MSRV of the whole repo, you are welcome to reopen.

Issue #991, opened 2024-02-29
Rename prost-derive feature to derive


tottoto:
`derive` seems to be used widely. Adding `derive` feature and removing `prost-derive` feature after that can relax the breaking change.

caspermeijn:
Solved in #992

PR #992, opened 2024-03-01
feat: add derive feature

tottoto:
Related to #991.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,8 +42,9 @@ exclude = [
 bench = false
 
 [features]
-default = ["prost-derive", "std"]
-prost-derive = ["dep:prost-derive"]
+default = ["derive", "std"]
+derive = ["dep:prost-derive"]
+prost-derive = ["derive"]     # deprecated, please use derive feature instead
 no-recursion-limit = []
 std = []
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -86,10 +86,10 @@ where
 // Based on serde's equivalent re-export [1], but enabled by default.
 //
 // [1]: https://github.com/serde-rs/serde/blob/v1.0.89/serde/src/lib.rs#L245-L256
-#[cfg(feature = "prost-derive")]
+#[cfg(feature = "derive")]
 #[allow(unused_imports)]
 #[macro_use]
 extern crate prost_derive;
-#[cfg(feature = "prost-derive")]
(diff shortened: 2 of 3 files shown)


caspermeijn (commented):
It would be good to document the features we publicly expose. I don't think `rustdoc` will do that for you. 

Are you willing to document the features in the README of each public crate?

tottoto:
> It would be good to document the features we publicly expose.

I think so too. However I think that it is out of scope from this pull request as there is not a document about `features` in the README (there is about no_std crates, but it is not a comprehensive document about features). I think it should be a separated concern. I think the diff scope should be minimum as possible.

tottoto:
> It would be good to document the features we publicly expose.

I think so too. However I think that it is out of scope from this pull request as there is not a document about `features` in the README (there is about no_std crates, but it is not a comprehensive document about features). I think it should be a separated concern. I think the scope should be minimum as possible.

caspermeijn:
> > It would be good to document the features we publicly expose.
> 
> I think so too. However I think that it is out of scope from this pull request as there is not a document about `features` in the README (there is about no_std crates, but it is not a comprehensive document about features). I think it should be a separated concern. I think the scope should be minimum as possible.

I understand

caspermeijn (changes requested):
- `prost-types` uses the old derive feature. I suggest replacing it: https://github.com/tokio-rs/prost/blob/e3d708c0e031d71b87bb5ba5c0c3aad3393275fd/prost-types/Cargo.toml#L25
- `test-no-std` uses the old derive feature. I suggest replacing it: https://github.com/tokio-rs/prost/blob/e3d708c0e031d71b87bb5ba5c0c3aad3393275fd/tests-no-std/Cargo.toml#L27
- CI used the old derive feature, but I think this is a good one, as it confirms this is a non-breaking change: https://github.com/tokio-rs/prost/blob/e3d708c0e031d71b87bb5ba5c0c3aad3393275fd/.github/workflows/ci.yml#L61

tottoto:
> - prost-types uses the old derive feature. I suggest replacing it: 

Since it will be a breaking change, it is not recommended.

> - test-no-std uses the old derive feature. I suggest replacing it: 

Updated.

caspermeijn (approved):
LGTM

caspermeijn:

Issue #993, opened 2024-03-03
Implement Ord & PartialOrd for prost_types::Timestamp


Heidar-An:
it's quite useful to compare timestamps, this could be done by implementing PartialOrd + Ord onto Timestamp. I can work on this if needed.

PR #995, opened 2024-03-03
Feat: Implements ord and partialord to timestamp

LevBeta:
Related to #993

### DIFF
--- a/prost-types/src/lib.rs
+++ b/prost-types/src/lib.rs
@@ -395,6 +395,15 @@ impl std::hash::Hash for Timestamp {
     }
 }
 
+impl Ord for Timestamp {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        match self.seconds.cmp(&other.seconds) {
+            std::cmp::Ordering::Equal => self.nanos.cmp(&other.nanos),
+            ordering => ordering,
+        }
+    }
+}
+
 #[cfg(feature = "std")]
 impl From<std::time::SystemTime> for Timestamp {
     fn from(system_time: std::time::SystemTime) -> Timestamp {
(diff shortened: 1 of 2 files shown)


prost-types/src/lib.rs
```diff
@@ -907,4 +917,38 @@ mod tests {
         // Must contain at least one "/" character.
         assert_eq!(TypeUrl::new("google.protobuf.Duration"), None);
     }
+
+    #[test]
+    fn check_timestamp_ord() {
```
caspermeijn, line 922: Maybe you could add a parameterized test. This way you could easily test many scenarios . Here is an example: https://github.com/tokio-rs/prost/blob/a1ab898a9bfed407cd005e6655529a6577fb0371/prost-types/src/lib.rs#L733-L805

prost-types/src/protobuf.rs
```diff
@@ -2293,7 +2293,7 @@ impl NullValue {
 /// the time format spec '%Y-%m-%dT%H:%M:%S.%fZ'. Likewise, in Java, one can use
 /// the Joda Time's [`ISODateTimeFormat.dateTime()`](<http://www.joda.org/joda-time/apidocs/org/joda/time/format/ISODateTimeFormat.html#dateTime%2D%2D>) to obtain a formatter capable of generating timestamps in this format.
 #[allow(clippy::derive_partial_eq_without_eq)]
-#[derive(Clone, PartialEq, ::prost::Message)]
+#[derive(Clone, PartialEq, PartialOrd,::prost::Message)]
```
caspermeijn, line 2296: This file is generated, so your changes will be overwritten. Maybe you could add this to the generator so that all types can benefit from comparisons.

prost-types/src/lib.rs
```diff
@@ -395,6 +395,16 @@ impl std::hash::Hash for Timestamp {
     }
 }
 
+#[cfg(feature = "std")]
+impl Ord for Timestamp {
```
caspermeijn, line 399: I believe you could just `#[derive(PartialOrd, Ord)]` on `struct Timestamp`. At least your unit test succeeds.

LevBeta: I noticed a potential issue in both the comparison logic when dealing with timestamps. For instance, when comparing:
`Timestamp {
seconds: 1,
nanos: 0 }`

to

`Timestamp {
seconds: 0,
nanos: 1_000_000_000
}`

The current implementation considers the first timestamp as greater than the second one. However, it's crucial to recognize that the second timestamp actually represents a same duration.

caspermeijn: There is a normalize function in Timestamp. I believe that does what you are looking for. https://docs.rs/prost-types/latest/prost_types/struct.Timestamp.html#method.normalize

caspermeijn:
@LevBeta Are you still interested in this PR?

caspermeijn:
@LevBeta I am preparing a breaking release. Are you still interested in this PR? If so, please update the code

avsaase:
I'm interested in this change. If @LevBeta doesn't want to work on it anymore I can make a new PR.

caspermeijn:
> I'm interested in this change. If @LevBeta doesn't want to work on it anymore I can make a new PR.

Well there has been no response for 2 months. Feel free to open a new PR with a response to the feedback. As stated, I am preparing a breaking release and this would fit in with that.

PR #996, opened 2024-03-08
build(deps): update to env_logger 0.11

caspermeijn:
`env_logger` is only used as dev-dependencies or in private crates


caspermeijn:
`env_logger 0.11` requires MSRV 1.71

Issue #179, opened 2019-04-30
Unsupported raw identifiers in 0.5


fujiapple852:
I have a proto file (that I do not control) which contains a field named `self` (of type `Unit`).  In PROST 0.4 this is generated as `Self_` which works whereas in PROST 0.5 it is generated as `r#Self` which fails to compile in stable rust (1.34) with `r#Self is not currently supported` (see discussion on why [here](https://internals.rust-lang.org/t/raw-identifiers-dont-work-for-all-identifiers/9094/2)).

Workaround is either to use PROST 0.4 (which in turn forces using Rust 2015 and therefore the PROST generated code needs to be in a separate crate for Rust 2018 projects) or some clunky pre or post processing of the proto file or generated code.

Suggested fix: revert to appending an `_` to keyword identifiers for the list of 5 keywords not supported by raw identifiers.

Aside: I tried using the `field_attributes` option to prepend an `_` to the field but this doesn't work as it is rendered as a separate token to the identifier:

        #[prost(message, tag="1")]
        _
        r#Self(super::Unit),

fujiapple852:
I made a simple local fix to the [to_snake](https://github.com/danburkert/prost/blob/master/prost-build/src/ident.rs#L7) and [to_upper_camel](https://github.com/danburkert/prost/blob/master/prost-build/src/ident.rs#L31) functions to resolve the issue.  

I'd be happy to submit a PR however i'm unclear on what the desired logic should be here given the current implementation.  Specifically for the `to_upper_camel` function which seems to deal only with the `Self` case explicitly at the moment:

```
/// Converts a `snake_case` identifier to an `UpperCamel` case Rust type identifier.
pub fn to_upper_camel(s: &str) -> String {
    let mut ident = s.to_camel_case();

    // Use a raw identifier if the identifier matches a Rust keyword:
    // https://doc.rust-lang.org/reference/keywords.html.
    if ident == "Self" {
        ident.insert_str(0, "r#");
    }
    ident
}
```

danburkert:
Fixed in #180, thanks!

PR #998, opened 2024-03-10
Handle keyword `Self` after stripping enum type prefix

### LINKED ISSUE #997: Use of "Self" in stripped enum variants
**Version:** 0.12.3

**Problem:**

Generated enum variants are only sanitized for "Self", if prefix-stripping is not used, or does not apply.

Minimal reproducable protobuf defininition:

```protobuf
enum Feeding {
  FEEDING_UNSPECIFIED = 0;
  FEEDING_ASSISTED = 1;
  // Careful: code generation resulted in "Self".
  FEEDING_SELF = 2;
}
```

**Reason:**

The existing code does this:

1. convert `FEEDING_SELF` into `FeedingSelf`
2. sanitize if that _equals_ `Self` to replace with `Self_`
3. if prefix stripping is enabled:
   1. Extracts the type name, in this case `Feeding`
   2. convert `Feeding` to `Feeding`
   3. sanitize if that _equals_ `Self` to replace with `Self_`
   4. Remove the prefix `Feeding` from `FeedingSelf`, resulting in `Self`

The final result is `Self`, which is not sanitized after.

MixusMinimax:
Fixes #997

Relates to #179

### DIFF
--- a/prost-build/src/code_generator.rs
+++ b/prost-build/src/code_generator.rs
@@ -1191,21 +1191,27 @@ fn unescape_c_escape_string(s: &str) -> Vec<u8> {
 ///
 /// It also tries to handle cases where the stripped name would be
 /// invalid - for example, if it were to begin with a number.
+///
+/// The stripped name can be `Self`, which is sanitized analogously to [to_upper_camel]
 fn strip_enum_prefix(prefix: &str, name: &str) -> String {
     let stripped = name.strip_prefix(prefix).unwrap_or(name);
 
     // If the next character after the stripped prefix is not
     // uppercase, then it means that we didn't have a true prefix -
     // for example, "Foo" should not be stripped from "Foobar".
-    if stripped
+    match if stripped
         .chars()
         .next()
         .map(char::is_uppercase)
         .unwrap_or(false)
     {
-        stripped.to_owned()
+        stripped
     } else {
-        name.to_owned()
+        name
+    } {
+        "Self" => "Self_".to_owned(),
+        // ... other cases to be added here as needed
+        s => s.to_owned(),
     }
 }
 
@@ -1328,5 +1334,11 @@ mod tests {
         assert_eq!(strip_enum_prefix("Foo", "Foo"), "Foo");
         assert_eq!(strip_enum_prefix("Foo", "Bar"), "Bar");
         assert_eq!(strip_enum_prefix("Foo", "Foo1"), "Foo1");
+    }
+
+    #[test]
+    fn test_strip_enum_prefix_resulting_in_keyword() {
+        assert_eq!(strip_enum_prefix("Foo", "FooBar"), "Bar");
+        assert_eq!(strip_enum_prefix("Foo", "FooSelf"), "Self_");
     }
 }
--- a/tests/src/enum_keyword_variant.proto
+++ b/tests/src/enum_keyword_variant.proto
@@ -1,0 +1,21 @@
+syntax = "proto3";
+
+package enum_keyword_variant;
+
+enum Feeding {
+  FEEDING_UNSPECIFIED = 0;
+  FEEDING_ASSISTED = 1;
+  // Careful: code generation resulted in "Self".
+  FEEDING_SELF = 2;
+}
+
+enum Grooming {
+  UNSPECIFIED = 0;
+  ASSISTED = 1;
+  SELF = 2;
+}
+
+enum Number {
+  NUMBER_UNSPECIFIED = 0;
+  NUMBER_1 = 1;
+}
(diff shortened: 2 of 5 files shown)


prost-build/src/code_generator.rs
```diff
@@ -1191,21 +1191,27 @@ fn unescape_c_escape_string(s: &str) -> Vec<u8> {
...
+///
+/// The stripped name can be `Self`, which is sanitized analogously to [to_upper_camel]
 fn strip_enum_prefix(prefix: &str, name: &str) -> String {
     let stripped = name.strip_prefix(prefix).unwrap_or(name);
 
     // If the next character after the stripped prefix is not
     // uppercase, then it means that we didn't have a true prefix -
     // for example, "Foo" should not be stripped from "Foobar".
-    if stripped
+    match if stripped
```
caspermeijn, line 1202: I think this is not very readable. I suggest to first assigning the stripped value to a variable and then replace the keywords

prost-build/src/code_generator.rs
```diff
@@ -1191,21 +1191,27 @@ fn unescape_c_escape_string(s: &str) -> Vec<u8> {
 ///
 /// It also tries to handle cases where the stripped name would be
 /// invalid - for example, if it were to begin with a number.
+///
+/// The stripped name can be `Self`, which is sanitized analogously to [to_upper_camel]
 fn strip_enum_prefix(prefix: &str, name: &str) -> String {
```
caspermeijn, line 1196: I think it makes sense to move this function to `ident.rs`

MixusMinimax:
The thing is, we know that the result starts with an upper-case letter, as it is already converted to PascalCase, so Self really was the only issue. I can definitely do that though!

Thank you for your consideration!

prost-build/src/code_generator.rs
MixusMinimax: Will do

MixusMinimax:
I have moved the sanitizing of identifiers to a common function that can be used by all the others. What do you think? I think it'd definitely a lot safer this way

prost-build/src/ident.rs
```diff
@@ -21,23 +18,52 @@ pub fn to_snake(s: &str) -> String {
...
-    ident
+/// Strip an enum's type name from the prefix of an enum value.
+///
+/// This function assumes that both have been formatted to Rust's
+/// upper camel case naming conventions.
+///
+/// It also tries to handle cases where the stripped name would be
+/// invalid - for example, if it were to begin with a number.
+///
+/// The stripped name can be `Self`, which is sanitized analogously to [to_upper_camel]
```
caspermeijn, line 49: This last line doesn´t add information, as `to_upper_camel` docs doesn't explain how it is handled. Either remove or describe how it is sanitized.

prost-build/src/ident.rs
```diff
@@ -158,4 +245,19 @@ mod tests {
...
+        assert_eq!(strip_enum_prefix("Foo", "Foo"), "Foo");
+        assert_eq!(strip_enum_prefix("Foo", "Bar"), "Bar");
+        assert_eq!(strip_enum_prefix("Foo", "Foo1"), "Foo1");
+    }
+
+    #[test]
+    fn test_strip_enum_prefix_resulting_in_keyword() {
+        assert_eq!(strip_enum_prefix("Foo", "FooBar"), "Bar");
+        assert_eq!(strip_enum_prefix("Foo", "FooSelf"), "Self_");
+    }
```
caspermeijn, line 262: I would combine these two tests

prost-build/src/ident.rs
```diff
@@ -47,6 +73,67 @@ mod tests {
 
     use super::*;
 
+    #[test]
+    fn test_sanitize_identifier() {
```
caspermeijn, line 77: Thanks for this thorough test.

tests/src/enum_keyword_variant.proto
```diff
@@ -0,0 +1,21 @@
...
+  FEEDING_UNSPECIFIED = 0;
+  FEEDING_ASSISTED = 1;
+  // Careful: code generation resulted in "Self".
+  FEEDING_SELF = 2;
+}
+
+enum Grooming {
+  UNSPECIFIED = 0;
+  ASSISTED = 1;
+  SELF = 2;
```
caspermeijn, line 15: Should this have a keyword as well? for example: `ELSE = 3;`

caspermeijn (commented):
Splitting into `sanitize_identifier`  is great! This handles much more cases now. 

I have a few small comments left.

MixusMinimax:
Should be all done now

caspermeijn (approved):

caspermeijn:
Thanks for your contribution

PR #999, opened 2024-03-11
refactor: Optimize code based on cargo clippy suggestions

tgolang:


caspermeijn (commented):
After this change, the next tool (rustfmt) has some suggestions 😉.

Could you also fix the `dereferencing a tuple pattern where every element takes a reference` suggestions from clippy?

gibbz00:
Note that #1008 does this too.

caspermeijn:
Fixed by https://github.com/tokio-rs/prost/pull/1008

tgolang:
Thanks!

PR #1002, opened 2024-03-13
fix: include_file should handle proto without package

### LINKED ISSUE #1001: include_file does not work if any of the modules does not declare a package
a simple .proto file with no package will result in a Module object with an empty vector.

However, the following code is called 1035 in lib.rs of prost-build:

```rust
self.write_includes(
    modules.keys().collect(),
    &mut file,
    0,
    if target_is_env { None } else { Some(&target) },
)?;
```

That function starts at depth 0, meaning it tries to check modules whether `module.part(0)` is equal to something, which panics, as that vector is empty:

```rust
let modident = entries[0].part(depth);
let matching: Vec<&Module> = entries
    .iter()
    .filter(|&v| v.part(depth) == modident)
    .copied()
    .collect();
```

MixusMinimax:
Fixes #1001

The solution is to do the same thing that some other tests already do manually, we need to write the include for the root package at the root level of the include file.


caspermeijn:
I don't understand what the problem is you are trying to fix. 

Could you try to explain the problem you are encountering? Maybe you could add a minimal, reproducible example?

MixusMinimax:
Hi! The included tests have an example:

```proto
syntax = "proto3";

message NoPackageWithMessageExampleMsg {
  string some_field = 1;
}
```

```rs
mod proto {
    include!(concat!(env!("OUT_DIR"), "/no_package/_includes.rs"));
}

#[test]
fn it_works() {
    assert_eq!(
        proto::NoPackageWithMessageExampleMsg::default(),
        proto::NoPackageWithMessageExampleMsg::default()
    );
}
```

```rs
 {
    let out = std::env::var("OUT_DIR").unwrap();
    let out_path = PathBuf::from(out).join("no_package");

    std::fs::create_dir_all(&out_path).unwrap();

    prost_build::Config::new()
        .out_dir(out_path)
        .include_file("_includes.rs")
        .compile_protos(&[src.join("no_package_with_message.proto")], includes)
        .unwrap();
}
```

MixusMinimax:
The function that generates `_includes.rs` panics if the proto files contain a file without a `package` statement.

MixusMinimax:
This is the only change necessary: prost-build/src/lib.rs

the rest of the changes in this PR are just tests to validate that everything works.

The issue #1001 also has some more description about this

prost-build/src/lib.rs
```diff
@@ -1168,6 +1168,31 @@ impl Config {
         let mut written = 0;
         entries.sort();
 
+        if depth == 0 {
```
caspermeijn, line 1171: I think this is wrong place to solve this issue. I believe the issue stems from a mismatch in file names of in this function and the `compile_fds` function. In that function a mapping between module name and file name is created. `write_includes` should reuse that mapping.

https://github.com/tokio-rs/prost/blob/b3de938c21e9875f17e3a4c116203f9c2f8e8640/prost-build/src/lib.rs#L995-L1003

tests/src/build.rs
```diff
@@ -146,6 +146,36 @@ fn main() {
...
+
+    {
+        let out = std::env::var("OUT_DIR").unwrap();
+        let out_path = PathBuf::from(out).join("complex_package_structure");
+
+        std::fs::create_dir_all(&out_path).unwrap();
+
+        prost_build::Config::new()
+            .out_dir(out_path)
+            .include_file("__.rs")
```
caspermeijn, line 170: This should use unique names to prevent a conflict with the other generated include files.

tests/src/no_package_with_message.proto
```diff
@@ -0,0 +1,5 @@
+syntax = "proto3";
+
+message NoPackageWithMessageExampleMsg {
+  string some_field = 1;
+}
```
caspermeijn, line 5: I believe this test already exists: `no_root_packages`. Can you extend the test to also test your usecase?

caspermeijn (commented):
The PR title should be something like: "include_file should handle proto without package". Please change it.

tests/src/build.rs
MixusMinimax: This file gets created in a subfolder specifically for this test, it can not collide with other tests

tests/src/complex_package_structure/proto/post/content.proto
```diff
@@ -0,0 +1,13 @@
+syntax = "proto3";
+
+package post.content;
```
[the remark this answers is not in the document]
MixusMinimax: This example has multiple files, some with packages, and some without. It also contains nested packages, and imports between everything. I wanted to make sure that the generated include file is correct for a combination of these edge cases.
I can remove it or simplify it once I know the solution works

prost-build/src/lib.rs
MixusMinimax: well, compile_fds does not have to worry about the tree structure of modules.

write_includes does, as it recursively writes the module declarations with their corresponding `include`s.

The filenames of the modules were correct, the problem was that the logic itself would panic for the base case of the recursion.

For reference, this is the generated include file of my test:

```rust
// This file is @generated by prost-build.
include!("__.default.rs");
pub mod image {
    include!("image.rs");
}
pub mod post {
    pub mod content {
        include!("post.content.rs");
    }
    include!("post.rs");
}
pub mod user {
    include!("user.rs");
}
```

The `write_includes` function needs to write handle modules without a path differently.

I think I will just change the function itself rather that hack some fix into it

caspermeijn: Good job!

MixusMinimax:
it seems that `Module::to_partial_filename` is no longer used. As it is a private function, should I remove it? It was used in the old implementation of `write_includes`.

prost-build/src/lib.rs
```diff
@@ -1407,6 +1398,8 @@ impl Module {
         self.components.is_empty()
     }
 
+    // TODO: remove if it stays unused
+    #[allow(dead_code)]
```
caspermeijn, line 1402: > it seems that `Module::to_partial_filename` is no longer used. As it is a private function, should I remove it? It was used in the old implementation of `write_includes`.

Yes, unused code should be removed

caspermeijn (changes requested):
This PR is almost ready. My only major open question is this:
- What do `no_package_with_message` and `complex_package_structure` test, that is not already tested by the existing tests?

MixusMinimax:
I have been quite busy, sorry, I'll get around to finishing that up over the weekend. Those other tests were more like integration tests rather than unit tests, just testing all the features that my changes could have affected, mostly a combination of using the write_include_file feature and proto package structures that contain nested packages, and content in the root package.
None of the other tests tested all of that in combination, which I needed because the features could affect each other.
But I agree, just altering the existing tests makes more sense.

MixusMinimax:
The `no_root_packages` test does not generate an includes file, and I did not want to mess with existing tests, but I think in this case it makes sense to just add that to `no_root_packages` so that `no_package_with_message` is not needed anymore.

MixusMinimax:
Ok this should be it

prost-build/src/module.rs
```diff
@@ -40,6 +39,15 @@ impl Module {
         self.components.iter().map(|s| s.as_str())
     }
 
+    #[must_use]
+    #[inline(always)]
+    pub fn starts_with(&self, needle: &[String]) -> bool
```
caspermeijn, line 44: I believe this function should be `pub(crate)`. Similar to the function you removed.

caspermeijn (changes requested):

prost-build/src/module.rs
MixusMinimax: Done 👍🏽

caspermeijn (approved):

caspermeijn:
Thank you for your contribution. I appreciate the patience during the review.

PR #1003, opened 2024-03-15
docs: Document feature flags

caspermeijn:
Open question:
- What is the function of `prost-derive` feature?


prost-build/src/lib.rs
```diff
@@ -125,6 +125,10 @@
 //! // Now compile your proto files via prost-build
 //! ```
 //!
+//! ## Feature Flags
```
gibbz00, line 128: Details about some of these features are currently placed before this. Perhaps place this first and reference the respective details sections?

caspermeijn: Good idea, I moved the Feature Flags section

LucioFranco (approved):

PR #1011, opened 2024-03-27
prost-build: refactor code generator (ground work for builders)

mzabaluev:
Refactoring of existing code taken out of #901:

* Collect message field data into vectors of structs that can be reused in multiple passes. This will be useful for the upcoming generation of builders.
* Add `prost_path` helper function to DRY repetitive expressions constructing the path to the prost crate for generated code.
* Add `boxed` helper method to capture the logic of deciding whether a field is boxed.

### DIFF
--- a/prost-build/src/code_generator.rs
+++ b/prost-build/src/code_generator.rs
@@ -952,8 +952,6 @@ impl<'a> CodeGenerator<'a> {
     }
 
     fn resolve_type(&self, field: &FieldDescriptorProto, fq_message_name: &str) -> String {
-        let prost_path = prost_path(self.config);
-
         match field.r#type() {
             Type::Float => String::from("f32"),
             Type::Double => String::from("f64"),
@@ -962,7 +960,7 @@ impl<'a> CodeGenerator<'a> {
             Type::Int32 | Type::Sfixed32 | Type::Sint32 | Type::Enum => String::from("i32"),
             Type::Int64 | Type::Sfixed64 | Type::Sint64 => String::from("i64"),
             Type::Bool => String::from("bool"),
-            Type::String => format!("{}::alloc::string::String", prost_path),
+            Type::String => format!("{}::alloc::string::String", prost_path(self.config)),
             Type::Bytes => self
                 .config
                 .bytes_type
--- a/prost-build/src/code_generator.rs
+++ b/prost-build/src/code_generator.rs
@@ -44,6 +44,49 @@ fn push_indent(buf: &mut String, depth: u8) {
         buf.push_str("    ");
     }
 }
+
+fn prost_path(config: &Config) -> &str {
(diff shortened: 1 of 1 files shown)


prost-build/src/code_generator.rs
```diff
@@ -44,6 +44,49 @@ fn push_indent(buf: &mut String, depth: u8) {
...
+            name: to_snake(proto.name()),
+            proto,
+            path_index,
+        }
+    }
+}
+
+struct OneofField {
+    name: String,
+    proto: OneofDescriptorProto,
```
caspermeijn, line 70: I believe this should be called `descriptor`

prost-build/src/code_generator.rs
```diff
@@ -159,21 +202,33 @@ impl<'a> CodeGenerator<'a> {
...
+                if proto.proto3_optional.unwrap_or(false) {
+                    Either::Left(Field::new(proto, idx))
+                } else if let Some(oneof_index) = proto.oneof_index {
+                    Either::Right((oneof_index, (proto, idx)))
                 } else {
-                    Either::Left((field, idx))
+                    Either::Left(Field::new(proto, idx))
                 }
             });
+        // Optional fields create a synthetic oneof that we want to skip
```
caspermeijn, line 220: This seems new functionality. Why is this needed? Does this change the generated code?

caspermeijn (commented):
Thank you for splitting this of the builder PR.

Your commit description mentions three different improvements. Please split the separate improvements into separate commits.

prost-build/src/code_generator.rs
mzabaluev: This does not change the generated code.

#901 adds a third pass over the oneof declarations to generate the setters for the builder, so I opted to collect the field data here in a pre-filtered and more usable form.

prost-build/src/code_generator.rs
```diff
@@ -246,14 +293,8 @@ impl<'a> CodeGenerator<'a> {
...
             self.path.pop();
 
-            for (idx, oneof) in message.oneof_decl.into_iter().enumerate() {
-                let idx = idx as i32;
-                // optional fields create a synthetic oneof that we want to skip
-                let fields = match oneof_fields.remove(&idx) {
-                    Some(fields) => fields,
-                    None => continue,
-                };
-                self.append_oneof(&fq_message_name, oneof, idx, fields);
```
mzabaluev, line 256: This was done with a moving iterator and moving the field data into `append_oneof`, but there was no reason to move the data because code generation only needs references. And I need the same data for the builder pass which will be added later.

caspermeijn: I see, the code just moved. That de-duplucation is nice.

mzabaluev:
Superseded by #1016.

PR #1014, opened 2024-03-29
build: Use git submodule to download protobuf sources

caspermeijn:
Instead of downloading a tarball, use a submodule to get the protobuf sources. Newer versions of protobuf require recursive submodules for its dependencies

In preparation of updating the protobuf version used for conformance tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,6 +44,8 @@ jobs:
           - windows-latest
     steps:
       - uses: actions/checkout@v4
+        with:
+          submodules: 'recursive'
       - name: install toolchain (${{ matrix.toolchain }})
         uses: dtolnay/rust-toolchain@master
         with:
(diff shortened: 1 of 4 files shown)


caspermeijn:
> @caspermeijn does this mean that when cloning the project you must pass the recursive flag to git? Or else tests will not pass with regards to the conformance test suite?

That is right. The submodule needs to be checked out. If not, the build will fail and suggest running `git submodule update --init --recursive` to fix the situation.

LucioFranco:
So essentially because protobuf has submodules now the tarball doesn't work and thus we must also follow suit?

caspermeijn:
> So essentially because protobuf has submodules now the tarball doesn't work and thus we must also follow suit?

Yes, that is how I understand the situation.

Alternatively, we could require all of protobuf build dependencies to be present, then the tarball works great. But I would rather not maintain those install instructions.

Or we require the build conformance tests to be installed somehow, but that degrades the current developer experience.

LucioFranco (approved):

PR #1008, opened 2024-03-17
Fix clippy warnings and enable clippy in CI

gibbz00:
Depends on #1007.

PR #1015, opened 2024-03-29
prost-build: Split `Config` into a separate module and file.

gibbz00:
See note about possibly changing the visibility of some helper methods to private ~~. Leaving this as a draft until it is resolved~~ 😊 

https://github.com/tokio-rs/prost/blob/eff01566f2e6e759c1fb951b7c413ae16adcb4c8/prost-build/src/lib.rs#L150-L153

Depends on #1008.


caspermeijn:
I already commented on the config split in #1020. I suggest doing the config and module split as two commits in a single PR.

gibbz00:
Oki, I think I'll just close both and redo this in a new PR. Think it will be easier to resolve new conflicts that way.

PR #1019, opened 2024-03-29
prost-build: `CodeGenerator::boxed` method

mzabaluev:
A helper method to capture the logic of deciding whether a field needs to be boxed. This follows the pattern with other methods like `optional`, and will allow reusing the logic in the upcoming builder codegen in #901.

### DIFF
--- a/prost-build/src/code_generator.rs
+++ b/prost-build/src/code_generator.rs
@@ -402,18 +402,8 @@ impl<'a> CodeGenerator<'a> {
         let repeated = field.descriptor.label == Some(Label::Repeated as i32);
         let deprecated = self.deprecated(&field.descriptor);
         let optional = self.optional(&field.descriptor);
+        let boxed = self.boxed(&field.descriptor, fq_message_name);
         let ty = self.resolve_type(&field.descriptor, fq_message_name);
-
-        let boxed = !repeated
-            && ((type_ == Type::Message || type_ == Type::Group)
-                && self
-                    .message_graph
-                    .is_nested(field.descriptor.type_name(), fq_message_name))
-            || (self
-                .config
-                .boxed
-                .get_first_field(fq_message_name, field.descriptor.name())
-                .is_some());
 
         debug!(
             "    field: {:?}, type: {:?}, boxed: {}",
@@ -1066,6 +1056,20 @@ impl<'a> CodeGenerator<'a> {
             Type::Message => true,
             _ => self.syntax == Syntax::Proto2,
         }
+    }
+
+    fn boxed(&self, field: &FieldDescriptorProto, fq_message_name: &str) -> bool {
+        let fd_type = field.r#type();
+        field.label != Some(Label::Repeated as i32)
+            && ((fd_type == Type::Message || fd_type == Type::Group)
+                && self
+                    .message_graph
+                    .is_nested(field.type_name(), fq_message_name))
+            || (self
+                .config
+                .boxed
+                .get_first_field(fq_message_name, field.name())
+                .is_some())
     }
 
     /// Returns `true` if the field options includes the `deprecated` option.
--- a/prost-build/src/code_generator.rs
+++ b/prost-build/src/code_generator.rs
@@ -1052,6 +1052,11 @@ impl<'a> CodeGenerator<'a> {
         }
     }
 
+    /// Returns whether the Rust type for this field needs to be `Box<_>`.
+    ///
+    /// This can be explicitly configured with `Config::boxed`, or necessary
+    /// to prevent an infinitely sized type definition in case when the type of
+    /// a non-repeated message field transitively contains the message itself.
     fn boxed(
         &self,
         field: &FieldDescriptorProto,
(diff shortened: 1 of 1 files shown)


prost-build/src/code_generator.rs
```diff
@@ -1036,6 +1026,20 @@ impl<'a> CodeGenerator<'a> {
         }
     }
 
+    fn boxed(&self, field: &FieldDescriptorProto, fq_message_name: &str) -> bool {
```
caspermeijn, line 1029: Please add a comment describing the function, for example: 
```
    /// Returns whether this field sould be `Box<_>`.
```

caspermeijn (commented):
At line 626, there is another instance of a `boxed` variable. Can that be replaced by the same function?

prost-build/src/code_generator.rs
```diff
@@ -1036,6 +1016,38 @@ impl<'a> CodeGenerator<'a> {
...
+    /// a non-repeated message field transitively contains the message itself.
+    fn boxed(
+        &self,
+        field: &FieldDescriptorProto,
+        fq_message_name: &str,
+        oneof: Option<&str>,
+    ) -> bool {
+        if field.label == Some(Label::Repeated as i32) {
+            return false;
+        }
```
mzabaluev, line 1032: Note that this changes the logic from the original: now any `boxed` match from the config will be ignored if the field is repeated, while previously it gave true. But I think that was wrong, because it never makes sense to box a `Vec`.

mzabaluev: To think of it, this should be reverted if we want this to merged for 0.12.x releases.

mzabaluev:
> At line 626, there is another instance of a `boxed` variable. Can that be replaced by the same function?

Done, with some complication and a questionable behavioral change that I have commented on above.

prost-build/src/code_generator.rs
```diff
@@ -1036,6 +1016,38 @@ impl<'a> CodeGenerator<'a> {
...
+            && self
+                .message_graph
+                .is_nested(field.type_name(), fq_message_name)
+        {
+            return true;
+        }
+        let config_path = match oneof {
+            None => Cow::Borrowed(fq_message_name),
+            Some(ooname) => Cow::Owned(format!("{fq_message_name}.{ooname}")),
+        };
```
caspermeijn, line 1044: I have a few points that are related:
- I don't like the concat of these names here. I understand the original code uses fq_message_name (few lines up) and the oneof_name (which is now this concat).
- The primary function of MessageGraph is to be used in this boxed check. However, MessageGraph is not aware of the boxed definition. Maybe it makes sense to move this boxed definition inside MessageGraph?
- MessageGraph can mark messages as recursive, but that doesn't consider manually boxed fields.
- Why are oneof fields not added to MessageGraph? If that was true, is_nested can also work with oneof_name and this function only needs a single name.

I think this boxed and MessageGraph code could use a refactor. I don't have concrete advice to give. Are you willing to dive into this?

prost-build/src/code_generator.rs
caspermeijn: The quick fix is to split this function into a `boxed_field` and `boxed_oneof` variants. Make sure to document why two variants are needed.

Maybe you could add a deprecation warning like this:
```rust
println!("cargo:warning=Field X is repeated and manually marked as boxed. This is deprecated and support will be removed in a later release");
```

caspermeijn (commented):
Do you think there are enough tests for boxed fields and oneof fields?

prost-build/src/code_generator.rs
mzabaluev: I could improve MessageGraph in a later PR, this one only moves code around without changing the functionality (apart from the logic change on repeated fields which I will revert).

caspermeijn: I agree that the `MessageGraph` should be changed in another PR.

prost-build/src/code_generator.rs
mzabaluev: > The quick fix is to split this function into a `boxed_field` and `boxed_oneof` variants. Make sure to document why two variants are needed.

I don't like this idea. The logic would need to be replicated, even though it's mostly the same for regular fields and oneof members. More so with the deprecation warning, which I have now added as per your suggestion.

prost-build/src/code_generator.rs
```diff
@@ -1068,6 +1052,47 @@ impl<'a> CodeGenerator<'a> {
...
+            .config
+            .boxed
+            .get_first_field(&config_path, field.name())
+            .is_some()
+        {
+            println!(
+                "cargo:warning=\
+                Field X is repeated and manually marked as boxed. \
+                This is deprecated and support will be removed in a later release"
+            );
```
caspermeijn, line 1090: This is incorrect. The warning will show for any field marked as boxed.

```suggestion
            if repeated {
                println!(
                    "cargo:warning=\
                    Field X is repeated and manually marked as boxed. \
                    This is deprecated and support will be removed in a later release"
                );
            }
```

caspermeijn (changes requested):
I like the current version. I found one logic error, then this is ready.

caspermeijn (approved):

caspermeijn:
Thank you for your contribution. I appreciate your patience during the review.

PR #1020, opened 2024-03-29
prost-build: Split `Config` and `Module` into a separate module and files

gibbz00:
PR contains a visibility question similar to #1015, which this PR builds upon,


prost-build/src/lib.rs
```diff
@@ -131,37 +131,34 @@
...
+use crate::ident::to_snake;
+
 mod message_graph;
 mod path;
 
+mod config;
+// HELP(gibbz00): I think `error_message_protoc_not_found`, `protoc_from_env`, and `protoc_include_from_env`
+// should be made private. Only used inside config and not in any other part
+// of prost. Leaving this as is for now as this would technically be breaking
+// changes.
```
caspermeijn, line 152: I agree these should be private. I am willing to accept this breaking change. If you want to prevent breaking changes, then this should change to a `TODO(breaking-change):` comment.

prost-build/src/lib.rs
```diff
@@ -159,6 +154,10 @@ pub use config::{
     error_message_protoc_not_found, protoc_from_env, protoc_include_from_env, Config,
 };
 
+mod module;
+// HELP(gibbz00): I think `Module` should be made pub(crate).
+pub use module::Module;
```
caspermeijn, line 159: I am not sure. It seems like it is used in this generate call: https://docs.rs/prost-build/latest/prost_build/struct.Config.html#method.generate

I don't understand the use case, but I think we should leave that `generate` method in place.

caspermeijn (commented):
I am amazed at how many lines the Config struct is. This move is a big improvement.

Needs a rebase to remove the clippy commits from this branch.

Can you check none of the moved code is changed meanwhile on the main branch?

prost-build/src/lib.rs
gibbz00: Alright, you're probably right :)

prost-build/src/lib.rs
gibbz00: Sweet. I'll keep that in mind next time :)

gibbz00:
Hi!

I've now resolved the comments and combined the commit form #1015. I was first thinking about closing both in order to make sure that I wasn't overriding any new changes, but none had come in.

gibbz00:
I think that the failing CI clippy job was caused by https://github.com/tokio-rs/prost/pull/1014. It should probably be amended or fixed in a separate PR though. Might be worth denying failing CI jobs from getting merged though.

caspermeijn (approved):
Looks good. I will fix CI in #1032 and then merge this

Issue #1021, opened 2024-03-30
Consider placing `Config::format` behind a feature flag.


gibbz00:
https://github.com/tokio-rs/prost/blob/50bab4ff9f5bae1cde6b54d6de02107d30f11453/prost-build/src/lib.rs#L940-L944

Think it is better to create a compile time error when someone is trying to enable/disable formatting when it isn't enabled.

PR #1022, opened 2024-03-30
Place `Config::format` behind the `format` feature flag.

gibbz00:
Fixes: #1021 

~~Depends upon: #1020.~~

### DIFF
--- a/prost-build/Cargo.toml
+++ b/prost-build/Cargo.toml
@@ -33,6 +33,7 @@ tempfile = "3"
 once_cell = "1.17.1"
 regex = { version = "1.8.1", default-features = false, features = ["std", "unicode-bool"] }
 
+# feature: format
 prettyplease = { version = "0.2", optional = true }
 syn = { version = "2", features = ["full"], optional = true }
 
--- a/prost-build/src/config.rs
+++ b/prost-build/src/config.rs
@@ -49,6 +49,7 @@ pub struct Config {
     pub(crate) skip_protoc_run: bool,
     pub(crate) include_file: Option<PathBuf>,
     pub(crate) prost_path: Option<String>,
+    #[cfg(feature = "format")]
     pub(crate) fmt: bool,
 }
 
@@ -735,10 +736,12 @@ impl Config {
         self
     }
 
+    // IMPROVEMENT: https://github.com/tokio-rs/prost/pull/1022/files#r1563818651
     /// Configures the code generator to format the output code via `prettyplease`.
     ///
     /// By default, this is enabled but if the `format` feature is not enabled this does
     /// nothing.
+    #[cfg(feature = "format")]
     pub fn format(&mut self, enabled: bool) -> &mut Self {
         self.fmt = enabled;
         self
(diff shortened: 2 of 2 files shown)


caspermeijn:
Please make this PR independent of #1020. 

My reasoning:
- The current diff is unreadable
- The other PR is in draft.

gibbz00:
Valid points 😊

Though of doing this at first, I wanted to avoid having to fix the conflicts that would arise. Do you think that this will be merged before #1020? I wouldn't mind making this an independent PR is that's the case.

caspermeijn:
Well, from my point of view:
- #1020 is in draft and this PR is not, so #1020 can't be merged first
- This PR and #1020 are completely unrelated, other than the same author and potential merge conflict.
- I can't execute a good review of this PR, so it either needs to be fixed or should be opened after the other merge.

The last commit of this branch seems like an improvement, so I think it makes sense to keep the PR open. I suggest making it independent and solve the merge conflict at a later time.

gibbz00:
Alrighty, a rebase later should be too hard anyway. I've now made this PR independent of the rest.

prost-build/src/lib.rs
```diff
@@ -942,9 +943,7 @@ impl Config {
     }
 
     /// Configures the code generator to format the output code via `prettyplease`.
-    ///
-    /// By default, this is enabled but if the `format` feature is not enabled this does
-    /// nothing.
+    #[cfg(feature = "format")]
```
caspermeijn, line 946: I would like it to be clear that this function is only available with the `format` feature. It appears that cargo doc will only show an information balloon for modules with a feature config.

```suggestion
    /// 
    /// Enabled by default.
    ///
    /// This configuration option is only available with the `format` feature enabled.
    #[cfg(feature = "format")]
```

prost-build/src/lib.rs
```diff
@@ -1262,8 +1261,13 @@ impl Config {
             }
         }
 
+        #[cfg(feature = "format")]
```
caspermeijn, line 1264: This is a cool trick.

gibbz00: 🤙

prost-build/src/lib.rs
gibbz00: What do you mean about information balloon?

Here's an example of it being shown for functions too: https://docs.rs/tokio/latest/tokio/net/struct.TcpStream.html#method.try_read_buf

I think that would be the more "correct way". This is unfortunately not enabled by default in cargo doc [yet](https://github.com/rust-lang/rust/issues/43781). One needs to add `#![cfg_attr(docsrs, feature(doc_auto_cfg))]` to each `lib.rs` and the bellow snippet to the following `Cargo.toml`s (maybe possible in workspace).

```toml
[package.metadata.docs.rs]
all-features = true
rustdoc-args = ["--cfg", "docsrs"]
```

I would personally be in favor of doing the latter, and open a separate PR for this as well 🤗

caspermeijn: Yes, I mean that information balloon. Good that it can be shown for functions as well. Too bad it is not the default. A command is fine for now.

caspermeijn (changes requested):
Please rebase to fix the merge conflict

gibbz00:
Rebased and fixed suggestions!

caspermeijn (approved):

caspermeijn:

PR #1026, opened 2024-04-13
WIP:  Transition to `quote` + `syn` for code generation.

gibbz00:
Throwing this out here ASAP as a draft so that other aware of it. Because it will clash with PRs such as #https://github.com/tokio-rs/prost/pull/1019.

Will create a proper write-up once things become presentable.

PR #1029, opened 2024-04-14
chore - codegen: `Syntax` to a separate file

gibbz00:
Part of a series of module splits which became clear from #1026. Also functions as a way to improve the final reviewability of #1026.

### DIFF
--- a/prost-build/src/code_generator.rs
+++ b/prost-build/src/code_generator.rs
@@ -23,11 +23,8 @@ use crate::Config;
 mod c_escaping;
 use c_escaping::unescape_c_escape_string;
 
-#[derive(PartialEq)]
-enum Syntax {
-    Proto2,
-    Proto3,
-}
+mod syntax;
+use syntax::Syntax;
 
 pub struct CodeGenerator<'a> {
     config: &'a mut Config,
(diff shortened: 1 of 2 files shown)


prost-build/src/code_generator.rs
```diff
@@ -66,18 +63,12 @@ impl<'a> CodeGenerator<'a> {
...
-            Some(s) => panic!("unknown syntax: {}", s),
-        };
-
         let mut code_gen = CodeGenerator {
             config,
             package: file.package.unwrap_or_default(),
             type_path: Vec::new(),
             source_info,
-            syntax,
+            syntax: file.syntax.as_ref().map(String::as_str).into(),
```
caspermeijn, line 71: When you change to `impl From<Option<&String>> for Syntax`, this could be simplified. Am I right?
```suggestion
            syntax: file.syntax.into(),
```

caspermeijn (commented):
Why is a separated module better for such a small type?

prost-build/src/code_generator.rs
gibbz00: I suppose it would. `&String` in method arguments is generally a bad practice, though.

gibbz00:
> Why is a separated module better for such a small type?

I could inline the module in `code_generator/mod.rs` if you prefer. I like to avoid mixing domain models in the same module. Makes it easier to reason about visibility, dependencies and usage.

prost-build/src/code_generator.rs
caspermeijn: Well, `&String` is wrong anyway. This seems to work:
```suggestion
            syntax: file.syntax.as_deref().into(),
```

caspermeijn (changes requested):
Can you solve the merge conflict?

prost-build/src/code_generator.rs
gibbz00: Thanks!

gibbz00:
Rebased and resolved comments!

caspermeijn (approved):

caspermeijn:

PR #1035, opened 2024-04-21
Use workspace dependencies.

gibbz00:
Makes dependency version handling easier. Relates in some way to https://github.com/tokio-rs/prost/pull/957

~~Depends on #1033~~ (merged into master and rebased here)


caspermeijn:
What problem does this solve? Dependabot will change all uses of a dependency when there is an update, right?

caspermeijn:
I am not interested in workspace dependencies. I feel like Dependabot already makes dependency management easy. I am open to having my mind changed.

PR #1038, opened 2024-04-22
chore: Update crate descriptions

caspermeijn:
Each published crate should have a unique description.


gibbz00:
I'm assuming you took inspiration for the descriptions from the respective READMEs. Think that's a good idea.

LGTM!

PR #1039, opened 2024-04-22
chore: Update internal crates to Rust edition 2021

caspermeijn:
The published crates were already using edition 2021. Update all internal crates as well.

Also add a new tests-2018 crate that runs all tests using edition 2018.

### DIFF
--- a/tests-2018/Cargo.toml
+++ b/tests-2018/Cargo.toml
@@ -1,14 +1,18 @@
 [package]
-name = "tests"
+name = "tests-2018"
 version = "0.0.0"
 authors = [
     "Dan Burkert <dan@danburkert.com>",
     "Tokio Contributors <team@tokio.rs>",
 ]
 publish = false
-edition = "2021"
+edition = "2018"
 
-build = "src/build.rs"
+build = "../tests/src/build.rs"
+
+[lib]
+doctest = false
+path = "../tests/src/lib.rs"
 
 [features]
 default = ["std"]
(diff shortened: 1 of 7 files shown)


gibbz00:
Hii, two things 😊

1) Why are we testing against old editions?

https://doc.rust-lang.org/edition-guide/editions/#editions-do-not-split-the-ecosystem
> The most important rule for editions is that crates in one edition can interoperate seamlessly with crates compiled in other editions. This ensures that the decision to migrate to a newer edition is a "private one" that the crate can make without affecting others.

2) Maybe the test-2018 addition should be made into a separate PR so that we can discuss 1) separately without blocking this one for #1036

tests-2018/Cargo.toml
gibbz00, line 1: Perhaps this change should be made after https://github.com/tokio-rs/prost/pull/1036 and https://github.com/tokio-rs/prost/pull/1035.

caspermeijn:
> 1. Why are we testing against old editions?

I am not sure, but I assume this is because we generate code that must be compatible with older and newer editions. For example, in some test code is a `cfg_if` that adds `extern crate` statements for 2015 edition only.

Well, I see two options:
1. Add the `tests-2018` crate. I mostly did this because the `tests-2015` crate exists. I am not sure what the added value is over just running the tests with the oldest edition.
2. Update main `tests` crate to edition 2021 and don't explicitly test 2018. I am fine with that, as I think the difference between 2018 and 2021 is small enough to not cause any problems.

What do you think is best?

LucioFranco (approved):
I think it makes sense to test old editions if its easy to do and I think option 1 here makes a lot of sense to me

gibbz00:
>  we generate code that must be compatible with older and newer editions.

Ahh, I see. Makes perfect sense.

Also agree that it's worth testing against the 2018 edition given that it's quite easy.

PR #1043, opened 2024-04-26
tests: Remove unused dependencies

caspermeijn:
Found these unused dependencies using: `cargo +nightly udeps && cargo +nightly udeps --all-targets`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,6 +59,26 @@ jobs:
       - name: Check unused dependencies
         run: cargo machete
 
+  unused_dependencies:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: 'recursive'
+      - uses: dtolnay/rust-toolchain@nightly
+      - name: install protoc
+        uses: taiki-e/install-action@v2
+        with:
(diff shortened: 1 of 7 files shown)


gibbz00:
I would suggest adding this step to CI. E.g.

```yaml
  unused_deps:
    name: Check for unused dependencies
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      ...
      - name: Run cargo-udeps
        uses: aig787/cargo-udeps-action@v1
        with:
          args: --all-features
```

Unsure if this has to be added in between: 

```yaml
with:
    submodules: 'recursive'
- name: install protoc
  uses: taiki-e/install-action@v2
  with:
    tool: protoc@${{ env.PROTOC_VERSION }}
- name: install ninja
  uses: ./.github/actions/setup-ninja
```

Also unsure if `--benches`, `--tests`, `--examples` and `--bins` are enabled by default.

caspermeijn:
> I would suggest adding this step to CI. E.g.

Good idea. 

> Unsure if this has to be added in between:

Yes, cargo-udeps does a full build and looks at the linked binary. So all dependencies are required.

> Also unsure if `--benches`, `--tests`, `--examples` and `--bins` are enabled by default.

`--all-targets` will do that for you: https://doc.rust-lang.org/cargo/commands/cargo-build.html#target-selection

PR #1044, opened 2024-04-26
tests: Create a separate tempdir for each test

caspermeijn:
Many tests from `prost-build` wrote to the same temporary directory. Therefore these tests overwrite each other files. Because `cargo test` runs multi threaded some tests failed sometimes.


gibbz00:
Awesome 😊

PR #1045, opened 2024-04-26
Change generated functions signatures to remove type parameters

### LINKED ISSUE #1042: `impl ::prost::Message` for a struct shadows the user-defined type
Minimal protobuf file to reproduce:

```proto
syntax = "proto3";
enum B {
  DEFAULT = 0;
  NON_DEFAULT = 1;
}
message A { B f1 = 1; }
```

Problem is that generated `impl ::prost::Message for A` contains a generic type parameter for a number of functions, that shadows the `B` enum:
```rust
            #[allow(unused_variables)]
            fn encode_raw<B>(&self, buf: &mut B)
            where
                B: ::prost::bytes::BufMut,
            {
                if self.f1 != B::default() as i32 {
                    ::prost::encoding::int32::encode(1u32, &self.f1, buf);
                }
            }
```

The error message from the compiler, that is not clear:
```
error[E0599]: no function or associated item named `default` found for type parameter `B` in the current scope
 --> tests/../.proto_out/test.rs:3:28
  |
3 | #[derive(Clone, PartialEq, ::prost::Message)]
  |                            ^^^^^^^^^^^^^^^^
  |                            |
  |                            function or associated item not found in `B`
  |                            function or associated item `default` not found for this type parameter
  |
  = help: items from traits can only be used if the type parameter is bounded by the trait
  = note: this error originates in the derive macro `::prost::Message` (in Nightly builds, run with -Z macro-backtrace for more info)
help: the following trait defines an item `default`, perhaps you need to restrict type parameter `B` with it:
  |
3 | #[derive(Clone, PartialEq, ::prost::Message: Default)]
  |                                            +++++++++

For more information about this error, try `rustc --explain E0599`.
```

There are at least two possible solutions:
1. Rename the generic type parameter to reduce the chance of ident clashes (e.g. `<ImplProstBytesBufMut>` instead of `<B>`);
2. Add a note or a known issue regarding this problem.

Probably, there are some other similar issues that are not noticed yet.

shizzard:
Fixes #1042

### DIFF
--- a/prost-derive/src/lib.rs
+++ b/prost-derive/src/lib.rs
@@ -174,19 +174,19 @@ fn try_message(input: TokenStream) -> Result<TokenStream, Error> {
     let expanded = quote! {
         impl #impl_generics ::prost::Message for #ident #ty_generics #where_clause {
             #[allow(unused_variables)]
-            fn encode_raw<B>(&self, buf: &mut B) where B: ::prost::bytes::BufMut {
+            fn encode_raw(&self, buf: &mut impl ::prost::bytes::BufMut) {
                 #(#encode)*
             }
 
             #[allow(unused_variables)]
-            fn merge_field<B>(
+            fn merge_field(
                 &mut self,
                 tag: u32,
                 wire_type: ::prost::encoding::WireType,
-                buf: &mut B,
+                buf: &mut impl ::prost::bytes::Buf,
                 ctx: ::prost::encoding::DecodeContext,
             ) -> ::core::result::Result<(), ::prost::DecodeError>
-            where B: ::prost::bytes::Buf {
+            {
                 #struct_name
                 match tag {
                     #(#merge)*
@@ -463,21 +463,21 @@ fn try_oneof(input: TokenStream) -> Result<TokenStream, Error> {
     let expanded = quote! {
         impl #impl_generics #ident #ty_generics #where_clause {
             /// Encodes the message to a buffer.
-            pub fn encode<B>(&self, buf: &mut B) where B: ::prost::bytes::BufMut {
+            pub fn encode(&self, buf: &mut impl ::prost::bytes::BufMut) {
                 match *self {
                     #(#encode,)*
                 }
             }
 
             /// Decodes an instance of the message from a buffer, and merges it into self.
-            pub fn merge<B>(
+            pub fn merge(
                 field: &mut ::core::option::Option<#ident #ty_generics>,
                 tag: u32,
                 wire_type: ::prost::encoding::WireType,
-                buf: &mut B,
+                buf: &mut impl ::prost::bytes::Buf,
                 ctx: ::prost::encoding::DecodeContext,
             ) -> ::core::result::Result<(), ::prost::DecodeError>
-            where B: ::prost::bytes::Buf {
+            {
                 match tag {
                     #(#merge,)*
                     _ => unreachable!(concat!("invalid ", stringify!(#ident), " tag: {}"), tag),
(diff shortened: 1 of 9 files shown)


caspermeijn (changes requested):
The changes look good. Just a one comment: You should add a test to confirm your example works.

shizzard:
> You should add a test to confirm your example works.
@caspermeijn I thought about it, that seem to be a good practice, but this test will only fail if somebody will add the generic type parameter with the same name (e.g. `pub fn encode<T>...` will still pass). It seems impossible to create a useful test that will ensure no shadowing will occur in the future. Are you sure this test is a good addition?

shizzard:
@caspermeijn Alright, what is the best place to put this test? I'm not really familiar with the prost project structure, can you help me with that?

caspermeijn:
> @caspermeijn Alright, what is the best place to put this test? I'm not really familiar with the prost project structure, can you help me with that?

You could do something similar to this: https://github.com/tokio-rs/prost/blob/master/tests/src/no_unused_results.proto

If you search for `no_unused_results`, you can find all the places where you need to add something.

shizzard:
The test is `no_shadowed_types`, I checked it was successfully executed:
```
...
test no_shadowed_types::dummy ... ok
...
```

caspermeijn (approved):
Looks good to me. This is a breaking change, so it has to wait for the next major release. Thank you for your contribution.

caspermeijn:
> @caspermeijn latest commit changes the rest of `::prost::Message` trait as you noted in discord.

Thank you. I am sorry to keep finding more stuff to do: In `prost/src/encoding.rs` and in `prost/src/lib.rs` are some more easy to replace `where B: Buf` statements. Could you also replace those?

I think when doing this breaking change, we should do all of them at once.

shizzard:
> I think when doing this breaking change, we should do all of them at once.

I did try to make these changes within `encoding.rs`, but there are several cases where it is not possible:
- `FnMut` type does not allow dynamic dispatch
- closures does not allow dynamic dispatch

Also, these functions are doing tight-loop decoding and may gain some performance benefits from static dispatch. There are also no way these functions type parameters may shadow user-defined types.

You still think to make these changes where it is possible?

caspermeijn:
> Also, these functions are doing tight-loop decoding and may gain some performance benefits from static dispatch. There are also no way these functions type parameters may shadow user-defined types.

`impl Buf` will not do a dynamic dispatch, maybe you are thinking of the `dyn` keyword. As I understand, the behavior is the same as the `where B: Buf` statement, but it doesn't have the explicit generic type. So I think it will be fine.

> You still think to make these changes where it is possible?

I agree those two complex `where` case should stay the same. I would like to see the simple cases changes and the more complicated cases can stay as is.

shizzard:
> `impl Buf` will not do a dynamic dispatch, maybe you are thinking of the `dyn` keyword.

Yes, you're right, for some reason I thought that this will do a dynamic dispatch.

I did the changes you wanted; there are 6 more places where the buffer is type parametrized (search for `"B: Buf"` in `encoding.rs` file).
I tried to do the same for `M: Message`, but it broke one test on compile time (with cryptic error message that I don't fully understand actually), so I reverted the `M: Message` type changes.

caspermeijn:
> I did the changes you wanted; there are 6 more places where the buffer is type parametrized (search for `"B: Buf"` in `encoding.rs` file). I tried to do the same for `M: Message`, but it broke one test on compile time (with cryptic error message that I don't fully understand actually), so I reverted the `M: Message` type changes.

That looks good. Thank you for this cleanup. We could always do the `impl Message` at another time.

caspermeijn (approved):

Issue #942, opened 2023-11-14
Can you clarify whether prost works with no_std?


shutej:
I attempted to add the dependencies documented in README.md this morning:

https://github.com/tokio-rs/prost#using-prost-in-a-no_std-crate

...but got the following:

```shell
cargo build
   Compiling bytes v0.5.6
error[E0463]: can't find crate for `std`
  --> /Users/shutej/.cargo/registry/src/index.crates.io-6f17d22bba15001f/bytes-0.5.6/src/lib.rs:82:1
   |
82 | extern crate std;
   | ^^^^^^^^^^^^^^^^^ can't find crate
   |
   = note: the `thumbv7em-none-eabihf` target may not support the standard library

For more information about this error, try `rustc --explain E0463`.
error: could not compile `bytes` (lib) due to previous error
```

shutej:
This is a problem with the version mentioned in the docs.  Use `cargo add prost` instead, folks.

caspermeijn:
As I understand, you used an older version. The new version does support `no_std` as you expect.

The readme has the most recent version number. I have fixed the release script to update the readme during the release process. You are welcome to reopen a PR to use `cargo add`.

PR #1046, opened 2024-04-28
fix: #942

gibbz00:
Changes version numbers to start imports rather than the latest, reducing the need to make sure that the documentation numbers are always up-to-date.

One downside is that the users must set an explicit the version number before publishing their own crates.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,9 +30,9 @@ First, add `prost` and its public dependencies to your `Cargo.toml`:
 
 ```ignore
 [dependencies]
-prost = "0.12"
+prost = "*"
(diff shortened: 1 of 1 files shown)


caspermeijn (commented):
Let's change these instructions to use `cargo add`. I believe it chooses the latest released version, but you need to check that.

https://doc.rust-lang.org/cargo/commands/cargo-add.html

caspermeijn:
I have fixed the release script to update the readme during the release process. Please don't hesitate to reopen the PR to use `cargo add`.

PR #1047, opened 2024-04-28
chore: improve `no_std` maintainability:

gibbz00:
Similar to how `rustls` does it:

https://github.com/rustls/rustls/blob/513e374b2e2ce9f1fb57ac78ab3ca053afc8f133/rustls/src/lib.rs#L353-L359

> ... `extern crate` plus the `#![no_std]` attribute changes the
default prelude from `std::prelude` to `core::prelude`. That forces one to _explicitly_ import (`use`) everything that is in `std::prelude` but not in `core::prelude`. This helps maintain no-std support as even developers that are not interested in, or aware of, no-std support and / or that never run `cargo build --no-default-features` locally will get errors when they rely on `std::prelude` API.


prost-build/src/fixtures/field_attributes/_expected_field_attributes.rs
```diff
@@ -2,33 +2,32 @@
...
     pub data: ::core::option::Option<container::Data>,
 }
 /// Nested message and enum types in `Container`.
 pub mod container {
     #[allow(clippy::derive_partial_eq_without_eq)]
-#[derive(Clone, PartialEq, ::prost::Oneof)]
+    #[derive(Clone, PartialEq, ::prost::Oneof)]
     pub enum Data {
-        #[prost(message, tag="1")]
-        Foo(::prost::alloc::boxed::Box<super::Foo>),
```
caspermeijn, line 14: I believe the generated code should always use a fully qualified type path, as they could otherwise clash with a user-generated type.

prost/src/facade.rs
```diff
@@ -0,0 +1,90 @@
+//! A facade around all the types we need from the `std`, `core`, and `alloc`
+//! crates. This avoids elaborate import wrangling having to happen in every
+//! module.
```
caspermeijn, line 3: I don't like the facade concept. If we keep it, it should work like the `::std::prelude`. But I think it is better to be explicit about your dependencies.

prost/src/lib.rs
```diff
@@ -1,6 +1,10 @@
 #![doc(html_root_url = "https://docs.rs/prost/0.12.2")]
-#![cfg_attr(not(feature = "std"), no_std)]
 #![doc = include_str!("../../README.md")]
+#![no_std]
+
+// See: https://github.com/tokio-rs/prost/pull/1047
```
caspermeijn, line 5: The code should not refer to Github issues, but instead contain the contents directly

caspermeijn (changes requested):
I'm in favor of better `no_std` maintainability, but this PR introduces multiple concepts. I think it would be better reviewable if those are split into multiple PRs.

caspermeijn:
I close this issue due to lack of activity. Feel free to split the PR into smaller PRs that are easily reviewable.

PR #1049, opened 2024-05-01
fix the link and badge for CI

mumbleskates:
the badge image has been incorrect and not reflective of CI status since august 2023 (9c877ce32ec0c465e437144c1bf4a27cb3aa705c).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-![continuous integration](https://github.com/tokio-rs/prost/workflows/continuous%20integration/badge.svg)
+[![continuous integration](https://github.com/tokio-rs/prost/actions/workflows/ci.yml/badge.svg?branch=master)](https://github.com/tokio-rs/prost/actions/workflows/ci.yml?query=branch%3Amaster)
 [![Documentation](https://docs.rs/prost/badge.svg)](https://docs.rs/prost/)
 [![Crate](https://img.shields.io/crates/v/prost.svg)](https://crates.io/crates/prost)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,4 +1,4 @@
-![continuous integration](https://github.com/tokio-rs/prost/workflows/continuous%20integration/badge.svg)
+[![continuous integration](https://github.com/tokio-rs/prost/actions/workflows/ci.yml/badge.svg?branch=master)](https://github.com/tokio-rs/prost/actions/workflows/ci.yml)
```
caspermeijn, line 1: The badge is for branch master. Should the link also be for branch master?

caspermeijn (commented):
So the badge broke due to a rename of the CI workflow, but the URL for the badge stayed valid. Good catch.

mumbleskates:
yeah i have no idea why the old ci badge remained valid and green to this day; seems error prone to me.

README.md
caspermeijn: The search query results in only showing CI results that the badge is referring to, so that makes more sense to me then showing CI results of random PRs. I leave it up to you.

mumbleskates: i guess it does make sense enough to show just master branch. and it's not like fragility based on whatever github decides to break matters much since that's kinda upstream of all the other parts of it anyway

caspermeijn (approved):

caspermeijn:
Thanks for your contribution

Issue #505, opened 2021-07-10
oneof field name can conflict with embedded message names


aquarhead:
This can be reproduced with the existing test for one_of, e.g. with this diff:

```diff
--- a/tests/src/oneof_attributes.proto
+++ b/tests/src/oneof_attributes.proto
@@ -7,8 +7,9 @@ syntax = "proto3";
 package Foo.Custom.OneOfAttrs;
 
 message Msg {
+    message Field {}
     oneof field {
         string a = 1;
-        bytes b = 2;
+        Field b = 2;
     }
 }
```

Then run `cargo test -p tests` will fail with build error:

```
error[E0428]: the name `Field` is defined multiple times
  --> /home/aqd/Projects/OSS/prost/target/debug/build/tests-e7a58fd76f19095c/out/foo.custom.one_of_attrs.rs:13:5
   |
9  |     pub struct Field {
   |     ---------------- previous definition of the type `Field` here
...
13 |     pub enum Field {
   |     ^^^^^^^^^^^^^^ `Field` redefined here
   |
   = note: `Field` must be defined only once in the type namespace of this module
```

The generated code is like this:

```rust
#[derive(Clone, PartialEq, ::prost::Message)]
pub struct Msg {
    #[prost(oneof="msg::Field", tags="1, 2")]
    pub field: ::core::option::Option<msg::Field>,
}
/// Nested message and enum types in `Msg`.
pub mod msg {
    #[derive(Clone, PartialEq, ::prost::Message)]
    pub struct Field {
    }
    #[derive(Eq, PartialOrd, Ord)]
    #[derive(Clone, PartialEq, ::prost::Oneof)]
    pub enum Field {
        #[prost(string, tag="1")]
        A(::prost::alloc::string::String),
        #[prost(message, tag="2")]
        B(Field),
    }
}
```

I would like to contribute a fix for this, I'm thinking of moving all `oneof` fields into a `::oneofs` child module. I considered just suffix/prefix the field name but it could still conflict with other names. Let me know if there are better alternatives?

LucioFranco:
Hmmm I'm not sure if I want to sacrifice ergonomics for this case. I wonder how often this actually comes up in practice where the user can't just change it.

aquarhead:
I do recognize this is a breaking change, but I'm not sure this is necessarily "sacrificing" ergonomic. 

[...]

PR #1059, opened 2024-05-10
Setting custom name for Rust enum resulted from a `oneof` field

yuxuan-xie:
A workaround for #505.

### DIFF
--- a/prost-build/src/code_generator.rs
+++ b/prost-build/src/code_generator.rs
@@ -575,11 +575,8 @@ impl<'a> CodeGenerator<'a> {
         fq_message_name: &str,
         oneof: &OneofField,
     ) {
-        let type_name = format!(
-            "{}::{}",
-            to_snake(message_name),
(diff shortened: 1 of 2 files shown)


caspermeijn (commented):
I understand your usecase for this workaround. However, I feel like `prost-build` should not provide config options to rename proto elements. 

The names in the proto file should be leading, and renaming will complicate the codebase of prost. If we allow renaming, then we have to keep track of the new names to be able to reference the types at a later time.

yuxuan-xie:
Thanks for the review. I fully understand your concern. Adapting some transparent strategy that does not support explicit renaming is much more advantageous in terms of maintenance and consistency across the whole codebase. But it seems that the usecase in #505, which is also my usecase, is a bit too much a corner case so developers are not willing to pay too many efforts to this. So for now, I will just use this roughly patched fork in my own project and close this pull request. If someone is willing to fix this problem with a more compatible solution, I am very glad to help.

Issue #1027, opened 2024-04-13
getter/setter generated for optional enum fields should use Option


jmhain:
Currently if you have an enum field marked as `optional`, `prost` wraps the `i32` field in an `Option` as with other scalar values. However, the generated getter in this case still returns `MyEnum` instead of `Option<MyEnum>`, silently converting a `None` into the default value. This burned me recently, and it's also inconsistent with the fact that the `i32` fields are wrapped in `Option`.

giangndm:
@caspermeijn I have the same issues, which make me have to use a trick to convert from Option<i32>. I checked the source code and found that it has some strange behavior in

https://github.com/tokio-rs/prost/blob/baddf9828596bb7e2b908f5d4997542ebe59a7c3/prost-derive/src/field/scalar.rs#L308-326 and https://github.com/tokio-rs/prost/blob/baddf9828596bb7e2b908f5d4997542ebe59a7c3/prost-derive/src/field/scalar.rs#L354-377

It returns a default value with Kind::Optional. I think that with Kind::Optional, we should return None if it's not set.
I tried to change the above code, but it affected a lot of other code, because #[prost(optional)] is used a lot.

For example, with

```rust
pub struct FieldDescriptorProto {
    #[prost(string, optional, tag = "1")]
    pub name: ::core::option::Option<::prost::alloc::string::String>,
```

Instead of checking if name is Some or None, some other code uses a function from a code macro like: 

```rust
impl OneofField {
    fn new(descriptor: OneofDescriptorProto, fields: Vec<Field>, path_index: i32) -> Self {
        Self {
            descriptor,
            fields,
            path_index,
        }
    }

    fn rust_name(&self) -> String {
        to_snake(self.descriptor.name())
    }
}
```

I think we should have another function if we need to get name with a fallback to a default value, like `name_fallback_default`. As I'm new to working with Prost source code, I'm not familiar with it enough to know the best direction. Could you please give me some advice on this case?

caspermeijn:
Would this PR also address your problem? https://github.com/tokio-rs/prost/pull/1079

 It changes the field type to `Option<OpenEnum<MyEnum>>`. `OpenEnum` had a `known()` helper function that returns `Option<MyEnum>`.

mzabaluev:
#1079 does not address this issue, as I did not change the behavior of prost-derive in other aspects than the value type of the enum fields.

PR #1060, opened 2024-05-11
fix getter/setter generated for optional enum fields should use option

giangndm:
This PR fixed mismatch between getter/setter generated for optional enum fields and enum value i32 as described in #1027

### DIFF
--- a/prost-derive/src/field/scalar.rs
+++ b/prost-derive/src/field/scalar.rs
@@ -306,18 +306,33 @@ impl Field {
                     }
                 }
                 Kind::Optional(ref default) => {
-                    let get_doc = format!(
+                    let get_doc_fallback = format!(
                         "Returns the enum value of `{}`, \
                          or the default if the field is unset or set to an invalid enum value.",
                         ident_str,
                     );
+                    let get_doc = format!(
+                        "Returns the enum value of `{}`, \
+                         or None if the field is unset or set to an invalid enum value.",
+                        ident_str,
+                    );
+                    let get_fallback =
+                        Ident::new(&format!("{}_fallback", ident_str), Span::call_site());
                     quote! {
-                        #[doc=#get_doc]
-                        pub fn #get(&self) -> #ty {
+                        #[doc=#get_doc_fallback]
+                        pub fn #get_fallback(&self) -> #ty {
                             self.#ident.and_then(|x| {
                                 let result: ::core::result::Result<#ty, _> = ::core::convert::TryFrom::try_from(x);
                                 result.ok()
                             }).unwrap_or(#default)
+                        }
+
+                        #[doc=#get_doc]
+                        pub fn #get(&self) -> ::core::option::Option<#ty> {
+                            self.#ident.and_then(|x| {
+                                let result: ::core::result::Result<#ty, _> = ::core::convert::TryFrom::try_from(x);
+                                result.ok()
+                            })
                         }
 
                         #[doc=#set_doc]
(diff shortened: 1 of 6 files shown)


tests/src/lib.rs
```diff
@@ -553,13 +553,16 @@ mod tests {
...
+            default_enum_value::PrivacyLevel::One
+        );
+        assert_eq!(
+            msg.privacy_level_3_fallback(),
             default_enum_value::PrivacyLevel::PrivacyLevelThree
         );
         assert_eq!(
-            msg.privacy_level_4(),
+            msg.privacy_level_4_fallback(),
             default_enum_value::PrivacyLevel::PrivacyLevelprivacyLevelFour
```
caspermeijn, line 566: I would turn this around to simplify the code.

```suggestion
            msg.privacy_level_4(),
            Some(default_enum_value::PrivacyLevel::PrivacyLevelprivacyLevelFour)
```

caspermeijn (commented):
Why is the `name_fallback` getter nessesary? You could do `.name().unwrap_or_default()` to get the same result, right? I think that will result in more idiomatic code.

prost-derive/src/field/scalar.rs
```diff
@@ -360,19 +375,38 @@ impl Field {
...
+                #[doc=#get_doc_fallback]
+                pub fn #get_fallback(&self) -> #ty {
                     match self.#ident {
                         #match_some
                         ::core::option::Option::None => #default,
                     }
                 }
+
+                #[doc=#get_doc]
+                pub fn #get(&self) -> ::core::option::Option<#ty> {
```
caspermeijn, line 404: I think this should only return an `Option<T>` for optional protobuf fields. I think it should remain the original code for non-optional fields

caspermeijn (commented):
I would like to see new test cases for an optional protobuf enum field.

giangndm:
> Why is the `name_fallback` getter nessesary? You could do `.name().unwrap_or_default()` to get the same result, right? I think that will result in more idiomatic code.

If we use unwrap_or_default, I don't think it can work with some custom default enum like this:

```
enum PrivacyLevel {
  PRIVACY_LEVEL_ONE = 1;
  PRIVACY_LEVEL_TWO = 2;
  PRIVACY_LEVEL_PRIVACY_LEVEL_THREE = 3;
  PRIVACY_LEVELPRIVACY_LEVEL_FOUR = 4;
}

message Test {
  optional PrivacyLevel privacy_level_1 = 1 [default = PRIVACY_LEVEL_ONE];
  optional PrivacyLevel privacy_level_3 = 2 [default = PRIVACY_LEVEL_PRIVACY_LEVEL_THREE];
  optional PrivacyLevel privacy_level_4 = 3 [default = PRIVACY_LEVELPRIVACY_LEVEL_FOUR];
}
```

Java implementation use `has_ ` for checking and the field is set with default value in this case. How about think about `has_` function?

caspermeijn (changes requested):
This PR should only change the generated code for optional enums. The generated code to non-optional enums should be unchanged. 

I think optional enums should have two getters: `fn #get(&self) -> Option<#ty>` and `fn #get _or_default(&self) -> #ty`. I believe that will cover your usecase and be more correct.

caspermeijn:
@giangndm I am preparing a breaking change. Are you still interested in this PR? If so, please update the code.

giangndm:
@caspermeijn Sure, I will update the code this weekend. 

Ps: Sorry for the late reply, I missed it among ton of Github notifications.

Issue #276, opened 2020-02-03
A more ergonomic enumeration representation


sfackler:
To deal with the "openness" of Protobuf enumeration types, Prost currently represents them in the message struct as `i32` and generates a Rust enum that can convert to and from `i32`. This is a bit awkward to work with, though.

A better approach could be to use a newtype with associated constants. To use the [`Syntax` enum](https://docs.rs/prost-types/0.6.1/prost_types/enum.Syntax.html) as an example, it would look like this:

```rust
#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, ...)]
pub struct Syntax(i32);

impl Syntax {
    pub const PROTO2: Syntax = Syntax(0);
    pub const PROTO3: Syntax = Syntax(1);

    // alternatively the field could just be public
    pub fn from_raw(raw: i32) -> Syntax {
        Syntax(raw)
    }

    pub fn as_raw(self) -> i32 {
        self.0
    }
}
```

This approach has a couple of advantages:

* It can be embedded directly in the message struct.
* It's more type safe since you normally won't have to interact with the raw `i32` value.
* You can still use it roughly like a Rust enum, and do things like match on the variants:
```rust
match syntax {
    Syntax::PROTO2 => { ... }
    Syntax::PROTO3 => { ... }
    syntax => panic!("unknown syntax variant {}", syntax.as_raw()),
}
```

danburkert:
@sfackler are you aware that type-safe getters are generated for enum fields?  For instance: https://docs.rs/prost-types/0.6.1/prost_types/struct.Type.html#method.syntax

danburkert:
I do like this idea, though.  It definitely has some merit.  My experience has been that working with enums through the typesafe getter/setter is 'good enough', but they have a discoverability problem.  I don't think they are actually documented anywhere other than the resulting rustdoc.

sfackler:
Ooh, I did not know about those!

sfackler:
I'd be happy to build out the implementation, but I'd be interested on your thoughts on how to structure it. 0.6 is pretty new, so this could just be a breaking change and prost bumps to 0.7, or it could end up as a new flag in prost_build::Config to allow users to opt into the new approach.

jakeswenson:
@sfackler @danburkert How's this going? 
I have been using prost more and more, and I do find the enums _slightly_ painful; I think this proposal goes a long way to easing that.

sfackler:
I'm going to work on this over the weekend.

[...]

PR #1061, opened 2024-05-12
`OpenEnum`: type-safe wrapper for enum field values

mzabaluev:
A solution to #276 as sketched in https://github.com/tokio-rs/prost/issues/276#issuecomment-2017157041


mzabaluev:
Superseded by #1079.

Issue #1062, opened 2024-05-16
`protoc_from_env` and `protoc_include_from_env` are not exposed in 0.12.5


TheLortex:
Hi, 

This commit https://github.com/tokio-rs/prost/commit/30b22d38f0d29bed2bca6a1cac1a9051fdd2443f introduces a breakage because the `pub` functions `protoc_from_env` and `protoc_include_from_env` are moved in the `config.rs` file.

Because `config.rs` is privately imported, its `pub` functions are not available anymore from outside crates.

PR #1063, opened 2024-05-16
fix(prost-build): re-export `error_message_protoc_not_found`, `protoc_from_env` & `protoc_include_from_env`

eopb:
fixes: #1062

I'm adding this to fix a breaking change introduced in https://github.com/tokio-rs/prost/pull/1020 where these items were moved.

Since `0.12.5` this is causing build failures after `cargo update`.

```
   Compiling grpc-build v6.1.0
error[E0432]: unresolved imports `prost_build::protoc_from_env`, `prost_build::protoc_include_from_env`
 --> /Users/ethan.brierley/.cargo/registry/src/index.crates.io-6f17d22bba15001f/grpc-build-6.1.0/src/lib.rs:3:19
  |
3 | use prost_build::{protoc_from_env, protoc_include_from_env, Module};
  |                   ^^^^^^^^^^^^^^^  ^^^^^^^^^^^^^^^^^^^^^^^ no `protoc_include_from_env` in the root
  |                   |
  |                   no `protoc_from_env` in the root
```

### DIFF
--- a/prost-build/src/lib.rs
+++ b/prost-build/src/lib.rs
@@ -149,7 +149,9 @@ mod message_graph;
 mod path;
 
 mod config;
-pub use config::Config;
+pub use config::{
+    error_message_protoc_not_found, protoc_from_env, protoc_include_from_env, Config,
+};
 
 mod module;
 pub use module::Module;


eopb:
@gibbz00 
Looking at #1015 I'm guessing that making these private was intentional. I'm guessing this was expected to be included only in a breaking release. If you'd like these to be private you can open a follow-up PR deprecating the re-exports so that the change remains non-breaking. The re-exports can then be removed in `0.13.0`.

LucioFranco:
Yup, let me yank v0.12.5, we didn't think this was going to cause an issue but clearly it does 😄 thanks for the report.

LucioFranco:
It's been yanked, we will work on a follow up release that fixes these semver issues.

prost-build/src/lib.rs
```diff
@@ -149,7 +149,7 @@ mod message_graph;
 mod path;
 
 mod config;
-pub use config::Config;
+pub use config::{protoc_from_env, protoc_include_from_env, Config};
```
caspermeijn, line 152: We should restore all move `pub` functions. This version will satisfy cargo-semver-checks:

```suggestion
pub use config::{error_message_protoc_not_found, protoc_from_env, protoc_include_from_env, Config};
```

caspermeijn (changes requested):
Sorry, I accepted the change that broke your build. I was naive of me to assume this is an “internal” function and nobody is using it.

prost-build/src/lib.rs
eopb: > Sorry, I accepted the change that broke your build. I was naive of me to assume this is an “internal” function and nobody is using it.

No worries. It was great to see it quickly yanked and sorted :raised_hands:. Thanks for spending your time working on the crates I use :)

> We should restore all moved pub functions. This version will satisfy cargo-semver-checks:

:+1: https://github.com/tokio-rs/prost/compare/1e5e904b47773b92ea46ba6e82feaaa2f006a9f4..f4be7b750a6039c8535ae3e105cf912c3b0c12e7

caspermeijn: > No worries. It was great to see it quickly yanked and sorted 🙌. Thanks for spending your time working on the crates I use :)

Well, you created this PR half an hour after publish. So, I am impressed with that as well :smiley:

caspermeijn (approved):

PR #1075, opened 2024-05-24
Feature so that prost-types depends on prost-derive only optionally.

ernoc:


caspermeijn:
Can you explain when this is useful?

ernoc:
When using Bazel+rules_rust instead of Cargo to, depending on prost-derive makes the entire crate index require std, which makes it impossible to build for bare metal. One workaround is to not depend on prost-derive as a "dep", and add the dependency as a proc macro (e.g. https://github.com/project-oak/oak/blob/88928b96e1fe72d173c106f511d7f6537ed30000/micro_rpc/BUILD#L30) .

The approach in the PR, however, doesn't seem to work, unless we were open to changing CI tests to always have "derive" feature active for prost-types.

PD: I'm making this PR a draft for now.

caspermeijn:
> When using Bazel+rules_rust instead of Cargo, depending on prost-derive makes the entire crate index require std, which makes it impossible to build for bare metal. One workaround is to not depend on prost-derive as a "dep", and add the dependency as a proc macro (e.g. https://github.com/project-oak/oak/blob/88928b96e1fe72d173c106f511d7f6537ed30000/micro_rpc/BUILD#L30) .

That seems like a bug in bazel. Because `prost-derive` only runs on the host.

> The approach in this PR, however, doesn't seem to work, unless we were open to changing CI tests to always have "derive" feature active for prost-types.

I believe that `prost-types` without `prost-derive` is not useful, as it could never compile all the useful types in `protobuf.rs`. Am I right?

caspermeijn:
I will close this for now. Please re-open when you think you have new information.

PR #1079, opened 2024-05-28
`OpenEnum`: an enum to represent protobuf's enumeration field values

mzabaluev:
Another solution for #276, amenable to destructuring.

### DIFF
--- a/prost-derive/src/field/map.rs
+++ b/prost-derive/src/field/map.rs
@@ -132,13 +132,13 @@ impl Field {
         let module = self.map_ty.module();
         match &self.value_ty {
             ValueTy::Scalar(scalar::Ty::Enumeration(ty)) => {
-                let default = quote!(#ty::default() as i32);
+                let default = quote!(::prost::OpenEnum::from(#ty::default()));
                 quote! {
                     ::prost::encoding::#module::encode_with_default(
                         #ke,
                         #kl,
-                        ::prost::encoding::int32::encode,
-                        ::prost::encoding::int32::encoded_len,
+                        ::prost::encoding::enumeration::encode,
+                        ::prost::encoding::enumeration::encoded_len,
                         &(#default),
                         #tag,
                         &#ident,
@@ -184,11 +184,11 @@ impl Field {
         let module = self.map_ty.module();
         match &self.value_ty {
             ValueTy::Scalar(scalar::Ty::Enumeration(ty)) => {
-                let default = quote!(#ty::default() as i32);
+                let default = quote!(::prost::OpenEnum::from(#ty::default()));
                 quote! {
                     ::prost::encoding::#module::merge_with_default(
                         #km,
-                        ::prost::encoding::int32::merge,
+                        ::prost::encoding::enumeration::merge,
                         #default,
                         &mut #ident,
                         buf,
@@ -221,11 +221,11 @@ impl Field {
         let module = self.map_ty.module();
         match &self.value_ty {
             ValueTy::Scalar(scalar::Ty::Enumeration(ty)) => {
-                let default = quote!(#ty::default() as i32);
+                let default = quote!(::prost::OpenEnum::from(#ty::default()));
                 quote! {
                     ::prost::encoding::#module::encoded_len_with_default(
                         #kl,
-                        ::prost::encoding::int32::encoded_len,
+                        ::prost::encoding::enumeration::encoded_len,
                         &(#default),
                         #tag,
                         &#ident,
@@ -275,17 +275,11 @@ impl Field {
             Some(quote! {
                 #[doc=#get_doc]
                 pub fn #get(&self, key: #key_ref_ty) -> ::core::option::Option<#ty> {
-                    self.#ident.get(#take_ref key).cloned().and_then(|x| {
-                        let result: ::core::result::Result<#ty, _> = ::core::convert::TryFrom::try_from(x);
-                        result.ok()
-                    })
+                    self.#ident.get(#take_ref key).cloned().and_then(|x| { x.known() })
                 }
                 #[doc=#insert_doc]
                 pub fn #insert(&mut self, key: #key_ty, value: #ty) -> ::core::option::Option<#ty> {
-                    self.#ident.insert(key, value as i32).and_then(|x| {
-                        let result: ::core::result::Result<#ty, _> = ::core::convert::TryFrom::try_from(x);
-                        result.ok()
-                    })
+                    self.#ident.insert(key, value.into()).and_then(|x| { x.known() })
                 }
             })
         } else {
--- a/prost-derive/src/field/scalar.rs
+++ b/prost-derive/src/field/scalar.rs
@@ -216,10 +216,10 @@ impl Field {
     fn debug_inner(&self, wrap_name: TokenStream) -> TokenStream {
         if let Ty::Enumeration(ref ty) = self.ty {
             quote! {
-                struct #wrap_name<'a>(&'a i32);
+                struct #wrap_name<'a>(&'a ::prost::OpenEnum<#ty>);
                 impl<'a> ::core::fmt::Debug for #wrap_name<'a> {
                     fn fmt(&self, f: &mut ::core::fmt::Formatter) -> ::core::fmt::Result {
-                        let res: ::core::result::Result<#ty, _> = ::core::convert::TryFrom::try_from(*self.0);
+                        let res = self.0.known_or(());
                         match res {
                             Err(_) => ::core::fmt::Debug::fmt(&self.0, f),
                             Ok(en) => ::core::fmt::Debug::fmt(&en, f),
@@ -297,12 +297,12 @@ impl Field {
                     quote! {
                         #[doc=#get_doc]
                         pub fn #get(&self) -> #ty {
-                            ::core::convert::TryFrom::try_from(self.#ident).unwrap_or(#default)
+                            self.#ident.unwrap_or(#default)
                         }
 
                         #[doc=#set_doc]
                         pub fn #set(&mut self, value: #ty) {
-                            self.#ident = value as i32;
+                            self.#ident = value.into();
                         }
                     }
                 }
@@ -315,15 +315,12 @@ impl Field {
                     quote! {
                         #[doc=#get_doc]
                         pub fn #get(&self) -> #ty {
-                            self.#ident.and_then(|x| {
-                                let result: ::core::result::Result<#ty, _> = ::core::convert::TryFrom::try_from(x);
-                                result.ok()
-                            }).unwrap_or(#default)
+                            self.#ident.and_then(|x| { x.known() }).unwrap_or(#default)
                         }
 
                         #[doc=#set_doc]
                         pub fn #set(&mut self, value: #ty) {
-                            self.#ident = ::core::option::Option::Some(value as i32);
+                            self.#ident = ::core::option::Option::Some(value.into());
                         }
                     }
                 }
@@ -334,20 +331,18 @@ impl Field {
                     );
                     let push = Ident::new(&format!("push_{}", ident_str), Span::call_site());
                     let push_doc = format!("Appends the provided enum value to `{}`.", ident_str);
+                    let wrapped_ty = quote!(::prost::OpenEnum<#ty>);
                     quote! {
                         #[doc=#iter_doc]
                         pub fn #get(&self) -> ::core::iter::FilterMap<
-                            ::core::iter::Cloned<::core::slice::Iter<i32>>,
-                            fn(i32) -> ::core::option::Option<#ty>,
+                            ::core::iter::Cloned<::core::slice::Iter<#wrapped_ty>>,
+                            fn(#wrapped_ty) -> ::core::option::Option<#ty>,
                         > {
-                            self.#ident.iter().cloned().filter_map(|x| {
-                                let result: ::core::result::Result<#ty, _> = ::core::convert::TryFrom::try_from(x);
-                                result.ok()
-                            })
+                            self.#ident.iter().cloned().filter_map(|x| { x.known() })
                         }
                         #[doc=#push_doc]
                         pub fn #push(&mut self, value: #ty) {
-                            self.#ident.push(value as i32);
+                            self.#ident.push(value.into());
                         }
                     }
                 }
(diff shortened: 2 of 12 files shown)


QuentinPerez:
Do you think that this logic could be applied to the oneof field ?

prost-derive/src/field/map.rs
```diff
@@ -275,17 +275,11 @@ impl Field {
...
-                    })
+                    self.#ident.get(#take_ref key).cloned().and_then(|x| { x.known() })
                 }
                 #[doc=#insert_doc]
                 pub fn #insert(&mut self, key: #key_ty, value: #ty) -> ::core::option::Option<#ty> {
-                    self.#ident.insert(key, value as i32).and_then(|x| {
-                        let result: ::core::result::Result<#ty, _> = ::core::convert::TryFrom::try_from(x);
-                        result.ok()
-                    })
+                    self.#ident.insert(key, value.into()).and_then(|x| { x.known() })
```
caspermeijn, line 282: Is it beneficial to provide a `get` and `insert` function with the new wrapper? I feel like the map can be used directly with the helper functions provided by OpenEnum.

prost-derive/src/field/scalar.rs
```diff
@@ -297,12 +296,12 @@ impl Field {
...
                         pub fn #get(&self) -> #ty {
-                            ::core::convert::TryFrom::try_from(self.#ident).unwrap_or(#default)
+                            self.#ident.unwrap_or(#default)
                         }
 
                         #[doc=#set_doc]
                         pub fn #set(&mut self, value: #ty) {
-                            self.#ident = value as i32;
+                            self.#ident = value.into();
                         }
```
caspermeijn, line 305: Is it still beneficial to provide a `get`, `set` and `push` functions with the new wrapper? I feel like the field can be used directly with the helper functions provided by OpenEnum.

prost-derive/src/field/scalar.rs
```diff
@@ -555,13 +550,13 @@ impl Ty {
             Ty::Bool => quote!(bool),
             Ty::String => quote!(&str),
             Ty::Bytes(..) => quote!(&[u8]),
-            Ty::Enumeration(..) => quote!(i32),
+            Ty::Enumeration(..) => unreachable!("an enum should never be queried for its ref type"),
```
caspermeijn, line 553: Can you explain me why it should never be queried by ref?

prost-derive/src/field/scalar.rs
```diff
@@ -555,13 +550,13 @@ impl Ty {
...
         }
     }
 
     pub fn module(&self) -> Ident {
         match *self {
-            Ty::Enumeration(..) => Ident::new("int32", Span::call_site()),
+            Ty::Enumeration(..) => Ident::new("enumeration", Span::call_site()),
```
caspermeijn, line 559: Does it make sense to use `self.as_str()` for enumerations as well?

prost/src/encoding.rs
```diff
@@ -721,6 +721,128 @@ fixed_width!(
...
+    }
+
+    pub fn encode_repeated<T, B>(tag: u32, values: &[OpenEnum<T>], buf: &mut B)
+    where
+        T: Clone + Into<i32>,
+        B: BufMut,
+    {
+        for value in values {
+            encode(tag, value, buf);
+        }
```
caspermeijn, line 759: Could `int32::encode_repeated` be reused here?

prost/src/encoding.rs
```diff
@@ -721,6 +721,128 @@ fixed_width!(
...
+        encode_key(tag, WireType::LengthDelimited, buf);
+        let len: usize = values
+            .iter()
+            .map(|value| encoded_len_varint(value.to_raw() as u64))
+            .sum();
+        encode_varint(len as u64, buf);
+
+        for value in values {
+            encode_varint(value.to_raw() as u64, buf);
+        }
```
caspermeijn, line 780: Could `int32::encode_packed` be reused here?

prost/src/encoding.rs
```diff
@@ -721,6 +721,128 @@ fixed_width!(
...
+                Ok(())
+            })
+        } else {
+            // Unpacked.
+            check_wire_type(WireType::Varint, wire_type)?;
+            let mut value = Default::default();
+            merge(wire_type, &mut value, buf, ctx)?;
+            values.push(value);
+            Ok(())
+        }
```
caspermeijn, line 809: Could `int32::merge_repeated` be reused here?

prost/src/encoding.rs
```diff
@@ -721,6 +721,128 @@ fixed_width!(
...
+
+    pub fn encoded_len_repeated<T>(tag: u32, values: &[OpenEnum<T>]) -> usize
+    where
+        T: Clone + Into<i32>,
+    {
+        key_len(tag) * values.len()
+            + values
+                .iter()
+                .map(|value| encoded_len_varint(value.to_raw() as u64))
+                .sum::<usize>()
```
caspermeijn, line 827: Could `int32::encoded_len_repeated` be reused here?

prost/src/encoding.rs
```diff
@@ -721,6 +721,128 @@ fixed_width!(
...
+    {
+        if values.is_empty() {
+            0
+        } else {
+            let len = values
+                .iter()
+                .map(|value| encoded_len_varint(value.to_raw() as u64))
+                .sum::<usize>();
+            key_len(tag) + encoded_len_varint(len as u64) + len
+        }
```
caspermeijn, line 842: Could `int32::encoded_len_packed` be reused here?

prost/src/open_enum.rs
```diff
@@ -0,0 +1,198 @@
...
+        &mut self,
+        tag: u32,
+        wire_type: WireType,
+        buf: &mut impl Buf,
+        ctx: DecodeContext,
+    ) -> Result<(), DecodeError> {
+        let mut raw = 0;
+        <i32 as Message>::merge_field(&mut raw, tag, wire_type, buf, ctx)?;
+        *self = OpenEnum::from_raw(raw);
+        Ok(())
```
caspermeijn, line 192: This is the same function as `encoding::enumeration::merge`, right?

caspermeijn:
Will this break compatibility with proto2/closed enums? Personally, I don't use proto2, so I am not sure whether it is properly supported at all.

caspermeijn:
I think it makes sense to provide a migration guide.

caspermeijn:
What is the error message for a `i32` field with `#[prost(enumeration)]`? Can we detect that scenario and print a nice error message?

caspermeijn:
I would like to see some tests for  `OpenEnum`.

caspermeijn (changes requested):

prost-derive/src/field/scalar.rs
mzabaluev: There is some convenience in the setter method:

```rust
msg.set_kind(Kind::Foo)
```

is nicer and more type-ahead friendly than

```rust
msg.kind = Kind::Foo.into()
```

The same applies to the push method for repeated fields.

The getter hides a bit much opinionated behavior, in my opinion, so it might be better to leave it to `OpenEnum` helper methods to explicitly decide on how unknown values should be treated.

prost-derive/src/field/map.rs
mzabaluev: As can be seen in the generated code itself, these methods provide considerable convenience for the most common use cases.

mzabaluev:
> Will this break compatibility with proto2/closed enums? Personally, I don't use proto2, so I am not sure whether it is properly supported at all.

I believe prost has never been conformant with closed enums: the raw values have been left in the message as is.
I think we could try to support closed enums in two ways:

1. Always represent an enum field as `OpenEnum`, but in the closed enum case, decode unknown values as `Known(Default::default())`. This leaves the possibility of producing unknown values on encoding.
2. Represent closed enums with their generated Rust enum type. Decode unknown values as `Default::default()`.

I prefer option 2, even though it requires more work. Protobuf edition 2023 has closed enums as a feature, so they need to be supported even if we ignore proto2.

prost-derive/src/field/map.rs
mzabaluev: Perhaps the insert method could be changed to return the possible old value as `Option<OpenEnum>` so the caller can choose the fallback behavior for unknown values. So yes, the only added convenience there would be the `.into()` conversion call.

prost-derive/src/field/scalar.rs
mzabaluev: Another thing to consider here is porting the existing code. If these methods get used instead of direct access to fields, which I assume happens a lot now because the fields are just `i32` or `Option<i32>`, updating to the version that generates `OpenEnum` would not require changes.

mzabaluev:
> I think it makes sense to provide a migration guide.

What would be a good place for it? I can add a section to the README.

mzabaluev:
> I would like to see some tests for `OpenEnum`.

I'm going to add at least one good example/doctest on the type.

prost/src/encoding.rs
mzabaluev: No, unless `int32::encode_repeated` is generalized to take an `impl IntoIterator` that produces `i32`. Which might be a good idea.

prost-derive/src/field/scalar.rs
caspermeijn: I think the better API is to have some sort of `set ` operation on `OpenEnum`. Maybe something like:
```rust
msg.kind.set(Kind::Foo)
```

Well, this change will break existing code in many ways, so I think we should directly go for the best API

caspermeijn:
Have thought some more about this PR: I don't want to break users in this way. At least not now.

I suggest making this an option in `prost-build` so that we can experiment with the API without breaking existing users. That way, interested users can opt in, and we don't have to create a perfect `OpenEnum` API on the first try. 

Once we feel good about the new API, we can think about changing the default.

Please look at [bytes](https://docs.rs/prost-build/latest/prost_build/struct.Config.html#method.bytes) for a good example of changing the generated data type.

ArjixWasTaken:
Instead of introducing a new type, a simple `Result<T, i32>` sounds more logical to me.
[`enum-repr-derive`](https://github.com/ssalonen/enum-repr-derive) follows this approach when parsing an enum from i32

mzabaluev:
> Instead of introducing a new type, a simple `Result<T, i32>` sounds more logical to me.

It's weird to have a `Result` as a struct field. The names of variants and methods of `Result` are less than intuitively applicable: it's not necessarily an error to receive an unknown enum value from the wire, so we should give the API users a speed bump to decide how to deal with them. There are convenience methods and the `TryInto` impl to convert the `OpenEnum` value to a `Result` if that's the chosen approach.

Another benefit of introducing a new type is for the add-on macros and code generators that derive something on structs generated by prost-build. These could deal with the `OpenEnum` type in some specific way, even without access to the proto descriptor data for the field. Doing the same (e.g. defining generic trait impls) for `Result` would feel like too much overloading.

caspermeijn:
How are default values handles in this solution? Especially default values set for a specific field in the proto file.

prost-derive/src/field/scalar.rs
mzabaluev: No, that would give "enum", but we can't use that as a module name. OTOH, maybe the current behavior of `Ty::as_str` should be changed: "enum" is not a real type name anyway.

prost/src/encoding.rs
mzabaluev: So that would need a breaking change for `encode_repeated` which I don't think is worth the little code duplication at this point.

prost/src/encoding.rs
mzabaluev: Same here, not worth changing the signature of `int32::encode_packed` for this.

prost/src/encoding.rs
mzabaluev: No, definitely not. This is merging a passed in `&mut Vec<OpenEnum<T>>`.

prost/src/encoding.rs
mzabaluev: Not without breaking changes, as explained above.

prost/src/encoding.rs
mzabaluev: Not without breaking changes, as explained above.

prost/src/open_enum.rs
mzabaluev: No, `<i32 as Message>::merge_field` calls to `skip_field` if the tag argument is not 1.

mzabaluev:
> How are default values handles in this solution? Especially default values set for a specific field in the proto file.

Good question. Maybe there should be a const generic parameter for this, but it will then crop up everywhere.

prost-derive/src/field/scalar.rs
```diff
@@ -799,7 +749,7 @@ impl DefaultValue {
 
     pub fn typed(&self) -> TokenStream {
         if let DefaultValue::Enumeration(_) = *self {
-            quote!(#self as i32)
+            quote!(::prost::OpenEnum::from(#self))
```
mzabaluev, line 752: This handles the custom default values for message fields.

mzabaluev:
The generated getter methods handled per-field defaults for optional fields. For non-optional fields, the default is initialized before merging field values from the wire.

I'm not sure how best to proceed here. Suppose we have this generated message:

```rust
pub struct Msg {
    #[prost(enumeration = "AnEnum", optional, default = "AnEnum::B")]
    pub field: Option<OpenEnum<AnEnum>>
}
```

It would be nice to have API implementing the default value behavior for proto2 and edition 2023.

[The protobuf guide] recommends a getter method that returns the default value, but whether out-of-range values are exposed is up to the `enum_type` protobuf feature. The "hazzer" method can be generated to check presence, but in the currently produced code one can just use `Option::is_some`.

mzabaluev:
> I suggest making this an option in `prost-build` so that we can experiment with the API without breaking existing users. That way, interested users can opt in, and we don't have to create a perfect `OpenEnum` API on the first try.

I have addressed this and most of the other comments in recent commits. There is now a `Config::typed_enum_fields` method to opt into type-checked fields.

The proto2 behavior (and the closed enums feature in future editions) is also implemented for enums, albeit with some current limitations: state available to the code generator does not track which syntax the enum types were defined in, that is needed for proto2 ← proto3 includes. I'll try to resolve this.

Getter methods are retained for the optional fields for convenience and to implement per-field default values. Protobuf guidelines also insist on there being getters returning non-nullable values.

mzabaluev:
@caspermeijn please remove the breaking change label as the revised implementation is purely opt-in.

mzabaluev:
> The proto2 behavior (and the closed enums feature in future editions) is also implemented for enums, albeit with some current limitations: state available to the code generator does not track which syntax the enum types were defined in, that is needed for proto2 ← proto3 includes. I'll try to resolve this.

This has been implemented in https://github.com/tokio-rs/prost/pull/1079/commits/46b14ef380e1c22b3eb4824964a22d53ef1b0a23
I should add a test for the enum type "stickiness" across includes soon.

mzabaluev:
@ArjixWasTaken I have updated the description, thank you for the suggestion.

PR #1080, opened 2024-06-03
Allow encode BufMut arg to be unsized

dspyz-matician:
This means it's possible to pass in a `&mut dyn BufMut`

This allows for creating a custom "zero-copy" implementor of `Message` for bytes fields which are themselves encoded protobuf values as follows (useful when trying to model generics with protobuf):

```
trait MySerializer {
  fn serialize_to(&self, b: &mut dyn BufMut);
  fn encoded_len(&self) -> usize;
}

/// message MessageValue {
///   bytes serialized_value = 1;
/// }
#[derive_where(Debug)]
pub enum MessageValue {
    Empty,
    Encoding(#[derive_where(skip)] Arc<dyn MySerializer>),
    Decoding(#[derive_where(skip)] Bytes),
}

impl From<Arc<dyn MySerializer>> for MessageValue {
    fn from(val: Arc<dyn MySerializer>) -> Self {
        MessageValue::Encoding(val)
    }
}

impl Message for MessageValue {
    fn encode_raw<B>(&self, mut buf: &mut (impl bytes::BufMut + ?Sized))
    where
        Self: Sized,
    {
        match self {
            MessageValue::Empty => (),
            MessageValue::Encoding(val) => {
                encode_metadata(val.encoded_len(), buf);
                val.serialize_to(&mut buf);
            }
            MessageValue::Decoding(val) => {
                encode_metadata(val.len(), buf);
                buf.put_slice(&val)
            }
        }
    }

    fn merge_field<B>(
        &mut self,
        tag: u32,
        wire_type: WireType,
        buf: &mut B,
        _ctx: DecodeContext,
    ) -> Result<(), prost::DecodeError>
    where
        B: bytes::Buf,
        Self: Sized,
    {
        if tag != 1 {
            return Ok(());
        }
        if wire_type != WireType::LengthDelimited {
            return Err(prost::DecodeError::new(format!(
                "invalid wire type; expected length delimited, got {wire_type:?}"
            )));
        }
        let len = prost::encoding::decode_varint(buf)?;
        *self = MessageValue::Decoding(buf.copy_to_bytes(len as usize));
        Ok(())
    }

    fn encoded_len(&self) -> usize {
        match self {
            MessageValue::Empty => 0,
            MessageValue::Encoding(val) => message_len_from_byte_len(val.encoded_len()),
            MessageValue::Decoding(val) => message_len_from_byte_len(val.len()),
        }
    }

    fn clear(&mut self) {
        *self = Self::Empty;
    }
}

fn encode_metadata(encoded_len: usize, buf: &mut impl bytes::BufMut) {
    prost::encoding::encode_key(1, prost::encoding::WireType::LengthDelimited, buf);
    prost::encoding::encode_varint(encoded_len as u64, buf);
}

fn message_len_from_byte_len(encoded_len: usize) -> usize {
    1 + prost::encoding::encoded_len_varint(encoded_len as u64) + encoded_len
}
```

This change won't break compilation for existing uses of `prost-build`, but custom implementations will have to add the `+ ?Sized` param, so it's not fully backwards-compatible

### DIFF
--- a/prost/src/encoding.rs
+++ b/prost/src/encoding.rs
@@ -22,7 +22,7 @@ use crate::Message;
 /// Encodes an integer value into LEB128 variable length format, and writes it to the buffer.
 /// The buffer must have enough remaining space (maximum 10 bytes).
 #[inline]
-pub fn encode_varint(mut value: u64, buf: &mut impl BufMut) {
+pub fn encode_varint(mut value: u64, buf: &mut (impl BufMut + ?Sized)) {
     // Varints are never more than 10 bytes
     for _ in 0..10 {
         if value < 0x80 {
@@ -292,7 +292,7 @@ impl TryFrom<u64> for WireType {
 /// Encodes a Protobuf field key, which consists of a wire type designator and
 /// the field tag.
 #[inline]
-pub fn encode_key(tag: u32, wire_type: WireType, buf: &mut impl BufMut) {
+pub fn encode_key(tag: u32, wire_type: WireType, buf: &mut (impl BufMut + ?Sized)) {
     debug_assert!((MIN_TAG..=MAX_TAG).contains(&tag));
     let key = (tag << 3) | wire_type as u32;
     encode_varint(u64::from(key), buf);
@@ -403,7 +403,7 @@ pub fn skip_field(
 /// Helper macro which emits an `encode_repeated` function for the type.
 macro_rules! encode_repeated {
     ($ty:ty) => {
-        pub fn encode_repeated(tag: u32, values: &[$ty], buf: &mut impl BufMut) {
+        pub fn encode_repeated(tag: u32, values: &[$ty], buf: &mut (impl BufMut + ?Sized)) {
             for value in values {
                 encode(tag, value, buf);
             }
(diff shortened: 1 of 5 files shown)


prost/src/encoding.rs
```diff
@@ -863,8 +863,8 @@ impl sealed::BytesAdapter for Bytes {
         *self = buf.copy_to_bytes(buf.remaining());
     }
 
-    fn append_to(&self, buf: &mut impl BufMut) {
-        buf.put(self.clone())
+    fn append_to(&self, buf: &mut (impl BufMut + ?Sized)) {
+        buf.put_slice(self)
```
dspyz-matician, line 867: Is this correct? Why was it the other way before?

prost/src/encoding.rs
```diff
@@ -881,15 +881,15 @@ impl sealed::BytesAdapter for Vec<u8> {
         self.put(buf);
     }
 
-    fn append_to(&self, buf: &mut impl BufMut) {
-        buf.put(self.as_slice())
+    fn append_to(&self, buf: &mut (impl BufMut + ?Sized)) {
+        buf.put_slice(self.as_slice())
```
dspyz-matician, line 885: Here too

dspyz-matician:
Never mind. This doesn't actually work without more invasive changes

dspyz-matician:
Okay, actually this does work. I was just missing one small piece

Issue #624, opened 2022-04-13
Roadmap to 1.0


LucioFranco:
- [ ] Make `prost-build` modular
- [ ] Add Json support
- [ ] Unknown fields
- [ ] Extensions

cc @danburkert @neoeinstein

PR #1081, opened 2024-06-13
WIP: Add json (serde) support

cynecx:
I started this work because I'm also working on adding Connect RPC support to the rust ecosystem (currently targeting tonic). So having proper protobuf json support is the first step into that goal.

This (unfortunately very big) PR adds json/serde support to prost. This is still a WIP effort but it is almost feature complete and passes all conformance tests except tests related to fieldmasks.

A non-exhaustive list of TODOs:

- Decide whether the whole "custom serde" design is the one we want to persue
- Properly document code and public api
- ...

So I am just putting this up here to get a general vibe check...


tests/src/lib.rs
```diff
@@ -246,6 +248,136 @@ where
...
+    /// or it could indicate that the data was invalid (eg. violating message invariants).
+    EncodeError(anyhow::Error),
+    /// The data could not be decoded. This could indicate a bug in prost,
+    /// or it could indicate that the input was bogus.
+    DecodeError(anyhow::Error),
+    /// Re-encoding or validating the data failed.  This indicates a bug in `prost`.
+    Error(anyhow::Error),
+}
+
+pub fn roundtrip2<M>(
```
caspermeijn, line 276: Why is a new roundtrip function needed? Can the original roundtrip be improved?

prost-types/src/any_v2.rs
```diff
@@ -0,0 +1,881 @@
...
+    Message, Name,
+};
+
+use crate::smallbox::{smallbox, SmallBox};
+
+mod private {
+    pub trait Sealed {}
+}
+
+pub trait AnyValue: CoreAny + Message + private::Sealed {
```
caspermeijn, line 21: Can you explain what any v2 is and why it is better?

prost-build/src/code_generator.rs
```diff
@@ -461,9 +472,14 @@ impl<'a> CodeGenerator<'a> {
             }
         }
 
+        if is_well_known_ty {
```
caspermeijn, line 475: Why is it important to know whether it is a well known type? Or asked differently: Why is well known type special?

caspermeijn (commented):
This is impressive work and this feature is on the project roadmap (#624). However this PR is way too big to review. I did scroll through the code and ask some questions that came to mind.

Can you split this work into smaller PRs?

tests/src/lib.rs
cynecx: This is just a thing that I haven't cleaned up yet. The new function is intended to replace the old one, I just haven't gotten to replace all the uses of the old one yet 😅.

prost-types/src/any_v2.rs
cynecx: This basically tries to address the "improved Any" bulletpoint in #624. The one Any implementation provides support for JSON and "cross" serialization (JSON <-> Binary).

The summary for that change is basically going from:

```rust
struct OldAny {
    type_path: String,
    data: Vec<u8>,
}
```

to

```rust
struct NewAny {
    type_path: String,
    kind: enum {
        Protobuf(Vec<u8>),
        Json(JsonValue),
        Dyn(Box<dyn AnyValue>),
    }
}
```

prost-build/src/code_generator.rs
cynecx: I mentioned this quite shortly in `prost-types/serde.rs`:

```
// Due to the orphan rules we can't really implement prost::serde::CustomSerialize for our
// well-known types that are primitives. So we have to use a local wrapper type instead.
```

Tbqh, I don't remember the exact details, but I went with putting these `CustomSerialize/CustomDeserialize` impls for scalar types in the prost-types crate instead of the main `prost`crate due to separation concerns. And I generally though it would give us more flexibility to avoid the orphan rules whenever we'd have to provide custom serialization impls.

cynecx:
> However this PR is way too big to review.
> Can you split this work into smaller PRs?

Yeah, it's actually on the TODO list. I just still haven't gotten to actually complete this as a whole yet.

prost-build/src/code_generator.rs
cynecx: But I do realized now that this approach might be flawed because now you'd need to always include the `prost-types` crate to have serialization working for all "built-in" types. That's certainly a no-go, so I'll try to put back the impls back into `prost` and will eventually remove the whole "is well known type" treatment.

prost-derive/src/field/mod.rs
```diff
@@ -172,6 +173,32 @@ impl Field {
             _ => None,
         }
     }
+
+    pub fn json(&self) -> Option<Option<&Json>> {
```
ehiggs, line 177: Wouldn't it make more sense to flatten this into `Option<&Json>`

cynecx: Conceptionally yes, but this is actually an actual tri-state kind of thing, because (iirc) we are skipping code generation for `Self::OneOf` fields. But we could also make this more flat and just add another check for an oneof field at the callsite 🤷‍♂️.

prost-types/src/duration.rs
```diff
@@ -107,7 +111,7 @@ impl fmt::Display for Duration {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         let mut d = *self;
         d.normalize();
-        if self.seconds < 0 && self.nanos < 0 {
+        if self.seconds < 0 && self.nanos <= 0 || self.seconds <= 0 && self.nanos < 0 {
```
ehiggs, line 114: This seems like an unrelated fix that can ship on its own.

prost-build/src/code_generator.rs
```diff
@@ -1160,35 +1248,43 @@ fn build_enum_value_mappings<'a>(
     do_strip_enum_prefix: bool,
     enum_values: &'a [EnumValueDescriptorProto],
 ) -> Vec<EnumVariantMapping<'a>> {
-    let mut numbers = HashSet::new();
     let mut generated_names = HashMap::new();
-    let mut mappings = Vec::new();
+    // Use an insertion-order preserving map here because the enum ordering must be preserved.
+    let mut mappings = IndexMap::<i32, EnumVariantMapping>::new();
```
cynecx, line 1253: This could be simplified. This is really just a sequence of `iter().unique_by().map().collect()`.

cblichmann:
Any news on this? This would be incredibly useful.

cynecx:
I've done a quick rebase and made few adjustments to pass the conformance tests again. It also seems like I need to do some tweaks to fix the CI for `no-std`.

> Any news on this? This would be incredibly useful.

I am still way too busy with other things. What this PR needs is a split and a bit of slight documentation work. I will probably get to that at the end of December perhaps. Thanks for your patience!

prost-derive/Cargo.toml
```diff
@@ -14,7 +14,8 @@ proc-macro = true
 
 [dependencies]
 anyhow = "1.0.1"
-itertools = ">=0.10.1, <=0.13"
+heck = "0.4.1"
```
ehiggs, line 17: 0.5 was released 9 months ago. Any reason to use a version from 2 years ago?

prost-types/src/duration.rs
```diff
@@ -69,6 +69,10 @@ impl Duration {
         result.normalize();
         result
     }
+
+    pub fn is_valid(&self) -> bool {
+        self.seconds >= -315_576_000_000 && self.seconds <= 315_576_000_000
```
ehiggs, line 74: Sorry if I'm a tourist asking silly questions here, but what does it mean to be a duration consisting of a negative number of seconds?

prost-derive/Cargo.toml
cynecx: Not sure anymore but I guess this PR or at least my first local impl predated this commit https://github.com/tokio-rs/prost/commit/2b6140cf6fe899e9817816bae97d19fe8e0eab60.

But yeh, we can certainly bump this.

prost-types/src/duration.rs
cynecx: There was a conformance test that checked this behavior... It's basically an api contract / invariant, that is mentioned here: https://github.com/protocolbuffers/protobuf/blob/2ad21af8406316d2b170230f93db7d10c36108dd/src/google/protobuf/duration.proto#L103.

prost/src/serde/de.rs
```diff
@@ -0,0 +1,142 @@
...
+    fn deserialize<D>(deserializer: D, config: &DeserializerConfig) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let val = <T as CustomDeserialize>::deserialize(deserializer, config)?;
+        Ok(Box::new(val))
+    }
+}
+
+// FIXME: Make `T` contravariant, not covariant, by changing the `T` in `PhantomData` to
```
ehiggs, line 51: looks like this was done since you have `PhantomData<fn() -> T>`

ehiggs:
@caspermeijn has asked for this to be split into smaller PRs:
https://github.com/tokio-rs/prost/pull/1081#pullrequestreview-2115852148

Reviewing this, I would also expect to see example messages which get extended and validating that before and after changes to the proto file, it works.

e.g.

```proto
message TestMessage {
    string name = 1;
}
```

Then a json payload that satisfies that is then stuck into:

```proto
message TestMessage {
    string name = 1;
    int32 age = 2;
}
```

Or is there an existing test framework and this is using it as another pass with serde enabled and I just don't see how it fits together.

I'm not a committer so take my suggestion with a grain of salt. You've already done an impressive bit of work here!

cynecx:
> @caspermeijn has asked for this to be split into smaller PRs:

Yeah, I've mentioned this in my earlier update comment as well. It's definitely on my to-do list! :)

> Or is there an existing test framework and this is using it as another pass with serde enabled and I just don't see how it fits together.

I haven't written any specific tests for this yet. That said, I'm not overly concerned because the conformance test suite already exercises this functionality by feeding data (with unknown/missing fields) for decoding/deserialization during our roundtrip conversions. However, having some dedicated tests in our own codebase wouldn't hurt either.

For now, I'll likely prioritize other tasks in the near future. However, I'd love to collaborate on this, if you have any ideas for tests, feel free to share :D

cynecx:
The most recent commits should address almost all conformance failures. There is just one remaining, but that is a rather uninteresting one (`Recommended.Proto2.JsonInput.FieldNameExtension.Validator`). I've also added some more cleanups and lint fixes.

I think I am approaching a state where I can actually focus on the other TODOs mentioned in my other comments above.

PR #1088, opened 2024-06-25
build: Add features `edition-2015` and `std` to all tests crates

caspermeijn:
Only the crate `tests-2015` uses the feature `edition-2015` but the code is shared with the other tests crates variants.

Same for `tests-no-std` which disables feature `std` but reuses the code with the `tests` crate.

Therefore all tests crates should define all features, but not enable it.

Cargo clippy reports:
```
warning: unexpected `cfg` condition value: `edition-2015`
 --> tests-no-std/../tests/src/build.rs:5:14
  |
5 |     if #[cfg(feature = "edition-2015")] {
  |              ^^^^^^^-----------------
  |              |
  |              help: remove the condition
  |
  = note: no expected values for `feature`
  = help: consider adding `edition-2015` as a feature in `Cargo.toml`
  = note: see <https://doc.rust-lang.org/nightly/rustc/check-cfg/cargo-specifics.html> for more information about checking conditional configuration
  = note: `#[warn(unexpected_cfgs)]` on by default
```

### DIFF
--- a/tests-2018/Cargo.toml
+++ b/tests-2018/Cargo.toml
@@ -17,6 +17,9 @@ path = "../tests/src/lib.rs"
 [features]
 default = ["std"]
 std = []
+# This crate shares code with `tests-2015` crate. Define this feature, but 
+# never enable it for this crate.
+edition-2015 = []
 
 [dependencies]
 anyhow = "1.0.1"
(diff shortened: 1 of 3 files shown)


caspermeijn:
This is not the right solution for this problem, as CI will enable all features.

PR #1096, opened 2024-07-07
fix(prost-types): Fix date-time parsing

mumbleskates:
Two bugs revealed so far from fuzz testing in `bilrost`:

* `parse_two_digit_numeric` does not check the length of its input before calling `split_at`, potentially causing a panic
* a logic error in the translation of `year_to_seconds` from the original musl implementation caused it to treat the year 1900 as a leap year, causing (i believe) every date from 1900-01-01 to 1900-03-01 to be offset one day into the past when converting from `DateTime` to `Timestamp`.

### DIFF
--- a/prost-types/src/datetime.rs
+++ b/prost-types/src/datetime.rs
@@ -327,7 +327,9 @@ fn parse_offset(s: &str) -> Option<(i8, i8, &str)> {
 /// string.
 fn parse_two_digit_numeric(s: &str) -> Option<(u8, &str)> {
     debug_assert!(s.is_ascii());
-
+    if s.len() < 2 {
+        return None;
+    }
     let (digits, s) = s.split_at(2);
     Some((digits.parse().ok()?, s))
 }
@@ -420,8 +422,8 @@ pub(crate) fn year_to_seconds(year: i64) -> (i128, bool) {
     let is_leap;
     let year = year - 1900;
 
-    // Fast path for years 1900 - 2038.
-    if year as u64 <= 138 {
+    // Fast path for years 1901 - 2038.
+    if (1..=138).contains(&year) {
         let mut leaps: i64 = (year - 68) >> 2;
         if (year - 68).trailing_zeros() >= 2 {
             leaps -= 1;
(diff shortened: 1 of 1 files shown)


mumbleskates:
Here's another suggested fix for some weirdness that fuzzing failures highlighted: https://github.com/mumbleskates/bilrost/commit/2e43e023664d2127ccd9dc650476b7c397bfc077

prost-types/src/datetime.rs
```diff
@@ -420,8 +422,8 @@ pub(crate) fn year_to_seconds(year: i64) -> (i128, bool) {
     let is_leap;
     let year = year - 1900;
 
-    // Fast path for years 1900 - 2038.
-    if year as u64 <= 138 {
+    // Fast path for years 1901 - 2038.
+    if (1..=138).contains(&year) {
```
caspermeijn, line 426: So basically, you are saying that the fast path was incorrect for 1900, but the slow path is correct. I can't reason why that is, but I assume that you are right :-)

caspermeijn (approved):
Thanks for your contribution

caspermeijn:
> Here's another suggested fix for some weirdness that fuzzing failures highlighted: [mumbleskates/bilrost@2e43e02](https://github.com/mumbleskates/bilrost/commit/2e43e023664d2127ccd9dc650476b7c397bfc077)

Could you create a PR for that? Thanks.

prost-types/src/datetime.rs
mumbleskates: The [cited function in musl](https://git.musl-libc.org/cgit/musl/tree/src/time/__year_to_secs.c) actually writes this condition as `if (year-2ULL <= 136) {`... which isn't the same thing as `if year as u64 <= 138 {` at all! if year is either 0 _or_ 1 in the original function it wraps to max.

the musl code's fast-path works within i32 seconds of the epoch, which ranges from 1901-12-13 to 2038-01-18, so it actually skips the fast-path for 1901 as well because the start of that year is an underflow. for our code, we are calculating and returning an i128 anyway so it doesn't matter. (the code in this pr is still equally correct here if we write `if (1..=199).contains(&year) {` :) )

what DOES matter is that the fast-path ignores the century rule for leap years, because 2000 is a multiple of 400 and happens to be one! so the fast-path logic is actually WRONG when applied to any year outside 1901..=2099, even without integer wrapping.

caspermeijn: I understand how this could happen. That is a subtle, but important difference. Thanks for the extensive explanation!

PR #1104, opened 2024-07-12
fix(prost-types): '+' is not a numeric digit

mumbleskates:
`u8 as FromStr` tolerates a leading `+` character, so before this change for any 2-digit component of the path, "05" and "+5" are basically equivalent. we shouldn't tolerate this, it's weird.

### DIFF
--- a/prost-types/src/datetime.rs
+++ b/prost-types/src/datetime.rs
@@ -330,6 +330,9 @@ fn parse_two_digit_numeric(s: &str) -> Option<(u8, &str)> {
     if s.len() < 2 {
         return None;
     }
+    if s.starts_with('+') {
+        return None;
+    }
     let (digits, s) = s.split_at(2);
     Some((digits.parse().ok()?, s))
 }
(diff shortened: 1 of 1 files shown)


prost-types/src/datetime.rs
```diff
@@ -330,6 +330,9 @@ fn parse_two_digit_numeric(s: &str) -> Option<(u8, &str)> {
     if s.len() < 2 {
         return None;
     }
+    if s.starts_with('+') {
```
caspermeijn, line 333: The `digits.parse` function will eventually call [`u8::from_str_radix`]([from_str](https://doc.rust-lang.org/std/primitive.u8.html#method.from_str_radix)), which allows to an optional `+` and then only digits. So checking for the `+` as the first character is enough to get all the weird cases of `parse`.

caspermeijn (approved):
Good find. Thanks for your contribution!

prost-types/src/datetime.rs
mumbleskates: yep, that seems to be the case!

PR #1106, opened 2024-07-19
fix(prost-types): Parse timestamp with long second fraction

caspermeijn:
`chrono` is able to parse timestamps which are more precise then nanoseconds. We can also do that but truncating the digits.

### DIFF
--- a/prost-types/src/datetime.rs
+++ b/prost-types/src/datetime.rs
@@ -260,8 +260,10 @@ fn parse_nanos(s: &str) -> Option<(u32, &str)> {
 
     // Parse the nanoseconds, if present.
     let (nanos, s) = if let Some(s) = parse_char(s, b'.') {
-        let (digits, s) = parse_digits(s);
-        ensure!(digits.len() <= 9);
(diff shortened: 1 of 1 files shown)


caspermeijn:
@mumbleskates I found this one while fuzzing against chrono

mumbleskates:
nice! i haven't gotten around to looking at `Duration` yet but i was thinking about doing it this weekend, eager to take a look.

mumbleskates:
oh i see you're also parsing directly with chrono! i will try that also.

mumbleskates (approved):

mumbleskates:
Hmm. What fuzzing condition did you enforce with this? There are so many differences between how this library parses dates (esp. the year) and how chrono does it that it's hard to assert that we parse a superset. did you just check that strings that parse with both libraries are equivalent?

caspermeijn:
This is the fuzzing test I executed: https://github.com/caspermeijn/prost/commit/8dd7f4afb8126e398ae2fa0ada665f66bfe6843a 
I will PR that at a later time, as fuzzing is currently broken on master: #1107

PR #1110, opened 2024-07-26
fix(prost-types): Format negative fractional duration

caspermeijn:
Duration `-0.123s` was formatted as a positive number.


caspermeijn:
@mumbleskates I took a shot at fuzzing duration. It is difficult to find another implementation to compare to. But I found this bug at least.

mumbleskates:
oh nice! thank you.

yes, the standards for time intervals i know of are fearsomely complicated and I'd be hesitant to make anything much more complicated. maybe adding hours and minutes would be friendly but it's not really necessary

PR #1115, opened 2024-08-02
chore: Remove allow clippy::derive_partial_eq_without_eq

tottoto:
Removes allow `clippy::derive_partial_eq_without_eq` config as it has been changed to be allowed as default.

- https://github.com/rust-lang/rust-clippy/pull/9535
- https://rust-lang.github.io/rust-clippy/rust-1.70.0/index.html#derive_partial_eq_without_eq


caspermeijn (approved):

PR #1127, opened 2024-08-16
tests(fuzz): Add `parse_date` fuzzing target

caspermeijn:
Add a fuzzing target for parsing into a Timestamp and comparing the result with crate chrono.


caspermeijn:
@mumbleskates This is the fuzzer I used for my testing. Do you have suggestions?

mumbleskates (approved):
this looks good! when i tried to do it the first time i think i called the wrong chrono parsing function or something and it was accepting years with too few digits so i had a lot of annoying problems with that, but this is very nice.

PR #1140, opened 2024-08-30
feat(prost-build): emit `rerun` commands

caspermeijn:
Inform `cargo` about the files and env vars used by `prost-build`. Then `cargo` can better determine when to rebuild a project.

- Emit `rerun-if-changed` for each proto file specified
- Emit `rerun-if-changed` for each include directory specified
- Emit `rerun-if-changed` if `file_descriptor_set_path` is set
- Emit `rerun-if-env-changed` for `PROTOC` and `PROTOC_INCLUDE`

https://doc.rust-lang.org/cargo/reference/build-scripts.html#rerun-if-changed

BREAKING CHANGE: Previously `cargo` assumed it had to rerun `build.rs` if any files in the project changed. `prost-build` will now emit `rerun` commands, which means only the explicitly marked files cause a rerun. If your `build.rs` is dependent on any other file paths than those given to `prost-build`, then your `build.rs` needs to emit `rerun` commands as well.

### DIFF
--- a/prost-build/src/config.rs
+++ b/prost-build/src/config.rs
@@ -889,6 +889,7 @@ impl Config {
     ) -> Result<FileDescriptorSet> {
         let tmp;
         let file_descriptor_set_path = if let Some(path) = &self.file_descriptor_set_path {
+            println!("cargo:rerun-if-changed={}", path.display());
             path.clone()
         } else {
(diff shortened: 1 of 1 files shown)


armandas:
Would really appreciate this feature!

PR #1147, opened 2024-09-04
feat: Relax Message Debug trait bound

tottoto:
Relaxes `Message`'s `Debug` trait bound.

BREAKING CHANGE: `trait Debug` was a supertrait of `trait Message`. This is no longer required by `prost`. If your code relies on `trait Debug` being implemented for every `impl Message`, you must now explicitly state that you require both Debug and Message. For example: `where M: Debug + Message`

### DIFF
--- a/prost/src/message.rs
+++ b/prost/src/message.rs
@@ -3,8 +3,6 @@ use alloc::boxed::Box;
 #[cfg(not(feature = "std"))]
 use alloc::vec::Vec;
 
-use core::fmt::Debug;
-
 use bytes::{Buf, BufMut};
 
 use crate::encoding::varint::{encode_varint, encoded_len_varint};
@@ -14,7 +12,7 @@ use crate::DecodeError;
 use crate::EncodeError;
 
 /// A Protocol Buffers message.
-pub trait Message: Debug + Send + Sync {
+pub trait Message: Send + Sync {
     /// Encodes the message to a buffer.
     ///
     /// This method will panic if the buffer has insufficient capacity.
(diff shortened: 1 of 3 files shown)


caspermeijn (commented):
- Can you explain why this is useful?
- Is this an API breaking change?

tottoto:
> Can you explain why this is useful?

When the `Debug` implementation is not needed, we can use `prost_build::Config::skip_debug()` to skip implementing `Debug`. However as the `Messages` trait requires `Debug` implementation, it is needed to implement dummy implementation to implement `Message` when it is not actually needed. Relaxing `Message`'s Debug trait bound requirement removes this.

> Is this an API breaking change?

I think this is a breaking change. Previously, when a value's type had a `Message` bound imposed on it, we could write code that relied on that value implementing `Debug`, after this change, we will get a compilation error and will need to explicitly state `Debug`.

caspermeijn:
> > Can you explain why this is useful?
> 
> When the `Debug` implementation is not needed, we can use `prost_build::Config::skip_debug()` to skip implementing `Debug`. However as the `Messages` trait requires `Debug` implementation, it is needed to implement dummy implementation to implement `Message` when it is not actually needed. Relaxing `Message`'s Debug trait bound requirement removes this.

I see, so `skip_debug` was already useful to write your own `impl Debug`. With this change you can use `skip_debug` to not implement `Debug` at all. That makes sense.

> > Is this an API breaking change?
> 
> I think this is a breaking change. Previously, when a value's type had a `Message` bound imposed on it, we could write code that relied on that value implementing `Debug`, after this change, we will get a compilation error and will need to explicitly state `Debug`.

I understand. I will prepare this for the next breaking release

caspermeijn (approved):

caspermeijn:
I am preparing a breaking release now. I want to include this PR. Can you solve the merge conflict?

tottoto:
> I am preparing a breaking release now. I want to include this PR. Can you solve the merge conflict?

Resolved.

PR #1154, opened 2024-09-10
style(generated-code): use Self in as_str_name

thecaralice:
Makes `as_str_name` generate `match self` using `Self::` where possible, consistent with `from_str_name`


caspermeijn (changes requested):
Thank you for this styling change.

I would like to prevent regression via CI, but I can't think of a way to do it reliably. I would rather not enable `clippy::use_self` as it is in nursery.

Please change the commit message and title to better reflect the changes. For example: 
```
style(generated-code): Use `Self` in `as_str_name`
```

caspermeijn (approved):

PR #1160, opened 2024-09-17
feature(prost-build): Generate less boxed if nested type is boxed manually

ldm0:
Address https://github.com/tokio-rs/prost/pull/1157#discussion_r1760622415

This branch is based on #1157.


tests/src/nesting_complex.proto
```diff
@@ -0,0 +1,47 @@
+syntax = "proto2";
+
+package nesting_complex;
+
+// ----- Directly nested
```
caspermeijn, line 5: I would like to have more documentation explaining how this structure is nested.

caspermeijn (changes requested):
That is an interesting problem to fix. 
- Is `nested` the correct term to use? I would use recursive, but I am not proficient in graph theory.
- I would like user documentation about how this works.
- This is a breaking change, so it has to wait for the next breaking release.

caspermeijn:
@ldm0 I am preparing a breaking release. Are you still interested in this? If so, please update the code.

PR #1161, opened 2024-09-17
fix(protobuf): `protobuf/build.rs` get stuck when patch is already applied

ldm0:
`protobuf/build.rs` is occasionally stucked. The reason is that `patch` requires user input if the patch has already been applied:

```shell
$ patch -p1 -i "../../protobuf/src/fix-conformance_test_runner-cmake-build.patch"
patching file 'cmake/conformance.cmake'
Reversed (or previously applied) patch detected!  Assume -R? [y] 
```

Adding the `-f` flag solves the issue.

### DIFF
--- a/protobuf/build.rs
+++ b/protobuf/build.rs
@@ -97,6 +97,7 @@ fn apply_patches(src_dir: &Path) -> Result<()> {
     patch_src.push("fix-conformance_test_runner-cmake-build.patch");
 
     let rc = Command::new("patch")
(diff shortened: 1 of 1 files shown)


caspermeijn:
On my machine, the command doesn't get stuck on it when the command is run by `cargo`. But I do occasionally see that output when another build step fails. 

I want to update protobuf to a newer version that doesn't require the patch anymore. I expect that to land in a week.

PR #1162, opened 2024-09-18
Implement PartialOrd on Duration

wcarmon:
allow duration comparisons (similar to chrono::TimeDelta, which has the same structure)

https://docs.rs/chrono/latest/src/chrono/time_delta.rs.html#52

### DIFF
--- a/prost-types/src/protobuf.rs
+++ b/prost-types/src/protobuf.rs
@@ -1815,7 +1815,7 @@ pub struct Mixin {
 /// encoded in JSON format as "3s", while 3 seconds and 1 nanosecond should
 /// be expressed in JSON format as "3.000000001s", and 3 seconds and 1
 /// microsecond should be expressed in JSON format as "3.000001s".
-#[derive(Clone, Copy, PartialEq, ::prost::Message)]
+#[derive(Clone, Copy, PartialEq, PartialOrd, ::prost::Message)]
 pub struct Duration {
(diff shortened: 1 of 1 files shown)


caspermeijn (changes requested):
I like the idea.

1) `protobuf.rs` is a generated file. When the tests are run, it will be overwritten. This could be fixed by moving the derive to here: https://github.com/tokio-rs/prost/blob/8d4cac58ef7c5e33ba3279c836966c9109d4554c/tests/src/bootstrap.rs#L42
Then running the tests will generate a new version with your derive.
2) `Duration` is not guaranteed to be normalized. That means that `nanos` field could be larger than 1 second and the ordering would be off. I am unsure how big this problem is, as the documentation states that it should be within range. Let's ignore normalization for now.
3) I would like to see some tests to confirm the ordering works as expected. As I understand the documentation of `derive(Ord)` it sorts based on the alphabetical order, which would be incorrect in this case: https://dev-doc.rust-lang.org/stable/std/cmp/trait.PartialOrd.html#derivable
4) Are you aware of this earlier work? Maybe you could take that as inspiration: https://github.com/tokio-rs/prost/pull/995

caspermeijn:
@wcarmon I am preparing a breaking release. Are you still interested in this PR? If so, please update the code

wcarmon:
Thanks for the heads up.  I've been using `foo_milliseconds: i64`, so this PR is not as significant.

PR #1175, opened 2024-10-23
feat: derive Eq and Hash trait for messages where possible

caspermeijn:
Integer and bytes types can be compared using trait Eq. Some generated Rust structs can also have this property by deriving the Eq trait.

Automatically derive Eq and Hash for:
- messages that only have fields with integer or bytes types
- messages where all field types also implement Eq and Hash
- the Rust enum for one-of fields, where all fields implement Eq and Hash

Generated code for Protobuf enums already derives Eq and Hash.

BREAKING CHANGE: `prost-build` will automatically derive `trait Eq` and `trait Hash` for types where all field support those as well. If you manually `impl Eq` and/or `impl Hash` for generated types, then you need to remove the manual implementation. If you use `type_attribute` to `derive(Eq)` and/or `derive(Hash)`, then you need to remove those.


LucioFranco (approved):
LGTM, would it make sense to make these derives optional?

caspermeijn:
> LGTM, would it make sense to make these derives optional?

I think you mean something like `skip_debug` but for these traits. I feel like always implementing `Eq` is acceptable, as that is also the case for `PartialEq`. 

The disadvantage of a skip function is that the complexity is raised in the code generator.

LucioFranco:
Yeah, I was thinking it may help for those backward compat issues.

PR #1190, opened 2024-11-19
Clean up `CodeGenerator` by moving compilation-global data and logic to `Context`

mzabaluev:
Refactoring improvements to prost-build:

* Move compilation-global context data and logic around it from `CodeGenerator` to a new `Context` struct.
* Lift similar logic into `Context` from `MessageGraph`, where it caused some data duplication.
* Resolve some awkwardness about the `Config` reference needing to be mutable for its user-provided service generator, by exposing both from `Context` in a more disciplined way.

This also fixes a bug in the logic that determines whether `Copy` can be derived on a message, where a `boxed` config rule matching a field that is part of a oneof would be disregarded.

### DIFF
--- a/prost-build/src/code_generator.rs
+++ b/prost-build/src/code_generator.rs
@@ -80,7 +80,7 @@ impl OneofField {
 }
 
 impl<'a, 'b> CodeGenerator<'a, 'b> {
-    pub fn generate(context: &mut Context<'b>, file: FileDescriptorProto, buf: &mut String) {
+    pub(crate) fn generate(context: &mut Context<'b>, file: FileDescriptorProto, buf: &mut String) {
         let source_info = file.source_code_info.map(|mut s| {
             s.location.retain(|loc| {
                 let len = loc.path.len();
--- a/prost-build/src/code_generator.rs
+++ b/prost-build/src/code_generator.rs
@@ -15,10 +15,8 @@ use prost_types::{
 };
 
 use crate::ast::{Comments, Method, Service};
-use crate::extern_paths::ExternPaths;
+use crate::context::Context;
 use crate::ident::{strip_enum_prefix, to_snake, to_upper_camel};
-use crate::message_graph::MessageGraph;
-use crate::Config;
 
 mod c_escaping;
 use c_escaping::unescape_c_escape_string;
@@ -26,14 +24,12 @@ use c_escaping::unescape_c_escape_string;
 mod syntax;
 use syntax::Syntax;
 
-pub struct CodeGenerator<'a> {
-    config: &'a mut Config,
+pub struct CodeGenerator<'a, 'b> {
+    context: &'a mut Context<'b>,
     package: String,
     type_path: Vec<String>,
     source_info: Option<SourceCodeInfo>,
     syntax: Syntax,
-    message_graph: &'a MessageGraph,
-    extern_paths: &'a ExternPaths,
     depth: u8,
     path: Vec<i32>,
     buf: &'a mut String,
@@ -43,10 +39,6 @@ fn push_indent(buf: &mut String, depth: u8) {
     for _ in 0..depth {
         buf.push_str("    ");
     }
-}
-
-fn prost_path(config: &Config) -> &str {
-    config.prost_path.as_deref().unwrap_or("::prost")
 }
 
 struct Field {
@@ -87,14 +79,8 @@ impl OneofField {
     }
 }
 
-impl CodeGenerator<'_> {
-    pub fn generate(
-        config: &mut Config,
-        message_graph: &MessageGraph,
-        extern_paths: &ExternPaths,
-        file: FileDescriptorProto,
-        buf: &mut String,
-    ) {
+impl<'a, 'b> CodeGenerator<'a, 'b> {
+    pub fn generate(context: &mut Context<'b>, file: FileDescriptorProto, buf: &mut String) {
         let source_info = file.source_code_info.map(|mut s| {
             s.location.retain(|loc| {
                 let len = loc.path.len();
@@ -105,13 +91,11 @@ impl CodeGenerator<'_> {
         });
 
         let mut code_gen = CodeGenerator {
-            config,
+            context,
             package: file.package.unwrap_or_default(),
             type_path: Vec::new(),
             source_info,
             syntax: file.syntax.as_deref().into(),
-            message_graph,
-            extern_paths,
             depth: 0,
             path: Vec::new(),
             buf,
@@ -139,7 +123,7 @@ impl CodeGenerator<'_> {
         }
         code_gen.path.pop();
 
-        if code_gen.config.service_generator.is_some() {
+        if code_gen.context.service_generator().is_some() {
             code_gen.path.push(6);
             for (idx, service) in file.service.into_iter().enumerate() {
                 code_gen.path.push(idx as i32);
@@ -147,7 +131,7 @@ impl CodeGenerator<'_> {
                 code_gen.path.pop();
             }
 
-            if let Some(service_generator) = code_gen.config.service_generator.as_mut() {
+            if let Some(service_generator) = code_gen.context.service_generator() {
                 service_generator.finalize(code_gen.buf);
             }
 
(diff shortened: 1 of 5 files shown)


mzabaluev:
I would like this in, before I add an enum type registry, which is necessary to properly implement imports (and later with editions, the `enum_type` feature) for #1079.

prost-build/src/code_generator.rs
```diff
@@ -87,14 +79,8 @@ impl OneofField {
...
-impl CodeGenerator<'_> {
-    pub fn generate(
-        config: &mut Config,
-        message_graph: &MessageGraph,
-        extern_paths: &ExternPaths,
-        file: FileDescriptorProto,
-        buf: &mut String,
-    ) {
+impl<'a, 'b> CodeGenerator<'a, 'b> {
+    pub fn generate(context: &mut Context<'b>, file: FileDescriptorProto, buf: &mut String) {
```
caspermeijn, line 83: This is an API breaking change, right? I don't understand how the `Context` instance should be created by `prost` users as there is no documentation changes included.

prost-build/src/code_generator.rs
```diff
@@ -15,25 +15,21 @@ use prost_types::{
...
 mod c_escaping;
 use c_escaping::unescape_c_escape_string;
 
 mod syntax;
 use syntax::Syntax;
 
-pub struct CodeGenerator<'a> {
-    config: &'a mut Config,
+pub struct CodeGenerator<'a, 'b> {
+    context: &'a mut Context<'b>,
```
caspermeijn, line 28: I don't understand why some fields are in `CodeGenerator`, `Context` or `Config`.  For example, `depth` also feels like it could be part of `Context`.

Can you explain when a field should be in each of those structs?

prost-build/src/context.rs
```diff
@@ -0,0 +1,285 @@
...
+
+use crate::extern_paths::ExternPaths;
+use crate::message_graph::MessageGraph;
+use crate::{BytesType, Config, MapType, ServiceGenerator};
+
+/// The context providing all the global information needed to generate code.
+/// It also provides a more disciplined access to Config
+/// and its mutable instance of ServiceGenerator.
+pub struct Context<'a> {
+    config: &'a mut Config,
```
caspermeijn, line 16: In the PR description, you mention that a mutable Config is awkward. But it is still `mut` can you further explain what has been improved?

caspermeijn (commented):
I see plenty of internal changes that I like: better naming, cleanup of boxed, more helper functions. 
- Is it possible to separate those from the API changes? Then we can that merge that before doing a version number bump
- Is it possible to separate each improvement into a separate commit? That will make reviewing easier.

prost-build/src/context.rs
```diff
@@ -0,0 +1,285 @@
...
+            message_graph,
+            extern_paths,
+        }
+    }
+
+    pub fn config(&self) -> &Config {
+        self.config
+    }
+
+    pub fn service_generator(&mut self) -> Option<&mut (dyn ServiceGenerator + 'static)> {
```
mzabaluev, line 38: ... but it can get a reference to the mutable object providing a `ServiceGenerator` vtable. Which was the only reason for passing a mutable reference to `Config` to `CodeGenerator` before, even though it does not need to (and really should not) mutate any other members of `Config`.

prost-build/src/context.rs
```diff
@@ -0,0 +1,285 @@
...
+            if self
+                .config
+                .boxed
+                .get_first_field(&config_path, field.name())
+                .is_some()
+            {
+                false
+            } else {
+                self.can_message_derive_copy(field.type_name())
+            }
```
caspermeijn, line 235: This logic has changed compared to `MessageGraph::can_field_derive_copy`. I don't expect that in a PR that mentions just to move code.

prost-build/src/context.rs
```diff
@@ -0,0 +1,285 @@
...
+    pub fn should_disable_comments(&self, fq_message_name: &str, field_name: Option<&str>) -> bool {
+        if let Some(field_name) = field_name {
+            self.config
+                .disable_comments
+                .get_first_field(fq_message_name, field_name)
+                .is_some()
+        } else {
+            self.config
+                .disable_comments
+                .get_first(fq_message_name)
```
caspermeijn, line 266: `PathMap::get_first` is subtly different from `PathMap::get` and then calling `.next()`. Is that intentional?

caspermeijn (changes requested):
Not every line is improved, but I think this is an improvement overall. I have a few questions that I like to be answered before merging.

prost-build/src/code_generator.rs
```diff
@@ -297,15 +285,16 @@ impl CodeGenerator<'_> {
             self.pop_mod();
         }
 
-        if self.config.enable_type_names {
+        if self.context.config().enable_type_names {
```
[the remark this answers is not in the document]
mzabaluev: I can add a `.config()` helper method to `CodeGenerator`.

prost-build/src/code_generator.rs
```diff
@@ -616,13 +587,16 @@ impl CodeGenerator<'_> {
         self.push_indent();
 
         let can_oneof_derive_copy = oneof.fields.iter().all(|field| {
-            self.message_graph
-                .can_field_derive_copy(fq_message_name, &field.descriptor)
+            self.context.can_oneof_field_derive_copy(
+                fq_message_name,
+                oneof.descriptor.name(),
```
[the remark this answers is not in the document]
mzabaluev: See the discussion on the `Context::can_field_derive_copy_impl` method.

prost-build/src/context.rs
mzabaluev: No. Would it behave differently in any cases?

prost-build/src/context.rs
mzabaluev: I will add a regression test to prove this implements correct behavior.

mzabaluev: For comparison, check the logic in [`Context::should_box_oneof_field`](https://github.com/mzabaluev/prost/blob/6bcdc18e013c1cce16cb804c75285e2654ef10b9/prost-build/src/context.rs#L129) which I did not modify from how it works in the baseline.

mzabaluev: However, this behavior for fields in a oneof is undocumented, and logically, all named fields in a oneof are still part of the enclosing message's field namespace. So the behavior matching for `boxed` field paths in `can_message_derive_copy` may in fact be a more logical one.

prost-build/src/code_generator.rs
caspermeijn: Yes `pub(crate)` is a great idea

prost-build/src/context.rs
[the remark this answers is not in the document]
caspermeijn: I was thinking this is the only place where a `mut` reference is taken (as is the stated intention of the PR). So marking that spot makes it easier to find in a month time.

prost-build/src/context.rs
caspermeijn: I agree there could be a problem here and a regression test is useful. However, this PR should not change any behavior. It should only reorganize the code. Please separate the changed behavior into a separate PR so that we can discuss that in isolation.

My primary concern is that this PR should not introduce a breaking change accidentally. Because it is a lot of code changes, it is difficult to review. If additional changes are slipped in, that makes it even more difficult to review.

prost-build/src/context.rs
caspermeijn: I have not dug into that. I just see that the code does something different 😉

I suggest using the old code and opening another (smaller) PR for adding tests to prove noting changed.

prost-build/src/context.rs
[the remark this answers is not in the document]
mzabaluev: https://github.com/tokio-rs/prost/pull/1190/commits/01854ba4a5cb10c2ec65ba185c490cb47369de08

prost-build/src/context.rs
mzabaluev: https://github.com/tokio-rs/prost/pull/1190/commits/146871e934f193d4f2c93ea5ebc43fdb84d79953

prost-build/src/context.rs
mzabaluev: https://github.com/tokio-rs/prost/pull/1190/commits/596574e4712d1fcea64dd682cc52eb135d340a06

caspermeijn:

PR #1199, opened 2024-11-29
Make `key_len` and `encoding_len_varint` const

mzabaluev:
The constness is not really needed anywhere in prost crates, but it won't hurt either.


caspermeijn:

PR #1202, opened 2024-12-10
Add support for marking fields as Cow

arilou:
Add support for marking fields as Cow thus avoiding the need to own a value, this can be done on Bytes and String types only.


caspermeijn (changes requested):
I understand the need for `Cow` in fields. However, I think this implementation introduces more variants of the code generator and that reduces maintainability.

I am open to redoing the `encoding` module to make support for `Cow` easier. But I don't have the time to do that. I do have ideas about how to do that, so feel free to talk to me about that.

This PR is missing tests. If you can extensively prove the correct behavior (especially in combination with other `Config` options), I will be able to accept a feature like this.

arilou:
Hi @caspermeijn,

Thank you for the feedback. I’ll focus on adding the necessary tests to this PR, particularly to verify its behavior with other Config options.

I’m also interested in your ideas for implementing Cow support more effectively. If you can share any specifics, it would help guide the implementation.

Let me know if there’s anything else you’d like addressed.

caspermeijn:
> I’m also interested in your ideas for implementing Cow support more effectively. If you can share any specifics, it would help guide the implementation.

That would be a refactor of the `encoding` module of `prost`. If you want to discuss that work, you should contact me on Discord.

PR #1204, opened 2024-12-12
let Message inherit AsAny

tinglou:
introduce crate as `as_any`, and let `Message` inherits `AsAny`, then `dyn Message` can downcast to a concrete type easier. 
Here is a example,
```
    fn encode(&self, msg: &dyn Message, buf: &mut Vec<u8>) -> Result<(), TcpTransportError> {
        let msg = msg
            .downcast_ref::<T>()
            .ok_or(TcpTransportError::Infallible)?;

        // println!("type name: {}", msg.type_name());
        msg.encode(buf)?;
        Ok(())
    }
```

### DIFF
--- a/prost/Cargo.toml
+++ b/prost/Cargo.toml
@@ -24,6 +24,7 @@ std = []
 
 [dependencies]
 bytes = { version = "1", default-features = false }
+as-any = "0.3"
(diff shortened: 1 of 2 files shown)


caspermeijn:
Closing this for now. Please don't hesitate to re-open if you can answer the earlier question.

PR #1208, opened 2024-12-20
[decode_length_delimiter] document behavior for error with 10 bytes buffer

NilsIrl:


caspermeijn (approved):

prost/src/encoding/length_delimiter.rs
```diff
@@ -40,7 +40,7 @@ pub fn length_delimiter_len(length: usize) -> usize {
 ///
 ///  * If the supplied buffer contains fewer than 10 bytes, then an error indicates that more
 ///    input is required to decode the full delimiter.
-///  * If the supplied buffer contains more than 10 bytes, then the buffer contains an invalid
+///  * If the supplied buffer contains 10 bytes or more, then the buffer contains an invalid
```
mumbleskates, line 43: This change is unfortunately wrong. Protobuf varints can be ten bytes in length; looking at the varint decoding functions proves that the implementation also works this way.

mumbleskates: The only valid varint encodings that are ten bytes long all have a typical final byte with the value `0x01` (or `0x00` if they are over-length encoded). That least-significant bit in the final byte is the 64th and final significant bit that can be set (each of the 9 other bytes has 7 significant bits and 1 continuation bit in the most-significant position, for a total of 63 other significant bits.

Any other value for that byte will also return errors about invalid varints, which might be what you are seeing if you got here by discovering that ten bytes don't decode successfully.

PR #1210, opened 2024-12-20
fix: boxed oneof field

### LINKED ISSUE #1159: `prost_config::Config::boxed` generates code doesn't compile
**Version**
`prost@0.13.2`, `prost-build@0.13.2`

**Platform**
The output of `uname -a` (UNIX), or version and 32 or 64-bit (Windows)

**Description**
`prost_config::Config::boxed` generates code doesn't compile. I tried this code:

`boxed_tests.proto`:
```protobuf
syntax = "proto3";

package boxed_tests;

message Container {
  oneof data {
    string baz = 5;
    string boxed_baz = 6;
  }
}

```

`build.rs`:
```rust
fn main() {
    prost_build::Config::new()
        .out_dir(".")
        .boxed("Container.data.boxed_baz")
        .compile_protos(&["boxed_tests.proto"], &[""])
        .unwrap();
}
```

`main.rs`
```rust
mod boxed_tests {
    include!("../boxed_tests.rs");
}

fn main() {}
```

I expected to see this happen: compiled successfully

Instead, this happened: compilation failed.
error message:
```
error[E0308]: mismatched types
  --> src/../boxed_tests.rs:9:32
   |
9  |       #[derive(Clone, PartialEq, ::prost::Oneof)]
   |                                  ^^^^^^^^^^^^^^ expected `Box<String>`, found `String`
10 |       pub enum Data {
   |  ______________-
11 | |         #[prost(string, tag = "5")]
12 | |         Baz(::prost::alloc::string::String),
13 | |         #[prost(string, tag = "6")]
14 | |         BoxedBaz(::prost::alloc::boxed::Box<::prost::alloc::string::String>),
   | |________________- arguments to this enum variant are incorrect
   |
   = note: expected struct `Box<String>`
              found struct `String`
   = note: for more on the distinction between the stack and the heap, read https://doc.rust-lang.org/book/ch15-01-box.html, https://doc.rust-lang.org/rust-by-example/std/box.html, and https://doc.rust-lang.org/std/boxed/index.html
note: tuple variant defined here
  --> src/../boxed_tests.rs:14:9
   |
14 |         BoxedBaz(::prost::alloc::boxed::Box<::prost::alloc::string::String>),
   |         ^^^^^^^^
   = note: this error originates in the derive macro `::prost::Oneof` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0308`.
```

xxchan:
fix #1159

Signed-off-by: xxchan <xxchan22f@gmail.com>

### DIFF
--- a/prost-derive/src/field/oneof.rs
+++ b/prost-derive/src/field/oneof.rs
@@ -3,7 +3,7 @@ use proc_macro2::TokenStream;
 use quote::quote;
 use syn::{parse_str, Expr, ExprLit, Ident, Lit, Meta, MetaNameValue, Path};
 
-use crate::field::{set_option, tags_attr};
+use crate::field::{set_bool, set_option, tags_attr, word_attr};
 
 #[derive(Clone)]
 pub struct Field {
@@ -15,6 +15,8 @@ impl Field {
     pub fn new(attrs: &[Meta]) -> Result<Option<Field>, Error> {
         let mut ty = None;
         let mut tags = None;
+        let mut boxed = false;
+
         let mut unknown_attrs = Vec::new();
 
         for attr in attrs {
@@ -34,6 +36,8 @@ impl Field {
                 set_option(&mut ty, t, "duplicate oneof attribute")?;
             } else if let Some(t) = tags_attr(attr)? {
                 set_option(&mut tags, t, "duplicate tags attributes")?;
+            } else if word_attr("boxed", attr) {
+                set_bool(&mut boxed, "duplicate boxed attribute")?;
             } else {
                 unknown_attrs.push(attr);
             }
--- a/tests/src/boxed_field.rs
+++ b/tests/src/boxed_field.rs
@@ -1,4 +1,5 @@
 include!(concat!(env!("OUT_DIR"), "/boxed_field.rs"));
+use foo::OneofField;
 
 #[test]
 /// Confirm `Foo::bar` is boxed by creating an instance
@@ -6,5 +7,6 @@ fn test_bar_is_boxed() {
     use alloc::boxed::Box;
     let _ = Foo {
         bar: Some(Box::new(Bar {})),
+        oneof_field: Some(OneofField::BoxQux(Box::new(Bar {}))),
     };
 }
--- a/prost-build/src/code_generator.rs
+++ b/prost-build/src/code_generator.rs
@@ -639,21 +639,21 @@ impl CodeGenerator<'_> {
 
             self.push_indent();
             let ty_tag = self.field_type_tag(&field.descriptor);
-            self.buf.push_str(&format!(
-                "#[prost({}, tag=\"{}\")]\n",
-                ty_tag,
-                field.descriptor.number()
-            ));
-            self.append_field_attributes(&oneof_name, field.descriptor.name());
-
-            self.push_indent();
-            let ty = self.resolve_type(&field.descriptor, fq_message_name);
-
             let boxed = self.boxed(
                 &field.descriptor,
                 fq_message_name,
                 Some(oneof.descriptor.name()),
             );
+            self.buf.push_str(&format!("#[prost({}", ty_tag));
+            if boxed {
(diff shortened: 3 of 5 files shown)


tests/src/boxed_field.rs
```diff
@@ -1,10 +1,12 @@
 include!(concat!(env!("OUT_DIR"), "/boxed_field.rs"));
+use foo::OneofField;
 
 #[test]
 /// Confirm `Foo::bar` is boxed by creating an instance
 fn test_bar_is_boxed() {
```
caspermeijn, line 6: This comment and name need to be updated as well

prost-derive/src/field/oneof.rs
```diff
@@ -15,6 +15,8 @@ impl Field {
     pub fn new(attrs: &[Meta]) -> Result<Option<Field>, Error> {
         let mut ty = None;
         let mut tags = None;
+        let mut boxed = false;
```
caspermeijn, line 18: I don't see where this variable is used. Can you explain what this does?

caspermeijn (changes requested):
I think this test will pass without the other changes. `CodeGenerator` already makes the OneOf field a `Box<T>` if `self.boxed() == true` on line 667.

Can you explain why the other changes are needed?

xxchan (commented):
Hi @caspermeijn, thanks for the prompt review! Reverted unnecessary changes, and made CI Green.

> I think this test will pass without the other changes. 

I'm not sure what you mean by "other changes" here.

But anyway, here's how it works:

As mentioned in #1159, the type for the oneof field was correct. What was wrong is the generated `merge` function.

I noticed there are `#[prost(boxed)]` on normal messages fields, but not on the `oneof` field. After adding it, it works.

```rust
pub struct Foo {
    #[prost(message, optional, boxed, tag = "1")]
    pub bar: ::core::option::Option<::prost::alloc::boxed::Box<Bar>>,
    #[prost(oneof = "foo::OneofField", tags = "2, 3")]
    pub oneof_field: ::core::option::Option<foo::OneofField>,
}
/// Nested message and enum types in `Foo`.
    pub mod foo {
        pub enum OneofField {
            #[prost(string, tag = "2")]
            Baz(::prost::alloc::string::String),
// This PR added "boxed" here 
            #[prost(message, boxed, tag = "3")]  
            BoxQux(::prost::alloc::boxed::Box<super::Bar>),
        }
	}
```
2. after this change,

prost-derive/src/field/oneof.rs
caspermeijn: Thanks for clarifying that

tests/src/boxed_field.rs
```diff
@@ -1,10 +1,19 @@
 include!(concat!(env!("OUT_DIR"), "/boxed_field.rs"));
 
+cfg_if! {
+    if #[cfg(feature = "edition-2015")] {
+        use boxed_field::foo::OneofField;
+    } else {
+        use foo::OneofField;
+    }
+}
+
```
caspermeijn, line 10: You could simplify this to:
```suggestion
use self::foo::OneofField;
```

xxchan:
Yes you are right. I found the problem is only for `string` oneof variants, but not `message` variant. 

`::prost::encoding::message::merge` accepts `M: Message`, and we have `impl<M> Message for Box<M>`. But `::prost::encoding::string::merge` only accepts `&mut String`. 🤔

xxchan:
But it seems not useful at all to `box` a `string`.

caspermeijn:
> But it seems not useful at all to `box` a `string`.

I agree, a `Box<String>` doesn't make sense. Maybe the code generator should prevent a string from being boxed. 

Perhaps it should just ignore `boxed` when the type is string. If it currently doesn't compile, then it is not a breaking change to change the field type.

To improve the tests, you could try to reuse `TestAllTypesProto3` and add `.boxed(".")` to the `build.rs`. That way, we can verify that all field types actually compile.

PR #1212, opened 2024-12-27
Add a config option to avoid passing include_source_info

### LINKED ISSUE #1213: Make include_source_info protoc flag configurable
From the ``protoc --help``:

```
  --include_source_info       When using --descriptor_set_out, do not strip
                              SourceCodeInfo from the FileDescriptorProto.
                              This results in vastly larger descriptors that
                              include information about the original
                              location of each decl in the source file as
                              well as surrounding comments.
```

We already support passing disable_comments to disable comment processing during the code generation.

However, when we compile FileDescriptorSet, we always pass --include_source_info. These might cause a problem, especially when the file is later included into the final binary using include_bytes! e.g. because of the following:
- The comments might contain sensitive private information (comments + file location) which we don't want to be visible in the binary (e.g. when someone disassemble it).
- It increases the size of the binary

kgrech:
Fixes #1213

We already support passing ``disable_comments`` to disable comment processing during the code generation.

However, when we compile FileDescriptorSet, we always pass ``--include_source_info``. These might cause a problem, especially when the file is later included into the final binary using ``include_bytes!`` e.g. because of the following:
- The comments might contain sensitive private information (comments + file location) which we don't want to be visible in the binary (e.g. when  someone disassemble it).
- It increases the size of the binary

Therefore, adding the ``skip_source_info`` argument allowing to disable it.  Altnerative fix could be to add it ``protoc_args`` by default, but this is not backwards compatible, because ``protoc_args`` are currently additive only.

### DIFF
--- a/prost-build/src/config.rs
+++ b/prost-build/src/config.rs
@@ -614,7 +614,6 @@ impl Config {
     ///     .skip_source_info()
     ///     .compile_protos(&["src/items.proto"], &["src/"]);
     /// ```
-    ///
     pub fn skip_source_info(&mut self) -> &mut Self {
         self.skip_source_info = true;
         self
--- a/prost-build/src/config.rs
+++ b/prost-build/src/config.rs
@@ -48,6 +48,7 @@ pub struct Config {
     pub(crate) disable_comments: PathMap<()>,
     pub(crate) skip_debug: PathMap<()>,
     pub(crate) skip_protoc_run: bool,
+    pub(crate) skip_source_info: bool,
     pub(crate) include_file: Option<PathBuf>,
     pub(crate) prost_path: Option<String>,
     #[cfg(feature = "format")]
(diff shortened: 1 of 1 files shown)


LucioFranco (approved):

prost-build/src/config.rs
```diff
@@ -597,6 +598,26 @@ impl Config {
         self
     }
 
+    /// This can be used to avoid passing ``--include_source_info`` argument when the `protoc`
+    /// command is invoked to generate the `FileDescriptorSet`. The ``--include_source_info``
+    /// flag is passed by default, however it results in vastly larger descriptors that
+    /// include information about the original location of each decl in the source file as
+    /// well as surrounding comments.
```
davidbarsky, line 605:
```suggestion
    /// Configures the code generator to remove surrounding comments and documentation.
    /// 
    /// If enabled, this will cause `protoc` to not be passed the `--include_source_info` argument.
    /// Typically, `--include_source_info` is passed by default, but it results in larger
    /// [`FileDescriptorSet`s](https://github.com/protocolbuffers/protobuf/blob/cff254d32f850ba8186227ce6775b3f01a1f8cf8/src/google/protobuf/descriptor.proto#L54-L66) that include information about the
    /// original location of each declaration in the source file as well as surrounding
    /// comments and documentation.
```

prost-build/src/config.rs
```diff
@@ -597,6 +598,28 @@ impl Config {
...
+    ///
+    /// In `build.rs`:
+    ///
+    /// ```rust
+    /// # let mut config = prost_build::Config::new();
+    /// config.file_descriptor_set_path("path/from/build/system")
+    ///     .skip_source_info()
+    ///     .compile_protos(&["src/items.proto"], &["src/"]);
+    /// ```
+    ///
```
davidbarsky, line 617:
```suggestion
```

PR #1215, opened 2024-12-30
Fix link in field_attribute doc

ayymart:


caspermeijn:
I took the liberty to replace all those HTML links with intra doc links. Can you verify this works for you? https://github.com/tokio-rs/prost/pull/1219

caspermeijn:
Fixed in #1219 

PR #1217, opened 2025-01-02
Add more no-default-features = false to prost-derive

kpp:
itertools and anyhow bring unnecessary std into prost-derive.

### DIFF
--- a/prost-derive/Cargo.toml
+++ b/prost-derive/Cargo.toml
@@ -13,8 +13,8 @@ rust-version.workspace = true
 proc-macro = true
 
 [dependencies]
-anyhow = "1.0.1"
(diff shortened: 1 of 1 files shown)


caspermeijn (changes requested):
Apparently they are needed, as CI is red :-)

`prost-derive` is run on the build machine. Why is `no-std` needed in that case?

kpp:
Because it turns on std feature in anyhow

caspermeijn:
> Because it turns on std feature in anyhow

But that is the anyhow that is a build-dependency, right? As I understand, the anyhow that is built for the no-std target is a different build.

caspermeijn:
Closing due to no response. Please reopen if you can explain why this is needed.

PR #1228, opened 2025-01-14
fix(prost-build): Make `type_name_domain` cumulative

willtrnr:
Changes the `type_name_domain` config to behave closer to the way `extern_path` does.

This allows setting more than 1 path to domain mappings.

The original behavior of clearing previous mappings didn't make a lot of sense to me, but perhaps it was working as intended. If that's the case, then this would be a feature instead of a fix I guess.

BREAKING CHANGE: The configuration for domain names of messages is now cumulative. All calls to prost_build::Config::type_name_domain are now concatenated. The previous behavior was that only the arguments of the last call were used. If you do multiple calls to type_name_domain, you need to remove all but the last call to maintain the same behavior.

### DIFF
--- a/prost-build/src/config.rs
+++ b/prost-build/src/config.rs
@@ -684,7 +684,6 @@ impl Config {
         S: AsRef<str>,
         D: AsRef<str>,
     {
-        self.type_name_domains.clear();
         for matcher in paths {
             self.type_name_domains
                 .insert(matcher.as_ref().to_string(), domain.as_ref().to_string());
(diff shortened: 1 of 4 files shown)


caspermeijn (approved):
I agree, this makes more sense. Thanks for your contribution.

This is a breaking change, so it has to wait for the next major release.

> BREAKING CHANGE: The configuration for domain names of messages is now cumulative. All calls to `prost_build::Config::type_name_domain` are now concatinated. The previous behaviour is that only the arguments of the last call was used. If you do multiple calls to `type_name_domain`, you need to remove all but the last call to maintain the same behaviour.

PR #1230, opened 2025-01-16
Add a 'foreign' interface to support scalars defined outside of prost

bickfordb:
Example

```rust
use prost_derive::Message;
use proto::account;
use sea_orm::entity::prelude::*;
use serde::{Deserialize, Serialize};
use uuid::Uuid;

pub mod codec {
    pub mod uuid {
        pub mod simple {
            use prost::encoding::string;
            use uuid::Uuid;
            pub fn default() -> Uuid {
                Uuid::nil()
            }
            pub fn encode(tag: u32, value: &Uuid, buf: &mut impl ::prost::bytes::BufMut) {
                let s = value.to_string();
                string::encode(tag, &s, buf)
            }

            pub fn encoded_len(tag: u32, value: &Uuid) -> usize {
                let s = value.to_string();
                string::encoded_len(tag, &s)
            }
            pub fn clear(value: &mut Uuid) {
                *value = Uuid::nil()
            }
            pub fn merge(
                wire_type: ::prost::encoding::WireType,
                value: &mut Uuid,
                buf: &mut impl ::prost::bytes::Buf,
                ctx: ::prost::encoding::DecodeContext,
            ) -> Result<(), prost::DecodeError> {
                let mut s: String = "".to_string();
                string::merge(wire_type, &mut s, buf, ctx)?;
                *value = Uuid::parse_str(&s)
                    .map_err(|_| prost::DecodeError::new("failed to decode uuid string"))?;
                Ok(())
            }
        }
    }
}

#[derive(Message, Clone, Debug, PartialEq, Eq, DeriveEntityModel, Deserialize, Serialize)]
#[sea_orm(table_name = "account")]
pub struct Model {
    #[sea_orm(primary_key, auto_increment = false)]
    #[prost(codec::uuid::simple, tag = 1)]
    pub id: Uuid,
    #[sea_orm(column_type = "Text")]
    #[prost(string, tag=2)]
    pub name: String,
}
```

### DIFF
--- a/prost-derive/src/field/scalar.rs
+++ b/prost-derive/src/field/scalar.rs
@@ -40,7 +40,6 @@ impl Field {
                 unknown_attrs.push(attr);
             }
         }
-
         let ty = match ty {
             Some(ty) => ty,
             None => return Ok(None),
@@ -112,7 +111,10 @@ impl Field {
             Kind::Repeated => quote!(encode_repeated),
             Kind::Packed => quote!(encode_packed),
         };
-        let encode_fn = quote!(::prost::encoding::#module::#encode_fn);
+        let encode_fn = match self.ty {
+            Ty::Foreign(ref path) => quote!(#path::#encode_fn),
+            _ => quote!(::prost::encoding::#module::#encode_fn),
+        };
         let tag = self.tag;
 
         match self.kind {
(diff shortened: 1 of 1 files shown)


caspermeijn:
I think this is a very interesting idea. I had always thought a trait would be required to make something like this work. Prost had an `encoding` module, the is doc hidden. It implements the encoding for all the types, and you want to reuse that structure for foreign types.

I think we should first clean up and document the `encoding` module. After that, it would be possible to allow foreign types.

caspermeijn:
One drawback of the modules is the lack of generics. Therefore, you can't make one encoding module that works for every boxed type.

Sculas:
Also, since the foreign type needs to start with `::`, how do you handle cases where a codec is at the root of the crate? (e.g. `crate::codec::uuid`) This won't work since `::crate` tries to look for the crate named `crate`. It should be fairly simple to fix, just trim the leading colon if the first path element is `crate` (no actual crate with that name exists, and it's probably reserved anyway).

PR #1237, opened 2025-02-07
fix(prost-build): Prevent repeated fields to be boxed

caspermeijn:
Repeated fields are stored in a `Vec`, and therefore they are already heap allocated.

BREAKING CHANGE: A repeated field that is manually marked as boxed was typed as `Vec<Box<T>>`. Those fields are now simply typed as `Vec<T>` to prevent double indirection. The `boxed` configuration is effectively ignored for repeated fields.


LucioFranco (approved):
LGTM I would maybe write a test as well?

caspermeijn:
> LGTM I would maybe write a test as well?

Good idea, added it.

PR #1247, opened 2025-02-24
feat: Remove prost-derive feature

tottoto:
Removes the deprecated prost-derive feature.

BREAKING CHANGE: Feature flag `prost-derive` is renamed to `derive`. Please rename any usage of `prost-derive` feature in your `Cargo.toml`.


caspermeijn (approved):

PR #1251, opened 2025-02-27
added extra-platforms feature

### LINKED ISSUE #1250: Add support for targets without Atomic CAS
The Bytes crate requires the target to support atomic CAS, even if default features are disabled. The Bytes crate has a feature for this case, that needs to be enabled, it is called ```"extra-platforms"```. A few experiments on my side did not show any problems.

MTelthoe:
Add feature for no atomic CAS support

closes #1250

### DIFF
--- a/prost/Cargo.toml
+++ b/prost/Cargo.toml
@@ -18,9 +18,10 @@ bench = false
 [features]
 default = ["derive", "std"]
 derive = ["dep:prost-derive"]
-prost-derive = ["derive"]     # deprecated, please use derive feature instead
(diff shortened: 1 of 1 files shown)


caspermeijn (changes requested):
Does this need to happen in `prost`? I feel like adding the feature to your own project is enough, as it will build `prost` with the same version of `bytes` as your own project.

Try adding this to your Cargo.toml:
`bytes = { version = "1", features = "extra-platforms" }`

MTelthoe:
@caspermeijn You're approach works and I agree, that this is a better solution.

PR #1252, opened 2025-02-27
Add serde feature flag to prost-types

### LINKED ISSUE #852: `serde` support for `prost-types`
Hi!

I'm trying to support `serde`-based serialization on some Protobuf types I have generated using `prost`. I do this conditionally by putting the following in the build script for my crate.

```rust
fn main() {
    let protos = 
    prost_build::Config::new()
        .type_attribute(".", "#[cfg_attr(feature = \"serde\", derive(serde::Serialize, serde::Deserialize))]");
        .compile_protos(&[/* list of files */], &[/* list of includes */])
        .unwrap();
}
```

This works great until I use any of the well-known types. The problem, of course, is that the well-known types have already been generated and exist in `prost-types`. The workaround is to use `compile_well_known_types()`, like so.

```protobuf
syntax = "proto3";

import "google/protobuf/timestamp.proto";

message LogWithTime {
    string log = 1;
    google.protobuf.Timestamp timestamp = 2;
}
```

```rust
fn main() {
    let protos = 
    prost_build::Config::new()
        .type_attribute(".", "#[cfg_attr(feature = \"serde\", derive(serde::Serialize, serde::Deserialize))]");
        .compile_well_known_types() // <---- added this line!
        .compile_protos(&[/* list of files */], &[/* list of includes */])
        .unwrap();
}
```

The problem now, however, is that the well-known types I generated are no longer compatible with any other crate's well-known types. So, it becomes difficult to do transformations between the types generated across crates. 

My workaround is to generate the well-known types with the appropriate `serde` attributes in a new shared crate that can be used by all the other crates in my workspace. But, I was wondering: would it be possible to add a feature flag to `prost-types` that would allow one to turn on support for `serde`? Or perhaps the feature flag is better situated in `prost-build`? I'm not sure exactly where the best place to put it is, but generally having support for `serde` would be nice.

allada:
Adds "serde_deserialize", "serde_serialize" and "serde" flags to prost-types. This will allow users to enable serde support to the well known types.

closes #852


prost-types/Cargo.toml
```diff
@@ -16,11 +16,15 @@ doctest = false
 default = ["std"]
 std = ["prost/std"]
 arbitrary = ["dep:arbitrary"]
+serde = ["serde_serialize", "serde_deserialize"]
+serde_serialize = ["std", "dep:serde"]
+serde_deserialize = ["std", "dep:serde"]
```
caspermeijn, line 21: Why is it important to split the feature? The compiler is smart enough to remove unused code from the binary, right? I suggest having a single `serde` feature.

prost-types/Cargo.toml
```diff
@@ -16,11 +16,15 @@ doctest = false
...
 
 [dependencies]
 prost = { version = "0.13.5", path = "../prost", default-features = false, features = ["prost-derive"] }
 arbitrary = { version = "1.4", features = ["derive"], optional = true }
 chrono = { version = "0.4.34", default-features = false, optional = true }
+serde = { version = "1.0.218", default-features = false, optional = true, features = ["derive", "std"] }
```
caspermeijn, line 27: Why is `std` feature needed?

caspermeijn (commented):
- I miss documentation of the feature flag
- What does this mean for stability guarantees? 
- When you serialize these messages to JSON, it will not be [ProtoJSON](https://protobuf.dev/programming-guides/json/) compatible. 
- If we merged this, then https://github.com/tokio-rs/prost/pull/1081 will be a breaking change. Because the JSON names change.

allada:
> * I miss documentation of the feature flag
Done.
> * What does this mean for stability guarantees?
(depends on response)
> * When you serialize these messages to JSON, it will not be [ProtoJSON](https://protobuf.dev/programming-guides/json/) compatible.
Yes, but serde provides more than just json serialization. Maybe we can add note to the documentation stating that json support in serde is not the same as proto's json serialization?
> * If we merged this, then [WIP: Add json (serde) support #1081](https://github.com/tokio-rs/prost/pull/1081) will be a breaking change. Because the JSON names change.
One idea is to gate this feature behind `serde_experimental` (or similar) to give flexibility of breaking changes while proto's json and serde's json gets hammered out?

However, it probably is best to stabilize proto-json and it'd likely get serde naturally. This would dramatically reduce the chance of creating a breaking change.

PR #1254, opened 2025-03-07
tests: Run tests using edition 2024

caspermeijn:


caspermeijn:
@tabokie Do you understand why tests are not failing for the reason in #1248?

tabokie:
I believe warnings from code generated by (external) proc-macros will be suppressed (-> https://github.com/rust-lang/rust/issues/53975#issuecomment-418886946). In our case the macro is manually expanded.

So that specific warning will only appear if you change the origin crate to edition 2024 (and maybe add a dummy struct to expand that macro locally).

caspermeijn:
> I believe warnings from code generated by (external) proc-macros will be suppressed (-> [rust-lang/rust#53975 (comment)](https://github.com/rust-lang/rust/issues/53975#issuecomment-418886946)). In our case the macro is manually expanded.
> 
> So that specific warning will only appear if you change the origin crate to edition 2024 (and maybe add a dummy struct to expand that macro locally).

I see. `prost-build` generates files on disk that is directly includes, so warning in that will trigger in the tests. `prost-derive` generates the code in memory and there warnings are ignored.

Can we enable those warnings somehow? I would like our test suite to find the problems such as #1248

tabokie:
I tested you can get some errors by setting both proc-macro crate and test crate to edition-2024.

I haven't looked closely to understand the mechanics, because this clearly contradicts https://github.com/rust-lang/rust/blob/dea1661cdb6ccaa636abdeea3bd0a9804d3db240/compiler/rustc_middle/src/lint.rs#L336-L338 (edition error should be ignored). But I think you can only get the "hard" compatibility errors in this way. Mild clippy warnings are still silenced.

caspermeijn:
> I tested you can get some errors by setting both proc-macro crate and test crate to edition-2024. (Maybe a similar `prost-derive-2024` that includes `../prost-derive/`?)

Good idea. When I set the edition of the whole workspace, the 2024 lints are applied. See https://github.com/tokio-rs/prost/pull/1256

PR #1266, opened 2025-04-04
change Update timestamp.rs

Hack666r:
replace old link with a new one


caspermeijn (changes requested):
Thank you for taking the time to fix the URL. That new URL doesn't have the correct tag on the end.

caspermeijn:
Fixed via #1336

Issue #1031, opened 2024-04-18
Add support for editions


ouillie:
[Editions](https://protobuf.dev/editions/overview/) are the future of protobuf syntax.

Currently, when you set `edition = "2023"` at the top of a `.proto` file, and run it through prost-build using `protoc_arg("--experimental_editions")`, it will panic on [this line](https://github.com/tokio-rs/prost/blob/04be6049ea676dde5dfac8a21da2559f4f380ed2/prost-build/src/code_generator.rs#L72) because the value of `FileDescriptorProto.syntax` [is `"editions"`](https://github.com/protocolbuffers/protobuf/blob/e5502c746ef62b3d76f589545803cfb46a5082ea/src/google/protobuf/descriptor.proto#L126). That's pretty much the extent of my investigation. I just figured I'd file this bug since support for editions doesn't seem to be tracked elsewhere.

This may sound like it's adding complexity to Prost, but it looks like it should actually simplify Prost once [Prototiller](https://protobuf.dev/editions/overview/#prototiller) is released, since that will effectively normalize all the different syntaxes and editions into a canonical representation using features, and Prost will only have to support that canonical implementation in the long term. It would be a somewhat heavy refactor up front, though.

caspermeijn:
I agree, editions are important. However, due to backwards compatibility of `protoc` itself, we need to keep supporting `proto2` and `proto3` for some time. 

Are you willing to do the work to add this support?

ouillie:
That makes sense. I won't have bandwidth for a couple months, but I'm down to contribute, especially after Prototiller is released.

mzabaluev:
Up to spec support for editions requires at least the following: 

* Unknown fields, as discussed in #2.
* Open and closed enumeration behavior. I've started on this in #1079.

esrauchg:
Drive by chiming in from someone on the Protobuf team at Google here.

Unfortunately open sourcing the prototiller tool is held up so we still don't have a date for that (sorry!), but in nearly all cases manually upgrading a file from Proto3 to Editions 2023 files shouldn't be too hard so there shouldn't be a very strong reason to wait

[...]

PR #1271, opened 2025-04-23
Update protobuf to v29.4

hmliu6:
Release note:
https://github.com/protocolbuffers/protobuf/releases/tag/v29.4


caspermeijn:
I believe we currently use v25.4, so I think this one release note is missing a bunch 😜

This `protobuf` submodule is used during testing and CI build. I see the CI is failing due to the new editions system. So support for the edition system should be added before this can be merged: https://github.com/tokio-rs/prost/issues/1031 Feel free to help out with that issue, then we can come back to this PR after that

PR #1278, opened 2025-05-05
Config option to use proto `enum`s as Rust `enum`s

tkr-sh:
By default, protos `enum`s are converted to Rust `i32`. Which can be annoying / not convenient.

I created a new `Config` field - `proto_enum_to_rust_enum` - that, when activated, converts proto's `enum`s into Rust's `enum` instead of `i32`,
The data transferred over the wire is exactly the same as `i32`. (_See: `prost-derive/src/field/enum_typed.rs`_)

---

_Not sure where to put the test in `prost-build/src/code_generator.rs`?_


tkr-sh:
This might be a fix to https://github.com/tokio-rs/prost/issues/276 btw

tkr-sh:
Yes! But this one is really easy to implement since it keeps the same declaration of enums, it's just that the types used in `message`s (in the Rust representation) are  `enum`s.
I think that it could be a nice addition, even to #1079 in case you don't want to handle open enums.
Also, would be really easy to add support for `Result<_, i32>` https://github.com/tokio-rs/prost/pull/1079#issuecomment-2197525664

PR #1279, opened 2025-05-11
feature(custom-derive) Adds config options to customize derives in code_generation

aenriii:
first time making a pr to a decently sized project, but here goes;

I was working on a personal project of mine and found that one of the few things that prost doesn't have compared to the regular rust protobuf crates is custom derives (though in that crate its an entire `Customizer` struct for that stuff), this just kinda adds a few little properties and a `fn` or two to get a similar functionality, based on my usages it seems to work well (I'm using this for `serde::Deserialize` and `serde::Serialize`)

There's almost definitely a better way to implement this but I think if this feature is to be implemented this is a good start

### DIFF
--- a/prost-build/Cargo.toml
+++ b/prost-build/Cargo.toml
@@ -10,10 +10,10 @@ edition.workspace = true
 rust-version.workspace = true
 
 [features]
-default = ["format"]
+default = ["format", "custom-derive"]
 format = ["dep:prettyplease", "dep:syn"]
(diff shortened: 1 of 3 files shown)


LucioFranco:
Hi thanks for the contribution! Maybe I am missing something but can you not just use https://docs.rs/prost-build/latest/prost_build/struct.Config.html#method.message_attribute to achieve the same result?

aenriii:
im gonna be really honest i was really tired when i was working on the project i was using this for and i think i completely missed that, ill convert my code over to that

PR #1283, opened 2025-05-26
Add rich error message for context message graph

### LINKED ISSUE #1282: Panic without good error message in `prost-build` when message not in graph
Just like #932, `prost-build` panics without a good error message when a message is somehow not in the message graph:

https://github.com/tokio-rs/prost/blob/1e93f5612d70a292d6d3919bfc6af22bf82e481a/prost-build/src/context.rs#L172-L180

Encountered when trying to compile potentially well-known protos within `google/rpc/*.proto`:

```
  thread 'main' panicked at /Users/kris/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/prost-build-0.13.5/src/context.rs:181:14:
  called `Option::unwrap()` on a `None` value
```

It would be nice rich errors were provided, potentially something simple like `anyhow::{Context, Error}`, instead of panicing without specifying which message name the build had a problem with.

kriswuollett:
Fixes #1282

### DIFF
--- a/prost-build/src/context.rs
+++ b/prost-build/src/context.rs
@@ -173,6 +173,7 @@ impl<'a> Context<'a> {
         assert_eq!(".", &fq_message_name[..1]);
         self.message_graph
             .get_message(fq_message_name)
+            .ok_or_else(|| format!("message `{fq_message_name}` not found in message graph"))
             .unwrap()
             .field
             .iter()


caspermeijn (approved):
This still doesn't help a user when it gets this error. But at least it is better than an unwrap on `None`

Issue #1284, opened 2025-05-26
`prost-build` Extern path message type not usable as a field type (panics)


kriswuollett:
The `Context::can_field_derive_copy` does not check if a message type is an extern before seeing if the type can copy which causes a panic as discovered in #1282:

https://github.com/tokio-rs/prost/blob/1e93f5612d70a292d6d3919bfc6af22bf82e481a/prost-build/src/context.rs#L208-L210

(same for `can_message_derive_eq`)

Looks like the externs need a way to saying what traits they implement. Possibly externs need to be configured with a rich struct instead of a just tuple of `&str` elements? Or just tag on things like `(".google.protobuf.Timestamp", "prost_wkt_types::Timestamp+Copy+PartialEq+Eq")` so that a new optional trait implementation does not break builds?

kriswuollett:
Perhaps it is not so much that it is an extern, but perhaps it is `compile_fds` has to include the full dependency set in the `FileDescriptorSet`? That's why I assume that is why I'm encountering this issue. Was trying out old code related to #880 again.

PR #1285, opened 2025-05-26
Don't crash when extern type used as field

kriswuollett:
Workaround for extern types lacking trait implementations metadata in #1284.


caspermeijn (changes requested):
This is a good workaround. Please add a test to prevent a future regression.

PR #1286, opened 2025-05-28
feat(config): add force_required_messages option to control message field generation in proto3

yordis:
This update introduces a new configuration option, `force_required_messages`, allowing users to enforce non-optional message fields in proto3. By default, prost generates `Option<T>` for message fields, but this option enables direct `T` types instead. A corresponding test has been added to verify the functionality.


yordis:
better name for `force_required_messages` welcome eventually, once the feature works

yordis:
A lot of changes are related to clipply 😭

yordis:
Hey @caspermeijn from a three thousand foot view, does this makes sense to you? I am not an Rust dev, and barely worked on protobuf stuff before 😰

mzabaluev:
In my understanding, message-typed fields are always nullable and their presence cannot be required by the specification in all published editions of Protobuf, so this option introduces out-of-spec behavior to implement custom application logic. In my opinion, such customizations are a poor fit for what Prost aims to do (more specifically, prost-build is **not** a code generator for your application's data domain types), and every customization feature adds to maintenance cost and may break add-on code generators and derive macros.

PR #1298, opened 2025-06-17
Fix typo: "unparseable" instead of "unparseble" in docs and proto

leopardracer:
Description:
Corrects a typo in documentation and proto comments, changing "unparseble" to the correct "unparseable" for clarity and consistency.


caspermeijn:
Please fix this upstream. This file is copied from upstream `protobuf` library: https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/compiler/plugin.proto

PR #1302, opened 2025-07-08
Add Serde Impls to json-like `prost-types` Structs

krakow10:
This allows the ListValue, Struct, and Value types to be serialized with serde by enabling a new "serde" crate feature.

This code was originally written by ChatGPT (and me) in July 2024, back when I would not have been able to write it entirely myself.  It had dependency issues and a strange nested Visitor implementation, but it worked well enough to use privately.  I've learned a lot of Rust in a year, and went over it today and was able to cut out 70 lines of ai cruft nonsense and use a crate feature.  The reason I'm upstreaming this now is because I saw that the "arbitrary" feature was added which looks similar to serde.  I briefly tested this in my game's staging deployment and it seems to work well enough to not cause any issues.

### DIFF
--- a/prost-types/src/serde.rs
+++ b/prost-types/src/serde.rs
@@ -1,0 +1,215 @@
+use crate::protobuf::value::Kind;
+use crate::protobuf::{ListValue, Struct, Value};
+use serde::de::{Error, MapAccess, SeqAccess, Visitor};
+use serde::ser::{SerializeMap, SerializeSeq};
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+
+impl Serialize for Struct {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let mut s = serializer.serialize_map(Some(self.fields.len()))?;
+        for (k, v) in &self.fields {
+            s.serialize_entry(k, v)?;
+        }
+        s.end()
+    }
+}
+impl<'de> Deserialize<'de> for Struct {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        struct StructVisitor;
+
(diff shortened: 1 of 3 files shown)


caspermeijn (changes requested):
Thank you for your contribution. This looks like a good first step for a JSON implementation. I would like to see the following additions:
- Describe the `serde` feature flag in the `lib.rs` doc comment. This has to at least say that not all types are serializable
- Please add some tests to prevent future regressions
- Can you confirm that this serialization follows the official [ProtoJSON format](https://protobuf.dev/programming-guides/json/)? It would be super if you could confirm that in a test case.

prost-types/src/serde.rs
```diff
@@ -0,0 +1,212 @@
...
+                }
+                Ok(Struct { fields })
+            }
+        }
+
+        deserializer.deserialize_map(StructVisitor)
+    }
+}
+
+impl Serialize for Value {
```
caspermeijn, line 49: Would it make sense for this impl to be on `Kind` and have an extra impl for `Value` that just passes it through to `Kind`?

krakow10: Great suggestion, I've implemented it.

krakow10:
> Thank you for your contribution. This looks like a good first step for a JSON implementation. I would like to see the following additions:
> 
>     * Describe the `serde` feature flag in the `lib.rs` doc comment. This has to at least say that not all types are serializable

I've added a description.  Please let me know if it's insufficient.

>     * Please add some tests to prevent future regressions

Done.

>     * Can you confirm that this serialization follows the official [ProtoJSON format](https://protobuf.dev/programming-guides/json/)? It would be super if you could confirm that in a test case.

Eyeballing it, it seems like it won't make something invalid for serialization, but for deserialization it looks like converting `{} -> protobuf.Empty` is not possible with the current types.  The current behaviour is to produce an empty `Value::StructValue`.  I'm not sure how to approach adherence to the specification in a testable way.

Issue #285, opened 2020-02-21
Improve the usage of decode_length_delimiter / Message::decode_length_delimited


mawenqi:
Say the buf received is like:
```
// the total length of following two struct with fixed width(4)
0x00, 0x00, 0x00, 0x32, // = 50
// varint encoding of struct A length 
0x1e, // = 30
0x08, 0x02, 0x10, 0x00, 0x18, 0xfd, 0xff, 0xff, 0xff, 0x0f,
0x22, 0x10, 0x87, 0xeb, 0x86, 0xd4, 0x9c, 0x95, 0x4c, 0x15,
0x8a, 0xb0, 0xd7, 0xbc, 0x2e, 0xca, 0xca, 0x37, 0x28, 0x01,
// varint encoding of struct B length 
0x12, // = 18
0x12, 0x0a, 0x0a, 0x08, 0x65, 0x6c, 0x65, 0x69, 0x62, 0x6f,
0x76, 0x69, 0x1a, 0x04, 0x70, 0x69, 0x6e, 0x67
```

How can we decode it? It seems we have to do this:
```
let len = buf.get_i32();  // len = 50
    
let a_len = decode_length_delimiter(buf.clone())?; // a_len = 30
let a_len_len = length_delimiter_len(a_len);  // a_len_len = 1
buf.advance(a_len_len);
let a_buf = buf.split_to(a_len); 
let a = A::decode(a_buf)?;
    
let b_len = decode(buf.clone())?;  // b_len = 18
let b_len_len = length_delimiter_len(b_len);  // b_len_len = 1
buf.advance(b_len_len);
let b_buf = buf.split_to(b_len); 
let b = B::decode(b_buf)?;
```
As you can see, it is tedious. Is there a better way?  Many thanks.

BTW, if decode_length_delimiter and Message::decode_length_delimited 
could just consume accurate size of the buf but not the whole, then I can do this:
```
let len = buf.get_i32();  // len = 50
let a = A::decode_length_delimited(&mut buf)?;
let b = B::decode_length_delimited(&mut buf)?;
```
Thanks.

danburkert:
In general, it's up to the application to decide how to do message delimiting (see [upstream docs](https://developers.google.com/protocol-buffers/docs/techniques#streaming)).  Part of the reason that looks awkward is that you are cloning `buf` before passing it into `decode_length_delimiter`, then calling `advance` manually; `decode_length_delimiter` should advance the internal buffer the correct amount automatically.

[...]

PR #1311, opened 2025-08-11
docs: add example for `decode_length_delimiter`

armandas:
From [documentation](https://docs.rs/prost/0.14.1/prost/fn.decode_length_delimiter.html), it's not obvious how to decode the delimiter given the larger buffer. I think it would be helpful if there was an example.

Looking through the issues (e.g. #285), I found out that using `Bytes` or `BytesMut` instead of `&[u8]` makes everything work. 

I pieced this example from what I could find in the issues, so let me know if this is not the most idiomatic way and I will update the PR.

### DIFF
--- a/prost/src/encoding/length_delimiter.rs
+++ b/prost/src/encoding/length_delimiter.rs
@@ -9,7 +9,7 @@ use crate::encoding::varint::{decode_varint, encode_varint, encoded_len_varint};
 
 /// Encodes a length delimiter to the buffer.
 ///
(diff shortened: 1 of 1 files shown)


armandas:
Any feedback on this? @caspermeijn

caspermeijn (changes requested):
This makes sense. Please rebase to latest master.

While you are at it, could you fix the link in the doc comment to `[Message::encode_length_delimited]`?

armandas:
Thanks for the feedback. Rebased and fixed the comments.

PR #1313, opened 2025-08-12
feat: make is_valid a constant functionin `#[derive(::prost::Enumeration)]`

rluvaton:


caspermeijn (approved):

PR #1318, opened 2025-09-06
Add comprehensive Edition 2023 support to prost-build

### LINKED ISSUE #1031: Add support for editions

techtoine:
## Summary
Adds comprehensive Edition 2023 support to prost-build, enabling Rust code generation from Edition 2023 protobuf files.

## Changes
- Add Edition2023 syntax variant to support editions syntax
- Implement explicit field presence (Option<T>) for Edition 2023
- Add reserved field handling with proper comments
- Support field options (deprecated, packed) for Edition 2023
- Add validation for Edition 2023 limits (3100 fields, 1700 enum values)
- Maintain backward compatibility with proto2/proto3

## Testing
Tested with real Edition 2023 proto files - generates correct Rust code with explicit field presence.

Fixes #1031

### DIFF
--- a/prost-build/src/code_generator.rs
+++ b/prost-build/src/code_generator.rs
@@ -172,6 +172,14 @@ impl<'b> CodeGenerator<'_, 'b> {
         let message_name = message.name().to_string();
         let fq_message_name = self.fq_name(&message_name);
 
+        // Validate Edition 2023 limits
+        if let Err(e) = self.validate_edition_2023_limits(&message) {
+            panic!("Edition 2023 validation failed: {}", e);
+        }
+
+        // Add reserved fields handling for Edition 2023
+        self.append_reserved_fields(&message);
+
         // Skip external types.
         if self
             .context
@@ -442,6 +450,18 @@ impl<'b> CodeGenerator<'_, 'b> {
                 .push_str(&format!(" = {:?}", bytes_type.annotation()));
         }
 
+        // Handle field options for Edition 2023
+        if self.syntax == Syntax::Edition2023 {
+            if let Some(ref options) = field.descriptor.options {
+                if options.deprecated.unwrap_or(false) {
+                    self.buf.push_str(", deprecated");
+                }
+                if options.packed.unwrap_or(false) {
+                    self.buf.push_str(", packed = \"true\"");
+                }
+            }
+        }
+
         match field.descriptor.label() {
             Label::Optional => {
                 if optional {
@@ -456,7 +476,7 @@ impl<'b> CodeGenerator<'_, 'b> {
                         .descriptor
                         .options
                         .as_ref()
-                        .map_or(self.syntax == Syntax::Proto3, |options| options.packed())
+                        .map_or(self.syntax == Syntax::Proto3 || self.syntax == Syntax::Edition2023, |options| options.packed())
                 {
                     self.buf.push_str(", packed = \"false\"");
                 }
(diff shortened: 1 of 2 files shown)


abarke:
Relates to https://github.com/tokio-rs/prost/issues/1031

alebar42:
@techtoine probably be worth updating the [README.md](https://github.com/tokio-rs/prost/blob/master/README.md?plain=1#L11C44-L11C72) to align with the new functionality.

mateuszlitwin:
Same, we use editions in our APIs and this prevents wider adoption in Rust.

esrauchg:
Driveby comment from a engineer on Google's Protobuf Team. We don't usually spend time validating/vouching/whatever open source implementations of Protobuf, so this is just comments from an engineer that the maintainers of Prost can take or leave.

Also, I only skimmed the PR and apologies if I missed some details already handled.

1) I think the listed implementation appears to be not quite enough to pass Google's published conformance test suite. I know that Prost is hooked up to that test suite and I believe has a good pass rate, so I think it should show up there.

I think likely you do want to do a change so that Prost's conformance testee is hooked up for Edition 2023 so you can iterate on making that pass if you wanted to commit this PR.

The main thing there is that under Editions behavior can be controlled by features which cascade down, so you can't just replace the syntax==Proto2 or syntax==Proto3 checks with a syntax==Edition2023, you have to resolve the corresponding features from the outer scopes, as described here:

https://protobuf.dev/editions/implementation/#:~:text=specific%20test%20cases-,Feature%20Resolution,-Editions%20use%20lexical

Concrete example is that in Editions, repeated fields are packed by default, but it can be overriden on outer scopes, so:

```
edition = "2023";

message Foo {
  option features.repeated_field_encoding = EXPANDED;

  repeated int32 x = 1;
  message Bar {
    repeated int32 y = 2;
    repeated int32 z = 3 [features.repeated_field_encoding=PACKED];
}
```

In this example, x and y are unpacked and z is packed.

Note that the 'feature resolution' (cascading the values down) is implemented in our own implementations and so generators which are built using one of our implementations don't need to implement this themselves (and Google's own generators are mostly implemented using C++Proto and leverage that one). Since Prost's generator is implemented in Rust without using a Protobuf library it would need to implement that feature-cascading behavior to properly support Editions (either by crawling upwards in the scopes every time you want to resolve a feature, or by ahead of time traversing downwards and building up structure that has the resolved features on the message/enums/fields).

2) Much smaller topic, but I think you don't need to try to enforce the additional limits in Edition 2023; maybe I'm confused but AFAIK we just enforce the limits in Protoc frontend which will run before it passes the parsed descriptor information to Prost's plugin.

Hope that helps!

caspermeijn (changes requested):
I understand the need for this, but it represents a big change. I expected a lot of tests for checking the behavior in the different editions. 

I want to see at least a test per edition 2023 specific code block and I want to see the conformance tests for edition 2023 pass.

This probably requires the `third_party/protobuf` repo to be updated, which causes some compile error I have not looked into.

caspermeijn:
The failing macos test is fixed in the latest main branch

PR #1327, opened 2025-09-11
build(deps): update petgraph to 0.8

timsaucer:
Update dependency

### DIFF
--- a/prost-build/Cargo.toml
+++ b/prost-build/Cargo.toml
@@ -19,7 +19,7 @@ heck = { version = ">=0.4, <=0.5" }
 itertools = { version = ">=0.10, <=0.14", default-features = false, features = ["use_alloc"] }
 log = "0.4.4"
 multimap = { version = ">=0.8, <=0.10", default-features = false }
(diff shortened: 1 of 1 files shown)


caspermeijn:
> Since the plan is to keep prost in line with tokio and tokio is [not interested in updating MSRV](https://github.com/tokio-rs/tokio/pull/7610) right now, the no-default-features for petgraph in 0.8 will not pass CI. That requires at least rust 1.81 which has `core::error::Error` stable for use with `#![no_std]`. We cannot pass this as a `features = ["std"]` to the petgraph dependency because versions 0.6 and 0.7 do not have it.

So, `prost` follows the same MSRV policy, but not nessesaribly the same MSRV version. So a version bump is possible if that is needed for the project. However, an MSRV bump for a build-dependency update doesn't feel right.

> Instead, just changing the dep here to stick it to 0.8 instead of allowing a range.

`petgraph 0.8` is now 6 months old, so dropping the older versions is fine with me.

caspermeijn (approved):
Thank you for bringing this to my attention.

PR #1330, opened 2025-09-18
style: Simplify DecodeError description to an enum

caspermeijn:
Replace the inner `description` field of `DecodeError` from a string to an enum. The enum implements `Display` that produces the same string.

Split the error for `Any::to_msg` into `DecodeAnyError`. This provides better separation between `prost` and `prost-types`.


prost/src/error.rs
```diff
@@ -67,7 +66,71 @@ impl fmt::Display for DecodeError {
...
+            inner: Box::new(Inner {
+                description,
+                stack: Vec::new(),
+            }),
+        }
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum DecodeErrorKind {
```
LucioFranco, line 85: do we want this to be either `non_exhaustive` or `pub(crate)`? I don't think we want to lock ourselves in like this?

caspermeijn:
Oops, I changed the return type of a `pub fn`, which is not yet implemented in `cargo-semver-checks`. This makes the current version a breaking change.

prost/src/error.rs
caspermeijn: You are right, this should be private

prost/src/error.rs
```diff
@@ -67,7 +70,78 @@ impl fmt::Display for DecodeError {
         for &(message, field) in &self.inner.stack {
             write!(f, "{message}.{field}: ")?;
         }
-        f.write_str(&self.inner.description)
+        write!(f, "{}", self.inner.description)
+    }
+}
+
+impl From<DecodeErrorKind> for DecodeError {
```
LucioFranco, line 77: thoughts on making this a fn on the struct rather than a trait impl, iirc this trait impl will expose the error variant that isn't public but if we have a constructor that can be doc hidden then we can avoid that public api item?

caspermeijn: Is that true? `DecodeErrorKind` is `pub(crate)`, so external crates can't name it. So I assumed they can't use this impl. 

I ran `cargo doc` and could not find `DecodeErrorKind` in the webpage.

I feel like a doc hidden function will increase the chance of someone using it as not intended. 

Did I understand your comment correctly?

LucioFranco: iirc for DecodeError there will be a `From<DecodeErrorKind>` where the `DecodeErrorKind` is greyed out (not clickable) but you can see the type. Its not that big of a deal more so just feels cleaner to not use trait conversion if the conversion points are universal. Now I will admit its fine if we have a lot of usage of `?` which use the into conversion. But I wonder if we can actually just move the From to a `Into<DecodeError>` on the enum variant and that would be hidden completely.

caspermeijn: I see what you mean. I believe that the documentation did do that in older toolchains, but it doesn't with the stable compiler. 

With rust 1.90, the word `kind` doesn't appear on the doc page for `DecodeError.`

LucioFranco (approved):

PR #1338, opened 2025-09-21
Allow encoding/decoding strings as `Arc<str>`

alex:
This raises the MSRV for some of the `MaybeUninit` APIs. This is acceptable under the tokio MSRV policy.


alex:
I suspect this is close, but alas I'm struggling to get tests passing locally (they fail building protobuf with bizzare errors) and debugging via CI is far too slow. Hopefully this will be useful to someone else down the line.

Issue #174, opened 2019-04-04
Allow ignoring fields


mashedcode:
If one has a structure which carries data and temporary state. One might want to encode only the data and ignore the rest.

It seems to me that the simplest way to implement this would be to simply ignore fields which are _not_ annotated.
As part of this I'd also stop generating [`impl Default`](https://github.com/danburkert/prost/blob/26adada0bdd16f45e327bdca21c60c9e49c935b9/prost-derive/src/lib.rs#L210) since a user may very well want to implement it himself to initialize the ignored fields correctly.
For generated code a simple `#[derive(Default` should do the trick IIUC.
Either way this would be a breaking change.

Suggestions?

I'd be happy to implement.

243011068:
I really need this

DontBreakAlex:
I also found myself needing this. I have a existing very large struct and I don't want to clone it only to drop a field...

commonsensesoftware:
I do not think this applies to types that use `*.proto` files because they simply would not be code-generated. I have a use case where I want to model a message and ProtoBuf just happens to be the serialization format I want to use. I don't what to have to create DTOs or other intermediate structs just to drop the unwanted field.

Example:

```rust
#[derive(Message)]
struct Event {
    #[prost(uint32)]
    id: u32,

    #[prost(skip)]
    version: u32,

   /// ... other fields
}
```

I have no interest in storing `version` because it's irrelevant to the encoded message. I need and store that information outside of the struct; however, it's a lot of extra work and ceremony to wrap that behavior over another struct or provide a conversion struct. Similarly, having `version: Option<u32>` and setting it to `None` isn't much help. It's still a field in ProtoBuf's eyes.

It doesn't really matter what it's called. `serde` uses `#[serde(skip)]` so this would be congruent, but `#[prost(ignore)]` is just as meaningful IMHO. This problem can be exacerbated by a field that is much larger in size. There should be _some_ way to ignore such a field. Since every field used this way requires an annotation, I'm also onboard with simply ignoring any field that doesn't have an annotation.

caspermeijn:
What do you expect the value to be when such a struct is deserialized? 

Having a `skip` attribute makes sense to me. It needs someone to actually implement this.

commonsensesoftware:
I'm open to possibilities, but I'd imagine rules similar to `serde`.

1. The field must support `Default`, which works for any primitive and recursively, in particular, for `Option`
[...]

PR #1339, opened 2025-09-22
Skip fields

maxpowa:
This change will allow Prost users to skip a field during serialization and deserialization. It also provides an optional `default` attribute for these fields, in the event the underlying object does not implement the `Default` trait.

My motivation for this is some reverse engineering effort, in which the existing datastructures are serialized in both protobuf and xml formats. I can use `quick_xml`/`serde` and ignore protobuf-only fields, but I can't ignore xml-only fields using `prost`, which results in duplicate data structures and more boilerplate to convert between the two types.

Usage:

```rs
#[derive(::prost::Message)]
struct Test {
    #[prost(int32, tag = "1")]
    a: i32,
    #[prost(skip)]
    b: i32,
    #[prost(skip, default = "create_foo")]
    c: Foo,
}

#[derive(Debug)]
struct Foo(i32);

pub fn create_foo() -> Foo {
    Foo(12)
}
```

Related: https://github.com/tokio-rs/prost/issues/174

---

This appears to work locally, but I can't get the full test suite to run (even with clean main branch). I'd like to add some integration tests for this, but it seems the only tests for derive currently are the existing Protobuf known-types (where this wouldn't really make sense), and a few negative test cases in the derive package.

### DIFF
--- a/prost-derive/src/lib.rs
+++ b/prost-derive/src/lib.rs
@@ -84,6 +84,9 @@ fn try_message(input: TokenStream) -> Result<TokenStream, Error> {
     // We want Debug to be in declaration order
     let unsorted_fields = fields.clone();
 
+    // Filter out ignored fields
+    fields.retain(|(_, field)| matches!(field, Field::Skip(..)));
+
     // Sort the fields by tag number so that fields will be encoded in tag order.
     // TODO: This encodes oneof fields in the position of their lowest tag,
     // regardless of the currently occupied variant, is that consequential?
--- a/prost-derive/src/lib.rs
+++ b/prost-derive/src/lib.rs
@@ -85,14 +85,19 @@ fn try_message(input: TokenStream) -> Result<TokenStream, Error> {
     let unsorted_fields = fields.clone();
 
     // Filter out ignored fields
-    fields.retain(|(_, field)| matches!(field, Field::Skip(..)));
+    fields.retain(|(_, field)| !matches!(field, Field::Skip(..)));
 
     // Sort the fields by tag number so that fields will be encoded in tag order.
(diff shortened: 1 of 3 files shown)


caspermeijn (changes requested):
I like the skip option. I see how this could be useful when using `prost-derive`.

I am not sure how I feel about the default attribute. Why doesn't each of the skipped types implement default?

I miss tests. You could take some inspiration from `tests/src/custom_debug.rs` or `prost-derive/src/lib.rs`.

I miss documentation. The documentation of derive attributes is not great, but this new attr should at least be mentioned in `README.md`

prost-derive/src/lib.rs
```diff
@@ -128,29 +136,36 @@ fn try_message(input: TokenStream) -> Result<TokenStream, Error> {
...
         .map(|(field_ident, field)| field.clear(quote!(self.#field_ident)));
 
+    // For Default implementation, use all_fields (including skipped)
     let default = if is_struct {
-        let default = fields.iter().map(|(field_ident, field)| {
-            let value = field.default(&prost_path);
+        let default = all_fields.iter().map(|(field_ident, field)| {
+            let value = match field {
+                Field::Skip(skip_field) => skip_field.default_value(),
+                _ => field.default(&prost_path),
```
caspermeijn: Why are skipped fields handle special in this file? I expect the `Field::Skip` to have the same API as the other fields.

caspermeijn (changes requested):
I like the documentation and tests.

prost-derive/src/lib.rs
maxpowa: Addressed, pushed down to `mod.rs`, similar to `Scalar` field handling.

PR #1354, opened 2025-10-28
tests: fix includes and versions

NFTESG:


NFTESG:
Yes i can. 
I re-enabled a test on an optional proto field, which needs the commandline parameter in 7e73f.
Then i added a mission path for this test in 31840.
Unfortunately, i pinned the toolchain to 1.88.0 without specifying dependencies like "clippy" which breaks the liting.
I added these dependencies, and will push in the next minutes again. It should run then.

PR #1355, opened 2025-11-08
feat: Add support for Arc<str> as a string type

alex:
This is modeled on the support for multiple types for bytes.


alex:
Understanding the new maintenance policy, I'm hoping this counts as a small improvement (and thus is under consideration), rather than a new feature :-)

If there's anything I can do to make this easier to review, please don't hesitate to ask.

PR #1367, opened 2025-11-14
fix: optional enum getter/setters

### LINKED ISSUE #1027: getter/setter generated for optional enum fields should use Option

ThoFrank:
- **Add failing test**
- **Getter for optional field now returns Option<T>**
- **Setter for optional field now accepts Option<T>**
- **Add field_or_default getter**

fixes #1027

### DIFF
--- a/prost-build/src/code_generator.rs
+++ b/prost-build/src/code_generator.rs
@@ -415,8 +415,8 @@ impl<'b> CodeGenerator<'_, 'b> {
     }
 
     fn append_field(&mut self, fq_message_name: &str, field: &Field) {
-        let type_ = field.descriptor.r#type().unwrap_or_default();
-        let repeated = field.descriptor.label().unwrap_or_default() == Label::Repeated;
+        let type_ = field.descriptor.type_or_default();
+        let repeated = field.descriptor.label_or_default() == Label::Repeated;
         let deprecated = self.deprecated(&field.descriptor);
         let optional = self.optional(&field.descriptor);
         let boxed = self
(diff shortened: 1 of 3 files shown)


ThoFrank:
I came across issue #1027 today and decided to try to fix it. Maybe it makes sense to also add a second getter maybe `field_or_default()` which would be just a shortcut for `field().unwrap_or_default()`.
What do you think?

tests/src/default_enum_value.proto
```diff
@@ -12,15 +12,15 @@ enum PrivacyLevel {
 }
 
 message Test {
-  optional PrivacyLevel privacy_level_1 = 1 [default = PRIVACY_LEVEL_ONE];
-  optional PrivacyLevel privacy_level_3 = 2 [default = PRIVACY_LEVEL_PRIVACY_LEVEL_THREE];
-  optional PrivacyLevel privacy_level_4 = 3 [default = PRIVACY_LEVELPRIVACY_LEVEL_FOUR];
+  required PrivacyLevel privacy_level_1 = 1 [default = PRIVACY_LEVEL_ONE];
```
caspermeijn: Why it this test changed to `required`?

caspermeijn (commented):
I agree that the current generated code for an optional is strange. However, this is a breaking change, that could result is a lot of code changes for users of the generated code. Your changes in `code_generator.rs` are a good example for that. 

I think we should park this PR until the next breaking release and re-evaluate at that time whether such a change is wanted.

PR #1378, opened 2025-12-01
feature: expose public error API for decode errors

### LINKED ISSUE #1377: feature: improve the error API
The current error API is insufficient for advanced use-cases.

The two main limitations are:
- there is no access to structured error information
- there are no public constructors

Some example use-cases where this causes issues:
- integration with error reporters using the `core::error::Error::source` API
- interoperability with types using other `Prost` major version (no way to bridge the error types)
- keeping structured information through serialization
- asserting the received error in tests

The previous stringy API was not great either, but it was possible to handle some of these use cases since the constructor was public. Following the changes in https://github.com/tokio-rs/prost/pull/1330, this is no longer possible without hacks.

demurgos:
This commit updates the Error API and exposes it publicly.

The location of the error becomes fully accessible to consumer code through the `ErrorPath` abstraction. It allows to read the logical path without exposing implementation details.

The error kind is turned into an enum of source errors. Each source error and the error kind itself are non-exhaustive. This allows to new fields in the future without breaking backwards compatibility.

Public constructors are provided to enable advanced use-cases in third-party crate that want to create new error types.

`DecodeError` uses the `core::error::Error` API to expose source errors. In particular, it does not print the source error in its own display message. This allows better integration with error reporting frameworks.

- Closes #1377

### DIFF
--- a/prost-derive/src/lib.rs
+++ b/prost-derive/src/lib.rs
@@ -117,7 +117,9 @@ fn try_message(input: TokenStream) -> Result<TokenStream, Error> {
             #(#tags)* => {
                 let mut value = &mut self.#field_ident;
                 #merge.map_err(|mut error| {
-                    error.push(STRUCT_NAME, stringify!(#field_ident));
+                    error.path_mut().push_segment(
+                        #prost_path::ErrorPathSegment::new(STRUCT_NAME, stringify!(#field_ident)),
+                    );
                     error
                 })
             },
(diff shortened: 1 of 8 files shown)


caspermeijn:
Sorry, but this is way too complex for what `DecodeError` is supposed to do. It is intended to provide the user some idea of what has happened. It is not intended to be a debugging tool. I understand there can be a need for improvements to the type, but is too much for the prost project to maintain at this moment.

I will close this for now. Feel free to open a smaller PR with a clear description why the improvement is necessary.

PR #1380, opened 2025-12-12
Add custom_scalar tag as a extension point

Guiguiprim:
I'm not sure if you will be interested by this PR but I will try my luck as I feel like other might use it (otherwise I will keep it on a fork).

Context: For performance and memory usage reasons I wanted to optimise the data structures I use with `prost`. In my use-case I wanted to use a string interner for some fields and arrays instead of bytes for others, but this could also be used for small string optimisation or tiny vec and more.

To do it in a generic way, I added:

- a trait `CustomScalarInterface`
- support for a `custom_scalar(interface_path)` tag in `prost-build` and `prost-derive`

This make it possible to implement as a crate user any scalar wrapping type without needing to touch `prost`. It is an expert API as the user would need to correctly implement `CustomScalarInterface` in a way that is compatible with the declared protobuf type.

### DIFF
--- a/prost-build/src/code_generator.rs
+++ b/prost-build/src/code_generator.rs
@@ -450,7 +450,7 @@ impl<'b> CodeGenerator<'_, 'b> {
         let type_tag = self.field_type_tag(&field.descriptor, custom_module_path.as_deref());
         self.buf.push_str(&type_tag);
 
-        if type_ == Type::Bytes {
+        if type_ == Type::Bytes && custom_module_path.is_none() {
             let bytes_type = self
                 .context
                 .bytes_type(fq_message_name, field.descriptor.name());
--- a/tests/build.rs
+++ b/tests/build.rs
@@ -196,6 +196,11 @@ fn main() {
                 ".custom_scalar.Msg.g.key",
             ],
         )
+        .custom_scalar(
+            prost_types::field_descriptor_proto::Type::Bytes,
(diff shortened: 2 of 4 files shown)


caspermeijn (changes requested):
Yes, I am interested. There have been a lot of requests for custom scalars.

I have been trying something similar to the suggested trait, but have not found the time to do it properly.

This PR is too large to review and merge, but I would like to work towards something like this. I don't have the time to review large PRs.

This has been [my take](https://github.com/tokio-rs/prost/compare/master...caspermeijn:prost:encoding#diff-c134a23edd392ec91f24f0a98d0a483d20cf33b86caf3f7a47a9f9ca09a4c87c) on the trait. I think yours is better.

However, I think this should not be focused at custom scalars. I see this PR adds a lot of code. Mainly: if custom scalar, do something different. This adds multiple code paths and therefore added maintainer load. I would like to reduce maintainer load.

Can we first use this trait for the internal types? When we are happy with the internal usage, then we open it to the public. This has mainly to do with breaking changes, which I want to keep to a minimum.

I suggest doing a simpler PR that adds this trait for `bytes`. There are currently two options for the rust type to represent protobuf `bytes` (see [`Config::bytes`](https://docs.rs/prost-build/latest/prost_build/struct.Config.html#method.bytes)), which makes it an interesting test case. This could be a simplification of the protobuf code and first step to scalars.

PR #1382, opened 2025-12-15
fix: Add back `DecodeError::new`

caspermeijn:
This function was meant for internal use only. Because of `doc(hidden)` it was publicly available and it is actually used by users. The prost 0.14.2 release broken some of the crate ecosystem. That release was yanked. To prevent more breakage, this PR adds the function back. This time as a public function that is documented to not be used.

The prost project intents to remove this function in the next breaking release.

### DIFF
--- a/prost/src/error.rs
+++ b/prost/src/error.rs
@@ -39,6 +39,7 @@ impl DecodeError {
         note = "This function was meant for internal use only. Because of `doc(hidden)` it was publicly available and it is actually used by users. The prost project intents to remove this function in the next breaking release."
     )]
     #[cold]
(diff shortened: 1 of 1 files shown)


jknovi:
This LGTM. Can you share the commit in which this was removed for more context?

jknovi (approved):

samuelcolvin:
It seems unfortunate if `DecodeError::new` becomes private and there's no alternative. Without being able to create `DecodeError`s, it's not really possible to implement the `Message` crate.

Is there a plan to provide a public interface for creawting `DecodeError`s?

PR #1385, opened 2026-01-04
fix some forgotten prost import paths

GlenDC:
For rama we make use of prost to generate code, however it's pretty inconvient
that a user would need to add also prost or prost-types to their direct dependencies
in their cargo.toml given that it means they have to worry about it being compatible
with what rama uses and that kind of bookkeeping work.

And while you already have an option to support a custom path for both
the `prost` and `prost-types` crate, it seems there were a couple of locations
where `::prost` was still hardcoded. Which makes me think that:

- either I do something wrong
- or that custom import path is not really verified/tested?

Anyhow, it's my first PR to this project so please do tell me if I'm off in the weeds with this one.

### DIFF
--- a/prost-build/src/code_generator.rs
+++ b/prost-build/src/code_generator.rs
@@ -573,13 +573,25 @@ impl<'b> CodeGenerator<'_, 'b> {
         ));
         self.append_field_attributes(fq_message_name, field.descriptor.name());
         self.push_indent();
-        self.buf.push_str(&format!(
-            "pub {}: {}<{}, {}>,\n",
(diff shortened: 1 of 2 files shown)


GlenDC:
FWIW, I am now declaring the prost dependencies from this exact branch, which unblocks me. So I can confirm that this was indeed the only missing pieces left to allow someone to be able to specify custom prost and prost-types import paths to allow you to run the project without having to add prost and prost-types as direct dependencies yourself.

For however much that verification is worth ofc.

PR #1387, opened 2026-01-08
Fix maps not preserving float `-0.0`.

aaronjeline:
Maps currently don't preserve `-0.0`. If you add `-0.0` as a value in a map, and read it, you will get `0.0`.

This isn't true of other structures. 

It happens because the default value is `0.0`, and when adding a new value, the code checks if it is equal to the default value using `PartialEq`, which for floats treats `0.0` and `-0.0` as equal.

### DIFF
--- a/prost-derive/src/lib.rs
+++ b/prost-derive/src/lib.rs
@@ -207,6 +207,8 @@ fn try_message(input: TokenStream) -> Result<TokenStream, Error> {
                 #default
             }
         }
+
+        impl #impl_generics #prost_path::encoding::IsDefault for #ident #ty_generics #where_clause {}
     };
     let expanded = if skip_debug {
         expanded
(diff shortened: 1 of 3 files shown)


caspermeijn (changes requested):
Sorry, but I don't want to extra complicate the library with this trait.

Rust chooses to do equality of floats this way. Even the library that cares about ordering [OrderedFloats](https://docs.rs/ordered-float/latest/ordered_float/struct.OrderedFloat.html) considers -0.0 and 0.0 equal.

I understand this could be crucial to your specific use case, but I feel like floats are not the right way to represent this data.

Maybe you could convince me to never skip values, if you can show that this opt-in is not too much code.

aaronjeline:
> Sorry, but I don't want to extra complicate the library with this trait.

That's totally fair, no worries if you don't think this is worth the complexity. Was fun to write regardless.

> I understand this could be crucial to your specific use case, but I feel like floats are not the right way to represent this data.

It's not, just thought it was a neat edge case.

PR #1391, opened 2026-01-19
Timestamp Arithmetic Operations (Part 1/3) of PR1307

lalvesl:
First of all, a huge thanks to all the contributors for this great library!

Splitting #1307 into smaller PRs as [requested](https://github.com/tokio-rs/prost/pull/1307#discussion_r2334458745) by @caspermeijn

### Changes
- Implements `Add<Duration>` and `Sub<Duration>` with saturation handling;
- Implements `Div` for primitive integers and floats;
- Add consts `Timestamp::MAX` and `Timestamp::MIN`;
- Includes unit tests and **Kani formal verification**.

### DIFF
--- a/benchmarks/Cargo.toml
+++ b/benchmarks/Cargo.toml
@@ -11,7 +11,7 @@ prost-build = { path = "../prost-build" }
 
 [dev-dependencies]
 prost = { path = "../prost" }
-criterion = { version = "0.7", default-features = false }
+criterion = { version = "0.8", default-features = false }
 
 [lib]
 # https://bheisler.github.io/criterion.rs/book/faq.html#cargo-bench-gives-unrecognized-option-errors-for-valid-command-line-options
--- a/prost-build/src/lib.rs
+++ b/prost-build/src/lib.rs
@@ -125,11 +125,12 @@
 //! ### Compiling `protoc` from source
 //!
 //! To compile `protoc` from source you can use the `protobuf-src` crate and
-//! set the correct environment variables.
+//! set the path to `protoc`.
 //! ```no_run,ignore, rust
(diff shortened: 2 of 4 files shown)


prost-types/src/timestamp.rs
```diff
@@ -112,6 +113,590 @@ impl Timestamp {
...
+                    Duration { seconds, nanos }
+                }
+            }
+        )*
+    };
+}
+
+impl_div_for_integer!(i8, u8, i16, u16, i32, u32, i64, u64, i128, u128);
+
+macro_rules! impl_div_for_float {
```
caspermeijn: I don't want to implement division. Primarily because `chrono` doesn't do that on its `DateTime` object. https://docs.rs/chrono/latest/chrono/struct.DateTime.html

prost-types/src/timestamp.rs
```diff
@@ -112,6 +113,590 @@ impl Timestamp {
...
+
+        kani::assume(ts.nanos <= Timestamp::MAX.nanos);
+        kani::assume(ts.nanos >= Timestamp::MIN.nanos);
+        kani::assume(dur.nanos <= Timestamp::MAX.nanos);
+        kani::assume(dur.nanos >= Timestamp::MIN.nanos);
+
+        let result = ts + dur;
+
+        assert!((Timestamp::MIN.nanos..=Timestamp::MAX.nanos).contains(&result.nanos));
+        assert!((Timestamp::MIN.seconds..=Timestamp::MAX.seconds).contains(&result.seconds));
```
caspermeijn: This doesn't actually test anything. It verifies that the numbers are between min and max of the underlaying type, which is always true.

I would verify the calculation by doing the calculation in `nanos: i128` as well.

PR #1398, opened 2026-02-10
add `ignore_deprecated_attribute` config to elide `deprecated` protobuf field attributes

tdyas:
Add an `ignore_deprecated_attribute` method to `prost_build::Config` to configure the code generator to ignore the `deprecated` field attribute.

The motivating use case is [a project](https://github.com/pantsbuild/pants/) dealing with a legacy protobuf specification where having to mark all call sites with `#[allow(deprecated)]` is too coarse grained of an approach for dealing with deprecated fields. (That project configures Clippy warnings as errors.)

Instead, it is more straightforward to configure the code generator to just ignore the protobuf `deprecated` field attribute on the fields where compatibility code still needs to access the deprecated fields.

This change is opt-in for users.


prost-build/src/code_generator.rs
```diff
@@ -665,7 +671,13 @@ impl<'b> CodeGenerator<'_, 'b> {
             self.append_doc(fq_message_name, Some(field.descriptor.name()));
             self.path.pop();
 
-            if self.deprecated(&field.descriptor) {
+            if self.deprecated(&field.descriptor)
+                && self
+                    .config()
+                    .ignore_deprecated_attributes
+                    .get_first_field(fq_message_name, field.descriptor.name())
+                    .is_none()
```
caspermeijn: This piece should move to inside `deprecated()`, so that it is only in one place.

prost-build/src/code_generator.rs
caspermeijn: This is a third place that should be ignored and tested.

prost-build/src/config.rs
```diff
@@ -375,6 +376,30 @@ impl Config {
         self
     }
 
+    /// Remove any `deprecated` field attribute from matching fields.
+    ///
+    /// # Arguments
+    ///
+    /// **`path`** - a path matching any number of fields. These fields will have any `deprecated` field attribute removed.
+    /// This is useful when you would like to make use of a `deprecated` protobuf field in your code without having to
+    /// mark all call sites as `#[allow(deprecated)]`.
```
caspermeijn: Most of this text belongs in the first alinea

caspermeijn (changes requested):
Normally, I don't like adding exceptions like this. But I understand how this could be useful. It helps that it is a small code change and you added tests.

The code needs some refinement, but I agree with the spirit of the PR.

prost-build/src/code_generator.rs
tdyas: Added logic to set `EnumVariantMapping.deprecated` based on the confiig. Tests also added.

prost-build/src/lib.rs
```diff
@@ -626,4 +626,44 @@ mod tests {
...
+            .ignore_deprecated_attribute(".")
+            .out_dir(tempdir.path())
+            .compile_protos(
+                &["src/fixtures/deprecated/all_deprecated.proto"],
+                &["src/fixtures/deprecated"],
+            )
+            .unwrap();
+
+        assert_eq_fixture_file!(
+            "src/fixtures/deprecated/_all_deprecated_ignore_deprecated.rs",
```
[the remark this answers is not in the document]
tdyas: I was able to come up with a test which uses assertions on the generated file. This allowed removing the extra fixtures.

prost-build/src/config.rs
tdyas: I edited the doc comment. Is it more in line with what you were thinking?

prost-build/src/code_generator.rs
tdyas: Done.

tdyas:
I'm not going to pursue this change any more since I found an alternative in just flagging code with `#[expect(deprecated)]`.

PR #1406, opened 2026-02-25
Introduce Encoding trait and use it for bytes

Guiguiprim:
Following this comment : https://github.com/tokio-rs/prost/pull/1380#pullrequestreview-3761479526

This extract the trait `CustomScalarInterface` from the previous MR and generalised it by renaming it to `ScalarEncoding` .
It is then used to rework the `Bytes` and `Vec<u8>` encoding.

On my fork, I also have two following branches: `scalar-interface-string` and `scalar-interface-numerical` which continue the migration work, by the end `prost-derive/src/field/scalar.rs` only generates  code the use the trait (In the mean time some code is duplicated). With all of this,  we should be in a good state to re-introduce the custom-scalar feature using the new `ScalarEncoding`.

### DIFF
--- a/prost/src/encoding.rs
+++ b/prost/src/encoding.rs
@@ -100,6 +100,41 @@ impl DecodeContext {
 pub const MIN_TAG: u32 = 1;
 pub const MAX_TAG: u32 = (1 << 29) - 1;
 
+pub trait Encoding {
+    type Type: Default;
+
+    fn encoded_len(tag: u32, value: &Self::Type) -> usize;
+    fn encode(tag: u32, value: &Self::Type, buf: &mut impl BufMut);
+    fn merge<B: Buf>(
+        wire_type: WireType,
+        value: &mut Self::Type,
+        buf: &mut B,
+        _ctx: DecodeContext,
+    ) -> Result<(), DecodeError>;
+
+    #[inline]
+    fn encoded_len_repeated(tag: u32, values: &[Self::Type]) -> usize {
+        values.iter().map(|v| Self::encoded_len(tag, v)).sum()
+    }
+    fn encode_repeated(tag: u32, values: &[Self::Type], buf: &mut impl BufMut) {
+        for value in values {
+            Self::encode(tag, value, buf);
+        }
+    }
+    fn merge_repeated<B: Buf>(
+        wire_type: WireType,
+        values: &mut Vec<Self::Type>,
+        buf: &mut B,
+        ctx: DecodeContext,
+    ) -> Result<(), DecodeError> {
+        check_wire_type(WireType::LengthDelimited, wire_type)?;
+        let mut value = Self::Type::default();
+        Self::merge(wire_type, &mut value, buf, ctx)?;
+        values.push(value);
+        Ok(())
+    }
+}
+
 /// Encodes a Protobuf field key, which consists of a wire type designator and
 /// the field tag.
 #[inline]
--- a/prost/src/encoding.rs
+++ b/prost/src/encoding.rs
@@ -817,6 +817,57 @@ pub mod bytes {
     }
 }
 
+pub struct VecU8Encoding;
+pub struct BytesEncoding;
+
+impl Encoding for VecU8Encoding {
+    type Type = Vec<u8>;
+
+    #[inline]
+    fn encoded_len(tag: u32, value: &Self::Type) -> usize {
+        bytes::encoded_len(tag, value)
+    }
+
+    #[inline]
+    fn encode(tag: u32, value: &Self::Type, buf: &mut impl BufMut) {
+        bytes::encode(tag, value, buf);
+    }
+
+    #[inline]
+    fn merge<B: Buf>(
+        wire_type: WireType,
+        value: &mut Self::Type,
+        buf: &mut B,
+        ctx: DecodeContext,
+    ) -> Result<(), DecodeError> {
+        bytes::merge_one_copy(wire_type, value, buf, ctx)
+    }
+}
+
+impl Encoding for BytesEncoding {
+    type Type = Bytes;
+
+    #[inline]
+    fn encoded_len(tag: u32, value: &Self::Type) -> usize {
+        bytes::encoded_len(tag, value)
+    }
+
+    #[inline]
+    fn encode(tag: u32, value: &Self::Type, buf: &mut impl BufMut) {
+        bytes::encode(tag, value, buf);
+    }
+
+    #[inline]
+    fn merge<B: Buf>(
+        wire_type: WireType,
+        value: &mut Self::Type,
+        buf: &mut B,
+        ctx: DecodeContext,
+    ) -> Result<(), DecodeError> {
+        bytes::merge(wire_type, value, buf, ctx)
+    }
+}
+
 pub mod message {
     use super::*;
 
--- a/prost-build/src/code_generator.rs
+++ b/prost-build/src/code_generator.rs
@@ -443,14 +443,6 @@ impl<'b> CodeGenerator<'_, 'b> {
         let type_tag = self.field_type_tag(&field.descriptor);
         self.buf.push_str(&type_tag);
 
-        if type_ == Type::Bytes {
-            let bytes_type = self
-                .context
-                .bytes_type(fq_message_name, field.descriptor.name());
-            self.buf
-                .push_str(&format!(" = {:?}", bytes_type.annotation()));
-        }
-
         match field.descriptor.label() {
             Label::Optional => {
                 if optional {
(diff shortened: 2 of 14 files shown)


Guiguiprim:
And yes, I realises this is a large PR, sorry for that :sweat_smile:

Guiguiprim:
I just realized that I was too focus on scalars.
This should be `Encoding` as I see no reason not to implement it for messages and groups and make this truly generic. For map we will need a `MapEncoding` but it will be able to use `Encoding` for the key and value...

Guiguiprim:
I updated the content of this PR:
- `ScalarEncoding` rename to more generic `Encoding`
- some quality of life improvement of  the code

Guiguiprim:
I now also have a `encoding-interface-message` branch that uses `Encoding` for messages.

Guiguiprim:
So far I only changed the internal of`prost-derive`, I haven't touched `prost-build`.

The original comment pointed to `prost_build::Config::bytes()` as a good study point.
From where I'm at, I think I have a good idea how we could change things in a generic and future-proof way.

The current behavior is:
```rust
    #[prost(bytes = "vec", tag = "7")]
    pub my_data: Vec<u8>,
// or
    #[prost(bytes = "bytes", tag = "7")]
    pub my_data: Bytes,
```

We could change it to (keeping the backward compat on the side):
```rust
    #[prost(bytes, tag = "7",  encoding = "VecU8Encoding")]
    pub my_data: Vec<u8>,
// or
    #[prost(bytes, tag = "7",  encoding = "BytesEncoding")]
    pub my_data: Bytes,
```

This could be easily extended to support custom types in the future:
```rust
    #[prost(bytes, tag = "7",  encoding = "MyBytesEncoding"), encoding_module = "my_crate::my_bytes"]
    pub my_data: MyBytes,
    #[prost(int32, tag = "8",  encoding = "MyIdEncoding"), encoding_module = "my_crate::my_id"]
    pub my_id: MyId,
    #[prost(string, tag = "9",  encoding = "MyStringEncoding"), encoding_module = "my_crate::my_string"]
    pub my_string: MyString,
    #[prost(string, tag = "10",  encoding = "MyMapEncoding"), encoding_module = "my_crate::my_map"]
    pub my_frist_map: MyMap<i32,  string>,
    #[prost(string, tag = "11",  value_encoding = "MyStringEncoding"), value_encoding_module = "my_crate::my_string"]
    pub my_string_map: HashMap<i32,  MyString>,
```

We would just need to see `encoding_module` as optional, if `None` the module is `prost::encoding`. In `prost-build`,  we could  have an API like `custom_encoding(&self, proto_type: Type, encoding_ty: &str, encoding_module: Option<&str>, paths: impl Interator<Item = &str>)`.

I'm going to try this method.

Guiguiprim:
I have done it  in a new branch `encoding-interface-prost-build`,  I will add its content in this PR once I rebased all my other branches (to make sure it holds).

prost/src/encoding/bytes.rs
```diff
@@ -0,0 +1,159 @@
...
+impl Encoding for VecU8Encoding {
+    type Type = Vec<u8>;
+
+    #[inline]
+    fn encoded_len(tag: u32, value: &Self::Type) -> usize {
+        key_len(tag) + encoded_len_varint(value.len() as u64) + value.len()
+    }
+
+    fn encode(tag: u32, value: &Self::Type, buf: &mut impl BufMut) {
+        encode_key(tag, WireType::LengthDelimited, buf);
```
caspermeijn: I would like the bytes related code in `prost::encoding` to remain in place. This function body can simply be a single call to `prost::encoding::bytes::encode()`.

 I have two reasons for that:
- To minimize the diff. Reviewing code moves is a lot of work for me, as I want to be sure no code is changed in the process.
- `prost::encoding` is doc hidden, but used by some dependencies. I prefer to have a complete new API ready before removing the old functions in the module. I prefer not to break those dependencies. You should assume you can't make breaking changes to the encoding module.

prost/src/encoding.rs
```diff
@@ -100,6 +103,48 @@ impl DecodeContext {
 pub const MIN_TAG: u32 = 1;
 pub const MAX_TAG: u32 = (1 << 29) - 1;
 
+pub trait Encoding {
```
caspermeijn, line 103: Should this trait also include `encode_packed` and `encoded_len_packed`? Or is that a different encoding variant?

prost/src/encoding.rs
```diff
@@ -100,6 +103,48 @@ impl DecodeContext {
...
+    ) -> Result<(), DecodeError> {
+        check_wire_type(WireType::LengthDelimited, wire_type)?;
+        let mut value = Self::Type::default();
+        Self::merge(wire_type, &mut value, buf, ctx)?;
+        values.push(value);
+        Ok(())
+    }
+}
+
+pub trait EncodingDefault: Encoding {
```
caspermeijn: Why is this trait split from the `Encoding` trait? 

Can all the default related changes be split into a separate PR? 

The existing code base doesn't have any function called `is_default`. Is this a behavior change of prost separate of the `Encoding` trait?

prost/src/encoding.rs
```diff
@@ -592,7 +637,7 @@ pub mod string {
             }
 
             let drop_guard = DropGuard(value.as_mut_vec());
-            bytes::merge_one_copy(wire_type, drop_guard.0, buf, ctx)?;
+            VecU8Encoding::merge(wire_type, drop_guard.0, buf, ctx)?;
```
caspermeijn: Is this a change in copy behavior? `merge_one_copy` seems to do something special.

caspermeijn (changes requested):

I have wanted this change for some time, but never got around to it. I think the trait will make encoding so much more extendable.

I am a bit worried about user of `prost-derive` directly. Please make sure that this is not a breaking change for them.

Can you explain why `prost-build` didn't have to change?

I skimmed your code. This is impressive work, but it takes a lot of time to review. Smaller PRs make that easier.

Guiguiprim:
Sorry I go a bit carried away. I'm going to simplify this PR.

## Design

As a said in other comments, I went ahead and implemented the trait for `String`,  numerical types and `Message` in other branches. Where I'm at, I have 3 traits and a vision for a fourth:

### Traits

```rust
ub trait Encoding {
    type Type: Default;

    fn encoded_len(tag: u32, value: &Self::Type) -> usize;
    fn encode(tag: u32, value: &Self::Type, buf: &mut impl BufMut);
    fn merge<B: Buf>(
        wire_type: WireType,
        value: &mut Self::Type,
        buf: &mut B,
        _ctx: DecodeContext,
    ) -> Result<(), DecodeError>;

    #[inline]
    fn encoded_len_repeated(tag: u32, values: &[Self::Type]) -> usize {...}
    fn encode_repeated(tag: u32, values: &[Self::Type], buf: &mut impl BufMut) {...}
    fn merge_repeated<B: Buf>(
        wire_type: WireType,
        values: &mut Vec<Self::Type>,
        buf: &mut B,
        ctx: DecodeContext,
    ) -> Result<(), DecodeError> {...}
}

pub trait EncodingDefault: Encoding {
    type RefType<'x>;
    fn is_default(value: &Self::Type) -> bool;
    fn get<'x>(value: &'x Option<Self::Type>) -> Self::RefType<'x>;
}

pub trait PackedEncoding: Encoding {
    fn encoded_len_packed(tag: u32, values: &[Self::Type]) -> usize;
    fn encode_packed(tag: u32, values: &[Self::Type], buf: &mut impl BufMut);
}

pub trait MapEncoding<Key, Value> {
  type Type: Default;
  // TODO
}
``` 

- `PackedEncoding` is only implemented for numerical types, as it is only supported there
- `EncodingDefault` would not make sense for `Message` and groups, anyway as  you pointed out this is more a feature addition (a relict of my custom scalar need)

#### Prost-build

For the `prost-build` part, I did implement this, but did not include it yet into the PR.

```rust
    #[prost(bytes, tag = "7",  encoding = "MyBytesEncoding"), encoding_module = "my_crate::my_bytes"]
    pub my_data: MyBytes,
```

I did keep the backward compatibility with the current syntax, in order not to be breaking.

## next iter

I'm going to simplify this PR
- I will keep all existing code in place
- I will implement the trait using the existing methods
- I will  include the `prost-build` code (that was the original goal :sweat: )
- I will remove`EncodingDefault` for the time being

prost/src/encoding.rs
Guiguiprim: It is not as I implemented `VecU8Encoding::merge` as `merge_one_copy` as it was the most optimise way to do it.

Guiguiprim:
Done !

I now  need  to rework all my others branches :p

prost/src/encoding.rs
Guiguiprim: Removed

prost/src/encoding.rs
Guiguiprim: This will come with a dedicated trait.

prost/src/encoding/bytes.rs
Guiguiprim: Done

Guiguiprim:
I added tests for the new `prost-derive` attributes, with retro-compat tests.

Guiguiprim:
I rebased my other work, I now have in my fork the branches:
- `encoding-string`: adds `Encoding` for `String` (and support for map  key with encoding in `prost-build`)
- `encoding-numerical`: adds `Encoding` for all numerical types (and use it for enum, and simplify `procst-build`)
- `encoding-message`: adds `Encoding` for `Message` (with support in maps)

magnet:
Interested in this feature and what it enables as well! @caspermeijn would you have time to review?

Issue #1373, opened 2025-11-28
Feature request: option to avoid embedding message/field names as plaintext in prost-derive (Message derive)


h3zh1:
Hi, and thanks for the great library!

I've noticed that when using:

``` rust
#[derive(Clone, PartialEq, Eq, Hash, ::prost::Message)]
pub struct Foo { ... }
```

the generated code from `prost-derive` causes both the message name and
its field names to be embedded as plaintext in the compiled binary.

Looking at `prost-derive` (currently 0.14.1), in `try_message` the
derive macro generates:

``` rust
const STRUCT_NAME: &'static str = stringify!(#ident);
```

and later in `merge_field`:

``` rust
error.push(STRUCT_NAME, stringify!(#field_ident));
```

All `stringify!(...)` calls produce `&'static str` constants, which end
up in the final binary.

-   It leaks internal schema names such as `user_password`,
    `internal_token`, etc.
-   It eases reverse engineering.

PR #1407, opened 2026-02-26
feat: add skip_field_names attribute to strip struct/field names

h3zh1:
prost-derive's Message trait impl embeds struct and field names via stringify!() in STRUCT_NAME constants and merge_field error paths. These survive even when skip_debug is enabled, leaking type information into compiled binaries.

Add a new #[prost(skip_field_names)] attribute (independent of skip_debug) that when set:
- Removes the STRUCT_NAME constant generation
- Removes .map_err() error context from merge_field
- Replaces unreachable!() messages in oneof merge with no-arg version

prost-build: add Config::skip_field_names() API, context query method, and code_generator attribute emission (mirroring skip_debug pattern).

### DIFF
--- a/prost-build/src/config.rs
+++ b/prost-build/src/config.rs
@@ -48,6 +48,7 @@ pub struct Config {
     pub(crate) protoc_executable: PathBuf,
     pub(crate) disable_comments: PathMap<()>,
     pub(crate) skip_debug: PathMap<()>,
+    pub(crate) skip_field_names: PathMap<()>,
     pub(crate) skip_protoc_run: bool,
     pub(crate) skip_source_info: bool,
     pub(crate) include_file: Option<PathBuf>,
@@ -433,6 +434,24 @@ impl Config {
         self.skip_debug.clear();
         for matcher in paths {
             self.skip_debug.insert(matcher.as_ref().to_string(), ());
+        }
+        self
+    }
+
+    /// Skips generating field names in `Message` trait implementations.
+    ///
+    /// When enabled, `stringify!()` calls that embed struct and field names
+    /// into the binary are removed. Useful for security-sensitive applications
+    /// where leaking type information into binaries is undesirable.
+    pub fn skip_field_names<I, S>(&mut self, paths: I) -> &mut Self
+    where
+        I: IntoIterator<Item = S>,
+        S: AsRef<str>,
+    {
+        self.skip_field_names.clear();
+        for matcher in paths {
+            self.skip_field_names
+                .insert(matcher.as_ref().to_string(), ());
         }
         self
     }
@@ -1214,6 +1233,7 @@ impl default::Default for Config {
             protoc_executable: protoc_from_env(),
             disable_comments: PathMap::default(),
             skip_debug: PathMap::default(),
+            skip_field_names: PathMap::default(),
             skip_protoc_run: false,
             skip_source_info: false,
             include_file: None,
@@ -1244,6 +1264,7 @@ impl fmt::Debug for Config {
             .field("protoc_args", &self.protoc_args)
             .field("disable_comments", &self.disable_comments)
             .field("skip_debug", &self.skip_debug)
+            .field("skip_field_names", &self.skip_field_names)
             .field("prost_path", &self.prost_path)
             .finish()
     }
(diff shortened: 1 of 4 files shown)


caspermeijn:

This PR adds additional code path that need to be tested. The project doesn't have enough maintainers to support additional code paths.

Can you explain why this is important? I don't understand what problem is fixed by this additional attribute.

h3zh1:
This PR implements the feature request in https://github.com/tokio-rs/prost/issues/1373: add an option to avoid embedding message/field names as plaintext in the binary via `stringify!()`.

Even with `#[prost(skip_debug)]` enabled, the generated code still includes:
- `const STRUCT_NAME: &'static str = stringify!(#ident);` → embeds the message name
- `error.push(STRUCT_NAME, stringify!(#field_ident));` in `merge_field` error paths → embeds field names

This leaks internal schema details (e.g., sensitive names like `user_password` or `internal_token`) into the binary strings table, making reverse engineering easier.

With the new `#[prost(skip_field_names)]` attribute (or globally via `Config::skip_field_names(["."])` to apply to all types), these remaining strings are stripped:
- No `STRUCT_NAME` constant is generated
- `merge_field` error paths lose the field name context
- `unreachable!()` in oneof merge paths uses the no-message version

Example proto:
```proto
syntax = "proto3";
package example;

message MyMessage {
    string name = 1;
    int32 id = 2;
    oneof test_oneof {
        string field_a = 3;
        int32 field_b = 4;
    }
}
```
In build.rs:
```
prost_build::Config::new()
    .skip_field_names(["."])  // Applies to root (all messages in the file/package)
    .compile_protos(&["proto/example.proto"], &["proto/"])?;
```
With this configuration:

Message names (e.g., "MyMessage") and field names (e.g., "name", "id", "field_a", "field_b") are no longer embedded as plaintext constants or in error strings.

image

bluec0re:
I also just looked into this to reduce the number of internal data leaks, so would love to see this merged.

benma:
https://github.com/BitBoxSwiss/bitbox02-firmware/ makes use of prost and this PR would be very helpful to us in reducing the embedded firmware binary size significantly.

prost-build/src/config.rs
```diff
@@ -437,6 +438,24 @@ impl Config {
         self
     }
 
+    /// Skips generating field names in `Message` trait implementations.
+    ///
+    /// When enabled, `stringify!()` calls that embed struct and field names
+    /// into the binary are removed. Useful for security-sensitive applications
+    /// where leaking type information into binaries is undesirable.
```
caspermeijn, line 445: I  think the security through obscurity argument is weak. I prefer to focus on the reduced file size.

prost-derive/src/lib.rs
```diff
@@ -113,18 +114,27 @@ fn try_message(input: TokenStream) -> Result<TokenStream, Error> {
...
+            quote! {
+                #(#tags)* => {
+                    let mut value = &mut self.#field_ident;
+                    #merge
+                },
+            }
+        } else {
+            quote! {
+                #(#tags)* => {
+                    let mut value = &mut self.#field_ident;
```
caspermeijn, line 127: Please prevent this code duplication.

prost-derive/src/lib.rs
```diff
@@ -479,7 +496,7 @@ fn try_oneof(input: TokenStream) -> Result<TokenStream, Error> {
             {
                 match tag {
                     #(#merge,)*
-                    _ => unreachable!(concat!("invalid ", stringify!(#ident), " tag: {}"), tag),
```
caspermeijn, line 482: Can we convince the compiler this path is actually unreachable? Then this line would be optimized away.

prost-build/src/config.rs
```diff
@@ -437,6 +438,24 @@ impl Config {
         self
     }
 
+    /// Skips generating field names in `Message` trait implementations.
```
caspermeijn, line 441: This should explain more of that the user visible result is of the option. If I understand correctly this removes the field names from decode error message. I think the documentation should reflect that.

caspermeijn (changes requested):
I better understand why this is useful. I think the security argument is weak, but I believe the reduced executable size is useful.

PR #1408, opened 2026-03-01
feat: Add optional custom string type support

nu11ptr:
Adds custom string type support in a similar way as `btree_map` and `bytes` `Config` methods, but this allows the user to specify the custom string type. The type must implement `prost::Message` ([example from FlexStr](https://github.com/nu11ptr/flexstr/blob/master/src/str.rs#L131)) and be wire compatible with a raw UTF8 protbuf string. `prost-build` generates a new `custom_string` type in the prost annotation which `prost` handles via custom encoder (almost entirely based off the the existing message functions). Tests added to ensure it works properly. I'm using this in a project currently (that will soon be production), and while it is early, the implementation seems to be working properly. I did my best to ensure the change was as minimal as possible. Let me know if there are any questions.


caspermeijn:

I can't accept this PR, because it adds additional code path that need to be tested. The project doesn't have enough maintainers to support this.

If you are really interested in this feature, you should work on custom scalars. That is a more maintainable way to add this kind of customization to prost. This comment has my latest insights on that: https://github.com/tokio-rs/prost/pull/1380#pullrequestreview-3761479526

nu11ptr:
> it adds additional code path that need to be tested

There is very little code added. Perhaps a 150-200 LoC tops. I was careful to do this in a way that had very little impact on anything outside the feature at hand. The rest is test related, so the new paths are tested.

> If you are really interested in this feature, you should work on custom scalars

I'm not against looking at that, but it seems like a much bigger feature. This just add custom string support which is arguably going to be the thing most will want to swap (the others likely being `Vec` and `HashMap` for something heapless), but takes no opinion on which one, immediately opening the gate for any string crates that want to implement `prost::Message`.

PR #1419, opened 2026-03-29
fix(prost-build): use Display instead of Debug for quoted string

### LINKED ISSUE #1418: prost-build generates invalid code when combined with -Zfmt-debug=none
The [enumeration attribute](https://github.com/tokio-rs/prost/blob/822b19b95588b194ed5c9f0e43dd95e5f256b984/prost-build/src/code_generator.rs#L1067) uses `{:?}` to format the type path, resulting in `enumeration = ,` code when debug output is elided.

I think this can be fixed by using `enumeration = "{}"` instead, but I'm not sure if there is the possibility of unprintable characters or quotes here. I don't think so as this is using idents? Also the [code just below](https://github.com/tokio-rs/prost/blob/822b19b95588b194ed5c9f0e43dd95e5f256b984/prost-build/src/code_generator.rs#L1076-L1077) doesn't care about these theoretical edge cases and uses `{}` instead. So either it is safe to use `{}` (and `{:?}` only really adds the quotes) or both needs to be fixed.

bluec0re:
Prevents the generation of broken code in case the fmt-debug unstable feature is used.

Fixes #1418

### DIFF
--- a/prost-build/src/code_generator.rs
+++ b/prost-build/src/code_generator.rs
@@ -1064,7 +1064,7 @@ impl<'b> CodeGenerator<'_, 'b> {
             Type::Group => Cow::Borrowed("group"),
             Type::Message => Cow::Borrowed("message"),
             Type::Enum => Cow::Owned(format!(
-                "enumeration = {:?}",
+                "enumeration = \"{}\"",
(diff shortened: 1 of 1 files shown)


caspermeijn (approved):

PR #1423, opened 2026-04-25
prost-build: add Config::eq_when_safe to gate auto Eq derives (#30)

### LINKED ISSUE #30: Implement 'Eq' trait for proto
Is there a way to specify custom trait implementations, such as `Eq`, in the proto file for prost-build to generate? 

If not, what is the correct way to implement a trait for the generated Rust code?

Thanks!

MukundaKatta:
## Summary
Closes #30. Adds an explicit `Config::eq_when_safe(bool)` builder so callers can opt out of automatic `Eq`/`Hash` derives. Default `true` preserves existing behavior.

## What's added
- `prost-build/src/config.rs` — new `eq_when_safe` field + builder method.
- `prost-build/src/context.rs` — gates `can_message_derive_eq` / `can_field_derive_eq` on the flag, with `RefCell<HashMap>` memoization and a cycle-break sentinel for recursive types.
- `prost-build/src/lib.rs` — unit tests for both flag states.
- New `tests/src/derive_eq.proto` and `tests/src/derive_eq.rs` with compile-time `TestEqIsImplemented: Eq` bound across all-int, float, double, composed-safe, and composed-unsafe messages.

## Test plan
- `cargo fmt --all -- --check` clean.
- `cargo clippy -p prost-build --all-targets --all-features -- -D warnings` clean.
- 539/539 tests pass across `prost`, `prost-build`, `prost-derive`, `prost-types`, `tests`.

### DIFF
--- a/prost-build/src/config.rs
+++ b/prost-build/src/config.rs
@@ -53,6 +53,7 @@ pub struct Config {
     pub(crate) include_file: Option<PathBuf>,
     pub(crate) prost_path: Option<String>,
     pub(crate) prost_types_path: Option<String>,
+    pub(crate) eq_when_safe: bool,
     #[cfg(feature = "format")]
     pub(crate) fmt: bool,
 }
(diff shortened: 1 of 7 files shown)


prost-build/src/config.rs
```diff
@@ -375,6 +376,32 @@ impl Config {
...
+    /// these traits manually if desired.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// # let mut config = prost_build::Config::new();
+    /// // Opt out of automatic `Eq` derives.
+    /// config.eq_when_safe(false);
+    /// ```
+    pub fn eq_when_safe(&mut self, enable: bool) -> &mut Self {
```
caspermeijn, line 400: If we want this config API, it should use accept paths, like https://docs.rs/prost-build/latest/prost_build/struct.Config.html#method.btree_map

That way it can be enabled or disabled on specific messages

tests/src/derive_eq.proto
```diff
@@ -0,0 +1,36 @@
+syntax = "proto3";
```
caspermeijn, line 1: I don't understand this diff. The `master` branch already contains the file: `tests/src/derive_eq.proto`

caspermeijn (changes requested):
What I don't understand is: Why do you want to disable `Eq` when it is safe? The trait is useful, so why remove it?

PR #1425, opened 2026-05-13
fix(prost-derive): avoid UTF-8 panic for enumeration names

### LINKED ISSUE #1417: `prost::Message` derive panics on non-ASCII invalid `map` type strings in `#[prost(...)]`
## Summary

When `#[prost(map = "...", ...)]` contains an invalid non-ASCII key or value type string, `prost::Message` derive can panic on UTF-8 byte slicing inside `scalar::Ty::from_str` instead of taking the normal validation/error path.

**Disclaimer**: this issue is probably being too picky on where the error should be reported. If this is the case, please feel free to close the issue.

## Reproduction

```rust
pub struct BadMap {
    #[prost(map = "éééééé, int32", tag = "1")]
    pub values: ::std::collections::HashMap<i32, i32>,
}
```

Then run:

```text
cargo check
```

Compilation fails during macro expansion with a raw UTF-8 boundary panic, e.g.:

```text
error: proc-macro derive panicked
  = help: message: byte index 11 is not a char boundary; it is inside 'é' (bytes 10..12) of `éééééé`
```

The panic comes from the `enumeration` prefix probe in `prost-derive/src/field/scalar.rs`:

```rust
s if s.len() > enumeration_len && &s[..enumeration_len] == "enumeration" => {
```

For `s = "éééééé"`, `s.len() == 12`, so `&s[..11]` panics before `Err(anyhow!("invalid type: {s}"))` can be returned.

`prost-derive/src/field/map.rs` makes this reachable from public `#[prost(map = ...)]` attributes through both of these paths:

```rust
let ty = scalar::Ty::from_str(s)?;
```

and

```rust
if let Ok(ty) = scalar::Ty::from_str(s) {
```

So both map key parsing and map value parsing inherit the same failure mode.

## Expected behavior

Invalid `map` type strings should take the same validation/error path as other bad type strings.

Given the current top-level `unwrap()` in the derive entrypoint, that may still surface as a proc-macro panic today, but it should be the usual contextual one (for example, including `invalid map key type` / `invalid map value type` and `invalid message field ...`), not a raw UTF-8 char-boundary panic from inside `Ty::from_str`.

SAY-5:
The enumeration prefix probe in `Ty::from_str` slices the input by byte index and panics on non-ASCII strings whose byte length exceeds `"enumeration".len()` but whose 11th byte is mid-codepoint (e.g. `"éééééé"` from a `#[prost(map = ...)]` attribute). Adding an `is_char_boundary` guard lets such inputs fall through to the existing `invalid type` error path.

Closes #1417

### DIFF
--- a/prost-derive/src/field/scalar.rs
+++ b/prost-derive/src/field/scalar.rs
@@ -483,8 +483,8 @@ impl Ty {
             "bool" => Ty::Bool,
             "string" => Ty::String,
             "bytes" => Ty::Bytes(BytesTy::Vec),
-            s if let Some(s) = s.strip_prefix("enumeration") => {
-                let s = s.trim();
+            s if s.starts_with("enumeration") => {
+                let s = s.strip_prefix("enumeration").unwrap().trim();
                 match s.chars().next() {
                     Some('<') | Some('(') => (),
                     _ => return error,
--- a/prost-derive/src/field/scalar.rs
+++ b/prost-derive/src/field/scalar.rs
@@ -466,7 +466,6 @@ impl Ty {
     }
 
     pub fn from_str(s: &str) -> Result<Ty, Error> {
-        let enumeration_len = "enumeration".len();
         let error = Err(anyhow!("invalid type: {s}"));
         let ty = match s.trim() {
             "float" => Ty::Float,
@@ -484,11 +483,8 @@ impl Ty {
             "bool" => Ty::Bool,
             "string" => Ty::String,
             "bytes" => Ty::Bytes(BytesTy::Vec),
-            s if s.len() > enumeration_len
-                && s.is_char_boundary(enumeration_len)
-                && &s[..enumeration_len] == "enumeration" =>
-            {
-                let s = &s[enumeration_len..].trim();
+            s if let Some(s) = s.strip_prefix("enumeration") => {
+                let s = s.trim();
                 match s.chars().next() {
                     Some('<') | Some('(') => (),
                     _ => return error,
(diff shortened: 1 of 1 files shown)


SAY-5:
The msrv job fails due to indexmap 2.14.0 requiring a newer Rust toolchain than the repo's MSRV (1.82), not this change. The same failure appears on the maint branch. All other CI checks pass.

prost-derive/src/field/scalar.rs
```diff
@@ -484,7 +484,10 @@ impl Ty {
             "bool" => Ty::Bool,
             "string" => Ty::String,
             "bytes" => Ty::Bytes(BytesTy::Vec),
-            s if s.len() > enumeration_len && &s[..enumeration_len] == "enumeration" => {
+            s if s.len() > enumeration_len
+                && s.is_char_boundary(enumeration_len)
+                && &s[..enumeration_len] == "enumeration" =>
+            {
                 let s = &s[enumeration_len..].trim();
```
caspermeijn: This seems to be a manual implementation of `starts_with`. Would it be possible to replace this with a `strip_prefix` call? I have not verified this suggestion.

```suggestion
            s if let Some(s) = s.strip_prefix("enumeration") => {
                let s = s.trim();
```

caspermeijn (changes requested):
Thank you for this work. I have a suggestion to make this even better. I will look into the build failure.

caspermeijn (approved):

caspermeijn (approved):

PR #1426, opened 2026-05-13
fix(prost-derive): return error for invalid enumeration default identifiers

### LINKED ISSUE #1416: `prost::Message` derive panics early on invalid enum `default` identifiers in `#[prost(...)]`
## Summary

When a field uses `#[prost(enumeration = "...", default = "...", ...)]` and the `default` string is not a valid Rust identifier, `prost::Message` derive panics inside `proc_macro2::Ident::new` instead of turning the bad value into a regular parse/validation error.

This is reachable from handwritten public `#[prost(...)]` attributes, not only from generated code.

**NOTE**: `prost`'s derive entrypoints currently `unwrap()` their internal `Result`, so many validation failures still end up reported as proc-macro panics at the top level. Even so, this case is still a bug (without much of a real consequence though) because it panics *earlier* than the normal validation path and loses the field-specific context that other `from_lit` errors would carry.

## Reproduction

```rust
#[derive(Clone, Copy, Debug, PartialEq, Eq, ::prost::Enumeration)]
#[repr(i32)]
pub enum ExampleEnum {
    Zero = 0,
}

pub struct BadDefault {
    #[prost(enumeration = "ExampleEnum", default = "not-valid!", tag = "1")]
    pub value: i32,
}
```

Then run:

```text
cargo check
```

SAY-5:
When a `#[prost(enumeration = "...", default = "...")]` field has a `default` string that is not a valid Rust identifier, `DefaultValue::from_lit` calls `Ident::new` which panics. Parsing via `parse_str::<Ident>` instead surfaces the same condition through the regular `Result` path and preserves the field-level context that other `from_lit` errors already carry.

Closes #1416

### DIFF
--- a/prost-derive/src/field/scalar.rs
+++ b/prost-derive/src/field/scalar.rs
@@ -672,7 +672,7 @@ impl DefaultValue {
                 let value = value.trim();
 
                 if let Ty::Enumeration(ref path) = *ty {
-                    let variant = Ident::new(value, Span::call_site());
+                    let variant = parse_str::<Ident>(value)?;
                     return Ok(DefaultValue::Enumeration(quote!(#path::#variant)));
                 }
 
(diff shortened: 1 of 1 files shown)


caspermeijn (approved):

PR #1440, opened 2026-06-24
fix: escape raw proto enum value names in generated str helpers

hibrian827:
The generated as_str_name() and from_str_name() helpers appended the raw EnumValueDescriptorProto.name directly between Rust string-literal quotes, treating descriptor metadata as trusted source text. With Config::compile_fds() accepting caller-supplied FileDescriptorSets (not re-validated by protoc), a crafted enum value name could break out of the string literal and tamper with the generated Rust source.

### DIFF
--- a/prost-build/src/code_generator.rs
+++ b/prost-build/src/code_generator.rs
@@ -43,6 +43,16 @@ fn push_indent(buf: &mut String, depth: u8) {
     }
 }
 
+fn push_rust_string_literal_body(buf: &mut String, value: &str) {
+    const HEX: &[u8; 16] = b"0123456789abcdef";
+
+    for b in value.bytes() {
+        buf.push_str("\\x");
+        buf.push(HEX[(b >> 4) as usize] as char);
+        buf.push(HEX[(b & 0x0f) as usize] as char);
+    }
+}
+
 struct Field {
     descriptor: FieldDescriptorProto,
     path_index: i32,
@@ -841,7 +851,7 @@ impl<'b> CodeGenerator<'_, 'b> {
             self.buf.push_str("Self::");
             self.buf.push_str(&variant.generated_variant_name);
             self.buf.push_str(" => \"");
-            self.buf.push_str(variant.proto_name);
+            push_rust_string_literal_body(self.buf, variant.proto_name);
             self.buf.push_str("\",\n");
         }
 
@@ -869,7 +879,7 @@ impl<'b> CodeGenerator<'_, 'b> {
         for variant in variant_mappings.iter() {
             self.push_indent();
             self.buf.push('\"');
-            self.buf.push_str(variant.proto_name);
+            push_rust_string_literal_body(self.buf, variant.proto_name);
             self.buf.push_str("\" => Some(");
             if variant.deprecated {
                 self.buf.push_str("#[allow(deprecated)] ");


caspermeijn (changes requested):
I understand the problem, but don't agree with the solution.

1) This destroys the readability of the generated code.
2) This PR will fix this problem for enum variant names, but I assume this problem also exists for attribute and message names.

I suggest validating the names and panic when unexpected characters are detected. Please check whether `protoc` already does some checking of identifiers, if not, `prost-build` can do validation before generation.

hibrian827:
Yes, but only on the path that parses .proto source. 
```
enum E { ZERO = 0; BAD" = 1; }     → bad_value.proto:4:12: String literals cannot cross line boundaries.
message M { int32 foo-bar = 1; }   → bad_ident2.proto:3:12: Missing field number.
```
protoc's tokenizer enforces [A-Za-z_][A-Za-z0-9_]* at parse time, so any FileDescriptorSet protoc emits is guaranteed well-formed.

But prost-build doesn't always parse source. It consumes a compiled FileDescriptorSet, and it can be handed one that never went through protoc's parser:

- `Config::skip_protoc_run()` + `file_descriptor_set_path(...)` : loads a precompiled .fds
- `Config::generate(requests)` : takes FileDescriptorProtos directly (hand-built, or from another tool / prost-reflect)

On those paths protoc's check never runs, and prost-build currently trusts the names verbatim. So the validation has to live in prost-build.

How about adding two tiny predicates in `ident.rs`:
```
/// Valid Protobuf identifier: non-empty `[A-Za-z_][A-Za-z0-9_]*`.
pub fn is_valid_ident(s: &str) -> bool {
    let mut chars = s.chars();
    matches!(chars.next(), Some(c) if c == '_' || c.is_ascii_alphabetic())
        && chars.all(|c| c == '_' || c.is_ascii_alphanumeric())
}

/// Package name: `.`-separated identifiers (empty = unset).
pub fn is_valid_package(s: &str) -> bool {
    s.is_empty() || s.split('.').all(is_valid_ident)
}
```

One pass at the top of CodeGenerator::generate, before any name is emitted:
```
pub(crate) fn generate(context: ..., file: FileDescriptorProto, buf: ...) {
    validate_proto_identifiers(&file);   // panics on the first bad name
    ...
}

fn check_ident(kind: &str, name: &str) {
    if !is_valid_ident(name) {
        panic!("{kind} name {name:?} is not a valid Protobuf identifier \
                (expected `[A-Za-z_][A-Za-z0-9_]*`)");
    }
}

fn validate_proto_identifiers(file: &FileDescriptorProto) {
    assert!(is_valid_package(file.package()), "invalid package {:?}", file.package());
    for m in &file.message_type { validate_message(m); }
    for e in &file.enum_type    { validate_enum(e); }
    for s in &file.service {
        check_ident("service", s.name());
        for method in &s.method { check_ident("method", method.name()); }
    }
}

fn validate_message(m: &DescriptorProto) {
    check_ident("message", m.name());
    for f in &m.field       { check_ident("field", f.name()); }
    for o in &m.oneof_decl  { check_ident("oneof", o.name()); }
    for n in &m.nested_type { validate_message(n); }   // recurse
    for e in &m.enum_type   { validate_enum(e); }
}

fn validate_enum(e: &EnumDescriptorProto) {
    check_ident("enum", e.name());
    for v in &e.value { check_ident("enum value", v.name()); }
}
```

caspermeijn:
Yes, such a validation is what I was thinking of.

That validation needs significant testing before accepted. I am thinking of a fuzzer that confirms that all used fields are validated. Or copying the whole structure where only validated fields are copied, which ensures all used fields are validated. This is important for maintainability. I want the compiler or test suite to yell when the promise is broken.

At that point I ask: Is that worth the effort? What are you trying to defend against? 

Both `prost-build` and `prost-derive` are executed by the compiler through a `build.rs`. This kind of build dependency can do all kind of evil stuff on the developer machine.

The attack path I see is: sneeking code into the runtime binary, without it being noticed in source code. But if you are worried about that, why are you loading an binary using `file_descriptor_set_path`?

hibrian827:
You are right, I think this matter is getting a bit out of hands. My original report was assessed as a simple bug and was suggested to file a PR probably for robustness fix. But adding codes for it while maintaining the original functionality seems a bit challenging. My last suggestion is, would you just like me to add some instructions of the trust model as a comment to the source code -  sometign like below?
```
/// In combination with `file_descriptor_set_path`, this can be used to provide a file
/// descriptor set as an input file, rather than having prost-build generate the file by calling
/// protoc.
///
/// # Trust
///
/// When protoc is skipped, the supplied `FileDescriptorSet` is used as-is and is **trusted**.
/// Unlike `.proto` sources parsed by protoc, its contents are not validated before 
/// being used to generate Rust code. Only supply a descriptor set from a source you trust; 
/// a malformed or hostile one may produce Rust that fails to compile, or that 
/// does not correspond to the apparent schema.
```

PR #1441, opened 2026-07-02
fix: Duration Display picks the wrong sign and overflows at i64::MIN

chatman-media:
Display was reading the sign off the raw (unnormalized) seconds/nanos instead of the normalized value, so something like `Duration { seconds: 1, nanos: -1 }` — which normalizes to a positive 0.999999999s — printed as `-0.999999999s`. On top of that, both Display and `TryFrom<Duration> for time::Duration` call signed `.abs()`/negation on i64, which panics once seconds lands on i64::MIN (or silently misbehaves in release builds).

Swapped those to `unsigned_abs()` and made Display derive the sign from the normalized value. Added a couple of tests for the sign flip and the i64::MIN edge cases on both paths.

### DIFF
--- a/prost-types/src/duration.rs
+++ b/prost-types/src/duration.rs
@@ -98,8 +98,8 @@ impl TryFrom<Duration> for time::Duration {
             ))
         } else {
             Err(DurationError::NegativeDuration(time::Duration::new(
-                (-duration.seconds) as u64,
-                (-duration.nanos) as u32,
+                duration.seconds.unsigned_abs(),
+                duration.nanos.unsigned_abs(),
             )))
         }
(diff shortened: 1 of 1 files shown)


prost-types/src/duration.rs
```diff
@@ -386,6 +386,60 @@ mod tests {
         );
     }
 
+    #[cfg(feature = "std")]
```
caspermeijn, line 389: std is not required for display. Please remove feature requirement

caspermeijn (changes requested):
Thank you for this fix including tests. The code looks good. I have a small nit pick.

PR #1452, opened 2026-08-05
fix(prost-derive): support unit structs in Message derive

### LINKED ISSUE #1424: Struct with no body fails compilation
```rust
// This works
#[derive(::prost::Message)]
struct Foo {}

// This fails
#[derive(::prost::Message)]
struct Bar;
```

The second variant shows this error:

```
error[E0618]: expected function, found `Bar`
  --> neuron/src/async_tokio.rs:16:8
   |
15 | #[derive(::prost::Message)]
   |          ---------------- call expression requires function
16 | struct Bar;
   | -------^^^
   | |
   | `Bar` defined here
```

iMostfa:
Deriving `Message` on a unit struct (`struct Foo;`) failed to compile with `error[E0618]: expected function, found Foo`, while the equivalent `struct Foo {}` worked.

Fixes #1424


iMostfa:
Hi, @LucioFranco Can i ask for a review here ? Thanks

Issue #1451, opened 2026-08-05
Reducing reallocations in `merge_repeated`


javierhonduco:
Hi,

See this CPU profile from our production environment. We use jemalloc and have an application with quite a bit of througput where a non-trivial amount of the CPU cycles are spent reallocating the vector that holds repeated fields.

<img width="2878" height="1514" alt="Image" src="https://github.com/user-attachments/assets/28340eb2-9121-4d98-8ff2-83b27432ca53" />

I believe this would involve a look-ahead to try to figure out the total size (or approximate) to reduce the number of reallocations. Is this something that perhaps you would consider? Let me know!

PR #1454, opened 2026-08-09
perf: reserve repeated message decode capacity

perfloop-agent:
## Summary

- Carry the generated repeated-message field tag through `DecodeContext` so `message::merge_repeated` can reserve capacity for contiguous same-tag fields after decoding the first element.
- Scan the common canonical one-byte tag/length envelope directly, with the generic varint scanner retained for all other inputs, and inline the generated-call-site helper to keep the decode hot path lean.
- Preserve the public four-argument `merge_repeated` API. The narrowly scoped Clippy allowance keeps the existing nested condition compatible with the crate's edition 2021.

## Compatibility and limits

Reservation only considers complete fields in the current buffer chunk and enclosing message scope, and remains bounded by the existing lookahead and reservation limits. Fragmented, incomplete, non-contiguous, or hand-written `Message` decode paths retain the existing fallback behavior. Decoded output is unchanged.

## Performance

**Workload:** `repeated-message-decode-64`

| Metric | Before | After | Change |
| --- | --- | --- | --- |
| `allocation_calls/op` | 5 | 1 | 80% lower |
| `reallocation_calls/op` | 4 | 0 | 100% lower |
| `requested_allocation_bytes/op` | 496 | 256 | 48.4% lower |

**Workload:** `repeated-message-decode-4096`

| Metric | Before | After | Change |
| --- | --- | --- | --- |
| `allocation_calls/op` | 11 | 3 | 72.7% lower |
| `reallocation_calls/op` | 10 | 2 | 80% lower |
| `requested_allocation_bytes/op` | 32752 | 28672 | 12.5% lower |

## Testing

Ran formatting, warning-denied Clippy including Rust 1.95, no-default-features checking, feature-specific test configurations, no-recursion-limit tests, benchmark compilation, and warning-denied rustdoc.

All 8 declared correctness checks passed.

---

Authored and verified by [Perfloop](https://app.perfloop.ai): every claim above was co-measured on both trees and independently re-verified before submission — the full record is public: [case_pfmqga57j8](https://app.perfloop.ai/t/oss/case_pfmqga57j8). Replies from this account are human-approved, and a human operator is accountable for this contribution.


tsenart:
Related to https://github.com/tokio-rs/prost/issues/1451

